PR #82, opened 2025-10-29
feat: add oauth support

joshsny:
Add OAuth flow to Array.

Would recommend trialing out with US / EU for now, rather than the Development option - I've got a seed script for the OAuth application locally that I'll put a PR out for.

### DIFF
--- a/tsconfig.web.json
+++ b/tsconfig.web.json
@@ -9,5 +9,10 @@
   },
   "exclude": ["src/main/**/*", "src/api/repoDetector.ts"],
   "extends": "./tsconfig.json",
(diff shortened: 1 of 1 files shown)


src/main/services/oauth.ts
```diff
@@ -0,0 +1,343 @@
...
+      const code = url.searchParams.get("code");
+      const error = url.searchParams.get("error");
+
+      if (error) {
+        const isAccessDenied = error === "access_denied";
+        res.writeHead(isAccessDenied ? 200 : 400, {
+          "Content-Type": "text/html",
+        });
+        res.end(`
+          <html>
```
joshsny: Note: prettify this and make it auto close ideally

PR #83, opened 2025-10-29
feat/repo selection

jonathanlab:
We now work on a per-repository basis instead of the selected working directory.

The user sets a workspace on start. We then use this to work with the users repos, or clone them if they're not available yet.

image
image


k11kirky (approved):
Approving to unblock, but do we really want to clone every connection that they have in git. I may not want repo's that are connected to my posthog. e.g. I might not want charts working at posthog, both other team members might not want array

k11kirky:
Nice ok 🚢

PR #108, opened 2025-11-01
feat: posthog implementation

k11kirky:
Adding posthog js and posthog node tracking to Array


jonathanlab:
I've updated this to be a map of type Record<Event, EventProperties>, instead using a generic tracking function, so we won't have to create a new function for each event we need to track.

daniloc (approved):

PR #126, opened 2025-11-11
Fix: Get array back into a working state

k11kirky:
- Removed repo first approach, using directory selection instead (I plan to do a deeper refactor here)
- Removed the tab store in favor of a simple navigation store (no more top nav tabs)
- Connected tree view to list real directory
- Read and display todo's in the new todo panel
- Moved plan output back into the markdown renderer (will add a raw view and set it to default)
- Reimplement ability to edit and save MD files - this got broken somehow

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,9 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  '@posthog/agent': link:../agent
-
 importers:
 
   .:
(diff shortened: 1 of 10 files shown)


src/renderer/features/editor/components/PlanEditor.tsx
```diff
@@ -52,63 +55,97 @@ export function PlanEditor({
...
-  // Auto-save with debounce
+  const handleManualSave = useCallback(() => {
+    handleSave(content);
+  }, [content, handleSave]);
+
+  // Track unsaved changes
   useEffect(() => {
-    if (saveTimeoutRef.current) {
-      clearTimeout(saveTimeoutRef.current);
+    if (content !== fetchedContent) {
```
joshsny: nit of all nits, this can be `setHasUnsavedChanges(content !== fetchedContent)`

PR #135, opened 2025-11-18
feat: add vitest testing infrastructure and utilities

jonathanlab:
This sets up `vitest`​ + `@testing-library/react`​ so we can actually verify the panel stuff and all of it's edge cases are handled properly, both in store and UI


lucasheriques (approved):
thank you!!!

PR #139, opened 2025-11-18
feat: integrate code viewer and refactored panels into task detail

jonathanlab:
Now that we've detached the panel logic from the renderer in the previous PRs, we can hook it all up here and get rid of the old taskPanelLayoutStore. Also look at how stupidly simple `TaskDetail.tsx`​ becomes :) Also added a whole bunch of panel splitting tests because there are a lot of edge cases there. Also hooked up the code mirror editor to the file tree panel.


src/renderer/features/task-detail/components/TaskDetail.tsx
```diff
@@ -65,138 +42,9 @@ export function TaskDetail({ task: initialTask }: TaskDetailProps) {
...
-                  }
-                >
-                  <TaskArtifactsPanel taskId={taskId} task={task} />
-                </PanelTab>
-              </PanelLeaf>
-            </PanelGroupTree>
-          </PanelGroupTree>
-        }
-      />
+      <PanelLayout taskId={taskId} task={task} />
```
lucasheriques: i love this

PR #149, opened 2025-11-21
chore: add bin/setup to setup package linking

joshsny:
We were sometimes committing the changes to the pnpm workspace file, 
this avoids committing it altogether and uses a bin/setup command to 
create it depending on your environment.

You can use AGENT_PACKAGE_LOCATION in your env var and then run 
bin/setup to get the pnpm workspace file.


jonathanlab (approved):
great, ty

PR #158, opened 2025-11-25
feat: add folders to sidebar

jonathanlab:
image

We persist these folders in Application Support as well.

Will refactor the incredibly messy sidebar code when I do more work on it later on.

This also scaffolds some logic for application storage that it _outside_ local storage. For example, in application support, we now define a data folders where we maintain a list of folders in Array via `folders.json`. We could also start storing preferences etc in there. Stuff that should persist across installs basically.


apps/array/src/renderer/features/task-detail/components/TaskInput.tsx
```diff
@@ -102,7 +116,37 @@ export function TaskInput() {
           zIndex: 1,
         }}
       >
-        <Box style={{ width: "50%" }}>
+        {/* <Flex direction="row" gap="4" className="w-full mb-4 items-center">
```
jonathanlab: scaffolded the UI here already but will add the rest later on

joshsny (approved):

PR #159, opened 2025-11-25
feat: support common languages in codemirror

jonathanlab:
This add syntax highlighting, not LSPs. Styling looks horrible but that's for another PR.


jonathanlab:
Looks pretty neat in light actually

PR #161, opened 2025-11-25
feat: diff viewer and turbo filetree

jonathanlab:
fastest file tree known to man plus a diff viewer, what more do you need. The code mirror editor is now actually real time, meaning changes to the file are immediately reflected. We also keep track of the file tree now. This happens via parcel/watcher.

The UI is not good, but that'll be worked on in a separate PR

image
image

---

- Adds a new "Changes" panel that shows all modified, added, deleted, and renamed files
- Implements a file system watcher using `@parcel/watcher` to detect file changes in real-time
- Adds a split/unified diff viewer using CodeMirror's merge view
- Improves file tree performance with lazy loading directory contents
- Adds real-time directory updates when files are added, modified, or deleted
- Adds support for detecting file renames in git
- Includes proper handling of file deletions by closing related tabs
- The file tree now only loads directories when they're expanded, making it much faster for large repositories.

### DIFF
--- a/apps/array/src/main/services/fileWatcher.ts
+++ b/apps/array/src/main/services/fileWatcher.ts
@@ -39,6 +39,22 @@ class FileService {
     return path.join(app.getPath("userData"), "snapshots", `${hash}.snapshot`);
   }
 
+  private flushPendingChanges(repoPath: string, state: RepoState): void {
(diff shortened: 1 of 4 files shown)


joshsny (approved):
woohoo

PR #172, opened 2025-11-28
feat: configurable workspaces

jonathanlab:


joshsny (approved):
Gave it a skim, looks good :)

PR #174, opened 2025-12-01
feat: set up ACP and Claude adapter

jonathanlab:
Based on Zed`s ACP claude adapter: https://github.com/zed-industries/claude-code-acp/tree/main


packages/agent/src/adapters/claude/claude.ts
```diff
@@ -0,0 +1,1186 @@
...
+
+    const permissionMode = "default";
+
+    // Extract options from _meta if provided
+    const userProvidedOptions = (params._meta as NewSessionMeta | undefined)
+      ?.claudeCode?.options;
+
+    const options: Options = {
+      systemPrompt,
+      settingSources: ["user", "project", "local"],
```
joshsny: how will this behave if the user has claude code installed locally and has settings there, how do we want it to behave?  
  
Should we inherit their settings, or set this to local and make sure we control the settings in Array?

packages/agent/src/adapters/claude/claude.ts
```diff
@@ -0,0 +1,1186 @@
...
+          message: "User refused permission to run tool",
+          interrupt: true,
+        };
+      }
+    };
+  }
+}
+
+async function getAvailableModels(query: Query): Promise<SessionModelState> {
+  const models = await query.supportedModels();
```
joshsny: Do you know what request this makes? We'll need to support it in the llm gateway

packages/agent/src/adapters/claude/claude.ts
```diff
@@ -0,0 +1,1186 @@
...
+          content: {
+            type: "image",
+            data: chunk.source.type === "base64" ? chunk.source.data : "",
+            mimeType:
+              chunk.source.type === "base64" ? chunk.source.media_type : "",
+            uri: chunk.source.type === "url" ? chunk.source.url : undefined,
+          },
+        };
+        break;
+      case "thinking":
```
joshsny: can we get loading states via this? would be nice to get those messages claude code gives

packages/agent/src/adapters/claude/claude.ts
```diff
@@ -0,0 +1,1186 @@
...
+    case "content_block_stop":
+      return [];
+
+    default:
+      unreachable(event, logger);
+      return [];
+  }
+}
+
+export function runAcp() {
```
joshsny: Do you know what this does?

packages/agent/src/adapters/claude/claude.ts
```diff
@@ -0,0 +1,1186 @@
...
+      },
+    };
+  }
+
+  async authenticate(_params: AuthenticateRequest): Promise<void> {
+    throw new Error("Method not implemented.");
+  }
+
+  async prompt(params: PromptRequest): Promise<PromptResponse> {
+    if (!this.sessions[params.sessionId]) {
```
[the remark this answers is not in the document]
jonathanlab: Yes, we'll call `loadSession` first usually, this is just a safety check

PR #178, opened 2025-12-01
feat: taskrunv2

jonathanlab:
Hooks up the claude adapter to agent.ts. Remove usage of EventAgent. Update example client to show usage.


packages/agent/src/agent.ts
```diff
@@ -125,66 +112,87 @@ export class Agent {
...
-    await this._configureLlmGateway();
+    // await this._configureLlmGateway();
 
     const task = await this.fetchTask(taskId);
     const cwd = options.repositoryPath || this.workingDirectory;
     const isCloudMode = options.isCloudMode ?? false;
     const taskSlug = (task as any).slug || task.id;
 
+    // Create a session for this task run
+    const sessionId = uuidv7();
```
joshsny: Can we use the task run id as the session id here? Don't really see a need to seperate them

joshsny (approved):
Looks good generally, left a couple comments, approving not to block you

PR #190, opened 2025-12-03
fix: extract sdkSessionId when reading from logs

jonathanlab:
we werent reading sdkSessionId from the logs anymore, which caused claude to lose context on session restore. this makes sure we extract it from the logs


apps/array/src/renderer/features/sessions/utils/parseSessionLogs.ts
```diff
@@ -71,12 +73,26 @@ export async function fetchSessionLogs(
         ) {
           notifications.push(stored.notification.params as SessionNotification);
         }
+
+        // Extract SDK session ID from _posthog/sdk_session notification
```
joshsny: Why do we need to do this, isn't session id = task run id?

joshsny: Gotcha, that makes sense!

PR #200, opened 2025-12-04
feat(cloud): add polling for cloud runs

joshsny:
This adds support for polling for cloud run logs, and adds a runTaskCloud method to the agent package which we'll use in cloud runs that uses the new ACP format


k11kirky (approved):
This looks awesome! Great work!

Only thing I am not sure on is why we have a stream for the notifications / interupts, but we still poll for logs - should we just move to streaming bi directionally at this point?

PR #241, opened 2025-12-10
fix/feat: Improve @ file tagging, reduce lag

charlesvien:
- Improve @ file tagging in the message editor, reduce lag by switching to `git ls-files` over `fs`. It is very fast because it reads from the git index. It also respects .gitignore, which I think is something we would want anyways?


joshsny (approved):
LGTM - left a couple of comments :)

PR #259, opened 2025-12-15
feat: Cleanup task creation (saga, services, di)

charlesvien:
Jonathan and I had a chat last Thursday about reliability being a concern because we can’t validate the product if people are churning because they are hitting too many bugs trying to use it. I took up the challenge of tackling the troublesome logic around creating a new task while Jonathan works on sessions 🚢. Creating a new task is currently a 10 step react hook which can fail at any point and leave the app in a semi-initialized or broken state. 

I started at the bottom with implementing a more reliable communication layer between the main and render process. No more duplicating types between the two and bloated boilerplate for simple functions. It seems everyone was happy with the decisions around tRPC with this and it’s working great with the four routers it’s running so far.

The next, and more important step was to create a custom saga pattern for us to use when we have large multi-step logic blocks that need to be durable and or atomic. You can see the new way we handle creating a task, setting up the workspace, etc and rolling back if any step fails. There’s also a new way of defining services and dependencies that will make writing tests a breeze in the future and in the short term it will help with making things more SOLID and decoupled from UI components which should be generally dumb.

I have two PR’s up for this currently, I’d recommend taking a closer look than usually at them as I’m not just pushing things to be more reliable with these changes I’m defining new contracts on how we develop features going forward which will make us faster and AI code better patterns. Unfortunately Claude Code/AI will happily write massive React hooks like `useTaskCreation` right now because we need to define a stronger foundation for it to build on and this should help it do that. 

**NOTE:** I've only converted what was needed to the new service, tRPC IPC, saga, etc to demonstrate the proposed changes and fix the task creation flow. If we all agree on these patterns we can gradually migrate them over time as we touch things 🚢🚢

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,224 @@
+# Contributing to Array
+
+## Architecture Overview
+
+Array is an Electron app with a React renderer. The main process handles system operations (stateless), while the renderer owns all application state.
+
+```
+Main Process (Node.js)                      Renderer Process (React)
+┌───────────────────────┐                   ┌───────────────────────────┐
+│  DI Container         │                   │  DI Container             │
+│  ├── GitService       │                   │  ├── TRPCClient           │
+│  └── ...              │                   │  └── TaskService, ...     │
+├───────────────────────┤                   ├───────────────────────────┤
+│  tRPC Routers         │ ◄─tRPC(ipcLink)─► │ tRPC Clients              │
+│  (use DI services)    │                   │  ├── trpcReact (hooks)    │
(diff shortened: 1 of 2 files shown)


jonathanlab (approved):
This looks great! As a general thought: It might be good to establish som rules about where we want to maintain application state. We could probably can make all services in main pure(-ish)/stateless, and keep all state in our renderer. Love this rollback pattern <3

jonathanlab:
Just another thought: We need consistent UI for handling and rendering errors from both backend and frontend similar to gitbutler: 
image

PR #267, opened 2025-12-15
feat: bash mode

jonathanlab:
image

image

### DIFF
--- a/apps/array/src/main/services/shell.ts
+++ b/apps/array/src/main/services/shell.ts
@@ -1,3 +1,4 @@
+import { exec } from "node:child_process";
 import { createIpcHandler } from "../lib/ipcHandler";
 import { shellManager } from "../lib/shellManager";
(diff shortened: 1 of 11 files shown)


apps/array/src/main/services/shell.ts
```diff
@@ -56,4 +57,23 @@ export function registerShellIpc(): void {
...
+
+  handle(
+    "shell:execute",
+    async (
+      _event,
+      cwd: string,
+      command: string,
+    ): Promise<{ stdout: string; stderr: string; exitCode: number }> => {
+      return new Promise((resolve) => {
+        exec(command, { cwd, timeout: 60000 }, (error, stdout, stderr) => {
```
k11kirky: how do we feel about a timeout here? I worry if people are using this to run a dev server it might be an issue

k11kirky (approved):
lgtm

PR #269, opened 2025-12-16
feat: Live plan mode

jamesefhawkins:
- [x] Remove AI slop from this code
- [x] Get the plan to persist across separate sessions (not sure how this will work yet - this could be a second PR though)

1) I did wonder if we should make this mode a configuration - i'd love feedback if this style of UX is useful for others. I like it personally....
2) should it say "live plan" not "plan" to be clearer

if i had more time, i'd want a bit more focus in the prompts on it creating a real todo list and ticking things off - it's somewhat haphazard atm


.posthog/4ff6481c-c37d-4791-bddc-9a8ab007a068/plan.md
joshsny: we can remove these plan files from the diff as we don't commit them any more :grinning:

PR #273, opened 2025-12-16
refactor: properly use inversify

jonathanlab:
I was doing some dumb stuff with .get<T> instead of just injecting the services.


charlesvien:
### Merge activity

* **Dec 16, 6:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/Array/273).

PR #275, opened 2025-12-16
refactor: add output schemas, extract and use input schema as types

jonathanlab:
We previously were not enforcing output schemas, and also not using
the input schemas as types, meaning we did not have 1 source of truth.


charlesvien:
### Merge activity

* **Dec 16, 6:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/Array/275).

PR #286, opened 2025-12-17
feat: Refactor git service

charlesvien:


joshsny (approved):
didn't review the git stuff in detail as looks like its just shuffling existing logic around, lmk if you want a detailed review there

PR #287, opened 2025-12-17
feat: Refactor agent service

charlesvien:


apps/array/src/main/trpc/routers/agent.ts
```diff
@@ -0,0 +1,80 @@
...
+      const options = opts.signal ? { signal: opts.signal } : undefined;
+
+      for await (const [payload] of on(
+        service,
+        AgentServiceEvent.SessionEvent,
+        options,
+      )) {
+        const event = payload as AgentSessionEventPayload;
+        if (event.sessionId === targetSessionId) {
+          yield event.payload;
```
joshsny: same here with tracked, do we want to use it?

apps/array/src/renderer/components/GlobalEventHandlers.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onToggleCommandMenu,
+  commandMenuOpen,
+}: GlobalEventHandlersProps) {
+  const { toggleSettings, navigateToTaskInput, goBack, goForward } =
+    useNavigationStore();
+  const clearAllLayouts = usePanelLayoutStore((state) => state.clearAllLayouts);
+  const toggleLeftSidebar = useSidebarStore((state) => state.toggle);
+  const toggleRightSidebar = useRightSidebarStore((state) => state.toggle);
+
+  const handleOpenSettings = useCallback(() => {
```
joshsny: does this callback do anything if the function itself is a dependency?

PR #362, opened 2025-12-19
fix: persist navigation state

### LINKED ISSUE #355: Current screen is not persisted on reload
super annoying

jonathanlab:
closes #355


apps/array/src/renderer/components/MainLayout.tsx
```diff
@@ -8,27 +8,35 @@ import { SettingsView } from "@features/settings/components/SettingsView";
...
     isOpen: shortcutsSheetOpen,
     toggle: toggleShortcutsSheet,
     close: closeShortcutsSheet,
   } = useShortcutsSheetStore();
+  const { data: tasks } = useTasks();
 
   useIntegrations();
   useTaskDeepLink();
 
+  useEffect(() => {
```
joshsny: is this necessary, I think everything here is in a store, also why do we pass `tasks` to `hydrateTask`?

Should it be `hydrateTasks(tasks)` or `for task in tasks; hydrateTask(task)`?

joshsny (approved):

PR #368, opened 2025-12-19
feat: Implement error toasts

jonathanlab:
This makes handling errors and warnings easier for our setup.

In development, we now render toasts for uncaught exceptions and unhandled rejections in both renderer and main, preload errors, and trpc query/mutation errors.

We can now also send any toast to the user from the main process via the `user-notifications` service.

I added a react error boundary, not sure if we want to keep this around.

### DIFF
--- a/apps/array/vite.main.config.mts
+++ b/apps/array/vite.main.config.mts
@@ -19,7 +19,7 @@ function _getBuildDate(): string {
 }
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const monorepoRoot = path.resolve(__dirname, "../..");
+const _monorepoRoot = path.resolve(__dirname, "../..");
 
 /**
  * Custom Vite plugin to fix circular __filename references in bundled ESM packages.
--- a/apps/array/src/renderer/utils/toast.tsx
+++ b/apps/array/src/renderer/utils/toast.tsx
@@ -1,15 +1,71 @@
 import {
   CheckIcon,
+  Copy,
   InfoIcon,
   WarningIcon,
   X,
   XIcon,
 } from "@phosphor-icons/react";
-import { Card, Flex, IconButton, Spinner, Text } from "@radix-ui/themes";
+import {
+  Box,
+  Button,
+  Card,
+  Dialog,
+  Flex,
+  IconButton,
+  Inset,
+  Spinner,
+  Text,
+  Tooltip,
+} from "@radix-ui/themes";
+
+import { useMemo, useState } from "react";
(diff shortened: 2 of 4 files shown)


apps/array/src/renderer/components/ErrorBoundary.tsx
jonathanlab: Maybe we should just render the error in the screen instead of a toast if it triggers the error boundary.

apps/array/src/renderer/lib/logger.ts
```diff
@@ -1,28 +1,17 @@
...
-// Ensure logs appear in dev tools console
 log.transports.console.level = "debug";
 
-export const logger = {
-  info: (message: string, ...args: unknown[]) => log.info(message, ...args),
-  warn: (message: string, ...args: unknown[]) => log.warn(message, ...args),
-  error: (message: string, ...args: unknown[]) => log.error(message, ...args),
-  debug: (message: string, ...args: unknown[]) => log.debug(message, ...args),
+const isDev = import.meta.env.DEV;
+const devErrorToastsEnabled =
```
jonathanlab: We set this 3 times, probably just extract it instead

apps/array/vite.main.config.mts
```diff
@@ -155,6 +156,7 @@ function copyClaudeExecutable(): Plugin {
 const forceDevMode = process.env.FORCE_DEV_MODE === "1";
 
 export default defineConfig({
+  envDir: monorepoRoot,
```
jonathanlab: Vite was struggling with the `.env` resolution

joshsny (approved):

apps/array/src/renderer/lib/error-handling.ts
```diff
@@ -0,0 +1,50 @@
+import { formatArgsToString } from "@shared/utils/format";
+import { toast } from "@utils/toast";
+import { IS_DEV } from "@/constants/environment";
+import { logger } from "./logger";
+
+const devErrorToastsEnabled =
+  IS_DEV && import.meta.env.VITE_DEV_ERROR_TOASTS !== "false";
+
```
jonathanlab: Is this consistent with how we set dev in main? I believe not.

apps/array/vite.main.config.mts
```diff
@@ -19,6 +19,7 @@ function _getBuildDate(): string {
 }
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const monorepoRoot = path.resolve(__dirname, "../..");
```
jonathanlab: Are we using this for anything?

PR #382, opened 2025-12-26
fix: Allow default keybinds such as minimize, hide, etc

charlesvien:


sakce:
Boo it needs the team review

PR #383, opened 2026-01-05
fix: ensure correct PATH when launched from Finder/Spotlight

### LINKED ISSUE #312: gh cli, pnpm etc, is not available in PATH in bundled app
Not sure if this one is still reproducible, otherwise can be closed

jonathanlab:
When you launch an electron app from the finder or from spotlight, the wrong `PATH` gets passed into the app. 

This fixes #312 by reading `PATH` on launch instead of assuming that it is always available.

Verified that it works by packaging the app and opening it via the finder.

Based on https://github.com/sindresorhus/fix-path

Closes #312

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,6 +33,8 @@
 
 ## Code Style
 
+- Prefer writing our own solution over adding external packages when the fix is simple
+- Keep functions focused with single responsibility
 - Biome for linting and formatting (not ESLint/Prettier)
 - 2-space indentation, double quotes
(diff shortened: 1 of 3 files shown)


CLAUDE.md
```diff
@@ -33,6 +33,8 @@
 
 ## Code Style
 
+- Prefer writing our own solution over adding external packages when the fix is simple
+- Keep functions focused with single responsibility
```
joshsny: SOLID <3

PR #393, opened 2026-01-05
feat: Repository and task handling improvements

charlesvien:
- We now have settings tab for the repository. Currently the only option is to delete the repository with all associated tasks and their workspaces.
- If the repository folder is moved (even in runtime) then it will let the user know they need to restore the folder to continue with their tasks. Previously this was just broken and asked the user if they wanted to initialize a git repo (in a loop?).
- Added a touch more padding on sidebar items, open to feedback on it.
- Implemented the ability to pin and delete a tasks from the sidebar. The default sort order for pinned and unpinned is by latest activity.

### DIFF
--- a/apps/array/src/renderer/features/sidebar/components/items/TaskItem.tsx
+++ b/apps/array/src/renderer/features/sidebar/components/items/TaskItem.tsx
@@ -10,6 +10,7 @@ import { trpcVanilla } from "@renderer/trpc";
 import { formatRelativeTime } from "@renderer/utils/time";
 import type { WorkspaceMode } from "@shared/types";
 import { useQuery } from "@tanstack/react-query";
+import { useMemo } from "react";
 import { SidebarItem } from "../SidebarItem";
 
 function useCurrentBranch(repoPath?: string, worktreeName?: string) {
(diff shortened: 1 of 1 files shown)


apps/array/src/renderer/App.tsx
```diff
@@ -16,6 +17,15 @@ function App() {
     initializePostHog();
   }, []);
 
+  // Reload folders when window regains focus to detect moved/deleted folders
```
joshsny: question for our future selves, I wonder how we want to handle these kinds of listeners, I am sure we'll collect a lot of them, some kind of listener service maybe?

apps/array/src/renderer/components/MainLayout.tsx
```diff
@@ -37,6 +42,19 @@ export function MainLayout() {
     }
   }, [tasks, hydrateTask]);
 
+  // Check if current task's folder became invalid (e.g., moved while app was open)
```
joshsny: nit: you control all dependecies of this useEffect, can it go in a store instead?

apps/array/src/renderer/features/sidebar/components/items/TaskItem.tsx
```diff
@@ -126,10 +177,29 @@ export function TaskItem({
     <span className="flex h-[12px] w-[12px] items-center justify-center text-[8px] text-green-11">
       ■
     </span>
+  ) : isPinned ? (
+    <PushPin size={12} className="text-accent-11" />
   ) : (
     <GitBranchIcon size={12} />
   );
 
+  const endContent = (
```
joshsny: nit: this should be pulled out, or wrapped in a `useMemo` so it has a stable reference

PR #411, opened 2026-01-08
Array Plan Mode

k11kirky:
- Removed a bunch of the now dead code from the agent package
- Added plan mode execution type to agent package
- Added UI / UX elements to support tool call permissions.


jonathanlab (approved):
🚢

PR #447, opened 2026-01-09
Always show panel icons

### LINKED ISSUE #442: We should always render the split and plus icon per panel
Currently we rely on focus tracking per panel, but that feels unreliable and buggy, let's simplify this to always show the split icon and plus icon for each panel.

jonathanlab:
Closes #442


jonathanlab:
* **#447** Always show panel icons 👈
* `main`

---

Merge from bottom to top, or use `arr merge`

joshsny (approved):

PR #463, opened 2026-01-12
feat: retain chatbox focus in common scenarios

### LINKED ISSUE #457: Better chatbox focussing
Currently we do not focus the chatbox when you:
- Interrupt a chat using esc or the stop button
- Navigate to the chat (either from another tab/pane, or from another task)
We should automatically focus in those cases.

jonathanlab:
We've had some cases where we would not be focussing the chat box when the user probably wants to interact with it:
- Navigating to a task
- Navigating from another tab to the chat tab
- Interrupting the agent

In these scenarios we'll now automatically focus the chatbox

Closes #457


jonathanlab:
* **#463** feat: retain chatbox scenarios in common scenarios 👈
* `main`

---

Merge from bottom to top, or use `arr merge`

PR #471, opened 2026-01-12
feat: persist session mode

### LINKED ISSUE #456: Persist selected mode for the same chat across sessions
Currently I have to set my mode on every restart of Array for a session which is tedious. Instead, we should just remember the previously active mode and use that.

jonathanlab:
Closes #456


charlesvien:
https://github.com/PostHog/Array/issues/446

PR #474, opened 2026-01-12
plan: cloud architecture

joshsny:
I’m going to be looking at the cloud experience in array next, so wanted to bash out some thoughts on the architecture.

Will use this as a place to iterate and get thoughts.

### DIFF
--- a/notes/CLOUD_ARCHITECTURE.md
+++ b/notes/CLOUD_ARCHITECTURE.md
@@ -190,7 +190,7 @@ Event Log (run_{id}.jsonl):
 **To recover current state:**
 
 1. Find latest `git_commit` event → checkout that commit
-2. Replay `file_change` events since → apply uncommitted changes
+2. Replay `file_change` events since the last commit → apply uncommitted changes
 3. Fetch file contents from S3 by hash
 
 ### Recovery Flow


notes/CLOUD_ARCHITECTURE.md
```diff
@@ -0,0 +1,498 @@
...
+  { method: "_posthog/git_commit", params: { sha: "def456" } }        ◄── latest commit
+  { method: "_posthog/file_change", params: { path: "src/foo.py", hash: "sha256_ccc" } }
+  { method: "_posthog/file_change", params: { path: "src/baz.py", hash: "sha256_ddd" } }
+```
+
+**To recover current state:**
+
+1. Find latest `git_commit` event → checkout that commit
+2. Replay `file_change` events since → apply uncommitted changes
+3. Fetch file contents from S3 by hash
```
jonathanlab: This does not make much sense to me. Why not just emit `file_change` events for the files that change based on the `git_commit` instead of replaying the entire history?

notes/CLOUD_ARCHITECTURE.md
```diff
@@ -0,0 +1,498 @@
...
+## Communication Protocol
+
+### Streamable HTTP
+
+Following [MCP's pattern](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http):
+
+- **POST** — Client sends messages (user input, file sync, cancel)
+- **GET** — Client opens SSE stream for server events
+- **Session-Id header** — Identifies the session (run ID)
+- **Last-Event-ID header** — Resume from where you left off
```
jonathanlab: This section implies that these will follow the [ACP format](https://agentclientprotocol.com/protocol/extensibility) right? 

I assume this means that all ACP notifications will also go over streamable HTTP, meaning we will not need to poll logs anymore in cloud?

notes/CLOUD_ARCHITECTURE.md
```diff
@@ -0,0 +1,498 @@
...
+  { method: "_posthog/file_change", params: { path: "src/bar.py", hash: "sha256_bbb" } }
+  { method: "_posthog/git_commit", params: { sha: "def456" } }        ◄── latest commit
+  { method: "_posthog/file_change", params: { path: "src/foo.py", hash: "sha256_ccc" } }
+  { method: "_posthog/file_change", params: { path: "src/baz.py", hash: "sha256_ddd" } }
+```
+
+**To recover current state:**
+
+1. Find latest `git_commit` event → checkout that commit
+2. Replay `file_change` events since → apply uncommitted changes
```
joshsny:
```suggestion
2. Replay `file_change` events since the last commit → apply uncommitted changes
```

notes/CLOUD_ARCHITECTURE.md
jonathanlab: >When a client wants to recover the state of the agent (e.g. due to disconnection, or a new client interacting with the task), it can pull from the latest commit but it also needs the changes the agent has made since that commit, which is why we replay the file_change events since that commit from the log.

But if an agent rolls back to let's say, commit A, and we have a bunch of unstaged work, doesn't the following happen (assuming we checkout in a way that does not preserve uncommited files):
- We delete all files that have not been committed to git yet. This emits `file_change`.
- We restore existing files to their state at commit A. This also emits a `file_change`.

So this means that that combined set of `file_changes`s will give us the exact same state right? Meaning we do not need to replay or anything like that, just handle the events generated by the checkout.

tatoalo:
Great stuff @joshsny 🐐, really like the direction of this!  🔥 

Some thoughts: 

I really like using Temporal for lifecycle only, we are using it in ph-ai as hot-path message bus (with redis streams in front) and we are incurring on latency and additional complexity that here is simply not needed.

On the `Content-Addressed Storage` part, not super sure about a single `jsonl` blob that grows given possible concurrent retries/writes/failures so we could probably think at segmenting with progressive ids inside the outer `run_id` to have atomic writes (each segment immutable), fetching segments is more efficient for picking up what to replay on recovery.

On the "local wins" approach, I agree on the direction but perhaps we would need some kind of "caching" layer but maybe this is an early optimization problem.

Maybe too technical at this stage but the watcher may need strong ignore rules and operation batching, let's say that we run a package manager command, the file watcher would have a ton of events in the same path so we might want to avoid emitting all of those at once.

I am mostly thinking about possible ordering of changes, as in: multiple writes can arrive out of order with respect to the local client but we can define this in another passage.

Q: what kind of ack are you thinking for `_posthog/ack`?

PR #485, opened 2026-01-13
fix: Restore draft on mount or when sessionId/editor changes

charlesvien:


charlesvien:
https://github.com/PostHog/Array/issues/469

PR #486, opened 2026-01-13
fix: Skip protocol registration in development to avoid hijacking deep links

charlesvien:


charlesvien:
https://github.com/PostHog/Array/issues/461

PR #530, opened 2026-01-17
feat: Implement `AppLifecycleService` and refactor to fix auto updater

charlesvien:
## Alright here's what I found in the cave...

**the bug:** when `autoUpdater.quitAndInstall()` was invoked, it triggered the `before-quit` event, but the handler would intercept it, prevent the default behavior, and forcibly exit the app before the updater could complete its installation flow.

**solution:** this PR introduces a new `AppLifecycleService` that centralizes all shutdown logic and tracks an `isQuittingForUpdate` flag. Now when installing an update, the `UpdatesService` first calls `lifecycleService.shutdown()` to gracefully clean up, then sets `lifecycleService.setQuittingForUpdate()` before calling `quitAndInstall()`. The `before-quit` handler now checks this flag and returns early if set, allowing the auto updater to proceed unblocked. This separates the "normal quit with cleanup" path from the "quit for update" path  
  
## NOTE: There is a different event sequence for `quitAndInstall()`
_Excerpt taken from electron docs_

The Event Sequence for `quitAndInstall()`

**Normal quit:**

1. `before-quit` emitted
2. Windows receive `close` events
3. Windows close
4. `will-quit` emitted
5. App exits

**`quitAndInstall()`** **sequence (different!):**

1. All windows receive `close` events first
2. Windows close
3. `before-quit` emitted **after** windows are closed
4. Installer runs

### DIFF
--- a/apps/twig/src/main/services/app-lifecycle/service.test.ts
+++ b/apps/twig/src/main/services/app-lifecycle/service.test.ts
@@ -1,4 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
+import { AppLifecycleService } from "./service.js";
 
 const { mockApp, mockAgentService, mockTrackAppEvent, mockShutdownPostHog } =
   vi.hoisted(() => ({
@@ -43,8 +44,6 @@ vi.mock("../../../types/analytics.js", () => ({
     APP_QUIT: "app_quit",
   },
 }));
-
-import { AppLifecycleService } from "./service.js";
 
 describe("AppLifecycleService", () => {
   let service: AppLifecycleService;


charlesvien:
* **#530** <a href="https://app.graphite.com/github/pr/PostHog/Array/530?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/Array/530?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

charlesvien:
This should also address "Point 5: Manual Update Issue" on the Array Feedback PDF

PR #572, opened 2026-01-22
feat(signals): Autonomy onboarding

Twixes:
# feat(signals): Show auto-detected tasks as suggested tasks

This PR adds support for displaying auto-detected tasks from session analysis as suggested tasks. Codename: Analyst.

As part of this: Integration with PostHog API to fetch task references from session recordings. A new task preview view that shows details of auto-detected tasks including session references. UI to display these references with expandable details.

In dev, those suggested tasks are mocked for easy preview.

(Also had to update the VS Code debug config to work post-rename.)


joshsny (approved):
I got this when running locally, I think we only request project:read right now for the oauth token, so we might need to update that.

Screenshot 2026-02-04 at 14 52 57

I liked this a lot:
Screenshot 2026-02-04 at 14 57 46

I couldn't seem to get past the auth bug after I tried, but generally think this'd be great to get out there and iterate on. Could we put it behind a feature flag for now so that we can play around a bit with the UX in development?

PR #650, opened 2026-01-27
feat: Implement permissions sorting

andrewjmcgehee:
allow once before allow always for folks that don't like allow always


charlesvien:
@jonathanlab do we still need this after your permissions stack?

PR #672, opened 2026-01-28
feat: Implement `twig` cli command to launch twig from cli

ReeceJones:
My workflow revolves around the CLI, and I launch all of my tools from the command line. Imagine my shock when I couldn't do the same with Twig!

This PR adds functionality to install and uninstall a twig cli that is used to launch the Twig electron application. 

[demo](https://www.loom.com/share/716e0956e43143a5ad61751c423cd4f6)


charlesvien:
Talked to Reece on Slack about this, love it, 2s thought here: how does cursor handle the relationship between their vscode fork and the cursor cli command? if I uninstall the fork does it clean up the CLI command properly? Someone should do a sanity check on how they've done it before we merge this so we make sure there's a binding relationship with the desktop app.

ReeceJones:
> Talked to Reece on Slack about this, love it, 2s thought here: how does cursor handle the relationship between their vscode fork and the cursor cli command? if I uninstall the fork does it clean up the CLI command properly? Someone should do a sanity check on how they've done it before we merge this so we make sure there's a binding relationship with the desktop app.

Looked into this it turns out cursor changed how they do this at some point. It used to be exactly like this PR where there was a "install cursor command" button in their settings. However, they now automatically install the CLI command when you install the application and uninstall the command when you uninstall the application. I think this is because they moved the install cli command option into their onboarding flow? I'm not 100% sure how they do this, but I think it is something like this:
1. At some point after opening the DMG or installing cursor, a new background service called `CursorUIViewService` is launched. This background process persists even after cursor has been force quit (cmd + q).
2. This background service watches the application directory for the cursor binary
3. If the cursor binary is present in the application directory it attempts to install the cli script if it doesn't exist yet
4. If the binary is not present it attempts to remove the cli script

It is not possible on macos to have a post-install and post-uninstall script run, so I think you need a persistent background process to listen to these events.

PR #678, opened 2026-01-29
chore: update cloud architecture notes after discussions

joshsny:
We talked about how we want to architect the cloud agents in london, this makes some edits to the architecture notes to reflect that.

### DIFF
--- a/notes/CLOUD_ARCHITECTURE.md
+++ b/notes/CLOUD_ARCHITECTURE.md
@@ -9,7 +9,6 @@ This creates two distinct experiences:
 **Interactive Mode** — "I'm watching"
 
 - Real-time feedback as the agent works
-- Files sync to your local machine instantly
 - You can interrupt, redirect, answer questions
 - Feels like pair programming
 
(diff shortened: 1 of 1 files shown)


notes/CLOUD_ARCHITECTURE.md
```diff
@@ -40,234 +41,483 @@ Most cloud agent implementations force you to choose one or the other. The goal
...
+- Cost-effective for storage
+
+### DynamoDB Schema
+
+**Table: `agent_events`**
+
+| Key                   | Type   | Description                                        |
+| --------------------- | ------ | -------------------------------------------------- |
+| `pk` (Partition Key)  | String | `{task_id}#{run_id}` - groups all events for a run |
+| `event_id` (Sort Key) | Number | Monotonic per run, used for Last-Event-ID          |
```
tatoalo: DDB atomic counter internally should be enough here

tatoalo (approved):
Great stuff @joshsny! Just added a couple of comments but looks fire 🔥 

Question (maybe not needed to be discussed here though, feel free to ignore):

In the larger arch diagram we are just putting everything inside `POSTHOG BACKEND`, are we still thinking of having this as an ad-hoc service or not something of priority right now?

PR #731, opened 2026-02-02
feat: improve workspace focus UX

jonathanlab:
I have added a panel beneath the Changes/Files panel that lets you focus your changes. This is identical to the button that was previously in the "Edit workspace" button, which was found confusing by a lot of users. This UI should be more obvious/explicit in explaining what the functionality does.

When clicked, the panel opens a terminal in your repository root.

We also got rid of the default terminal that opens either in the same panel as your chat, or split below the chat. Instead, this terminal in the users workspace path is now opened in same pane as the focus panel, below the Changes/Files panel.

The shortcut for focusing is _CMD+R_. I've moved hot reload to _CMD+SHIFT+R_.

Also moved command palette from CMD+K to CMD+SHIFT+P because the latter conflicts with clearing your terminal

image
image


joshsny (approved):
Looks way nicer and more explicit 🙌

PR #742, opened 2026-02-03
feat: Make it so users can edit queued messages (up arrow)

charlesvien:


charlesvien:
https://github.com/PostHog/Twig/issues/735

PR #747, opened 2026-02-03
feat: Implement SleepService + setting to prevent sleep while the agent is working

charlesvien:


charlesvien:
https://github.com/PostHog/Twig/issues/741

PR #749, opened 2026-02-03
feat(cloud): agent server

joshsny:
Adds the agent server that will run in the sandbox.

This has a server that clients can connect to directly to stream events and send commands.

---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #749 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #748


tatoalo (approved):
Looks great, just a couple of nits and maybe adding heartbeat for SSEs to prevent timeouts down the line?

PR #806, opened 2026-02-05
feat(ui): add cloud workspace mode option

tatoalo:
Adds a "Cloud" option to the workspace mode selector in the task creation UI, this is currently gated behind [twig-cloud-mode-toggle](https://us.posthog.com/project/2/feature_flags/534008)  FF.
 
- Tweaked task creation hook to support cloud workspace mode


tatoalo:
@joshsny added FF support, and created [twig-cloud-mode-toggle](https://us.posthog.com/project/2/feature_flags/534008) so we can play around with it more easily.

PR #863, opened 2026-02-09
feat: Create a preview session on task input page

k11kirky:
### TL;DR

Added a preview session feature for the task input page that loads adapter-specific configuration options without creating a real task.

### What changed?

- Created a preview session system that loads model options, reasoning levels, and execution modes for the selected adapter
- Added a visual mode indicator that shows the current execution mode with an icon and label
- Separated model and reasoning level selectors from the editor toolbar for better organization
- Disabled log persistence for preview sessions since they don't correspond to real tasks
- Added keyboard shortcut (Shift+Tab) to cycle through execution modes
- Implemented version tracking to prevent race conditions when rapidly switching adapters

### How to test?

1. Open the task input page and verify that model options and reasoning levels load correctly
2. Change the adapter and confirm that the available options update accordingly
3. Use Shift+Tab to cycle through execution modes and verify the mode indicator updates
4. Check that no logs are persisted for preview sessions
5. Rapidly switch between adapters to ensure no race conditions occur

### Why make this change?

The task input page needed to display adapter-specific configuration options before task creation, but previously these options were only available after creating a task. This change improves the user experience by showing relevant options upfront, allowing users to configure their task properly before submission. It also prevents unnecessary log persistence for these temporary preview sessions.

### DIFF
--- a/apps/twig/src/main/services/agent/service.ts
+++ b/apps/twig/src/main/services/agent/service.ts
@@ -442,13 +442,16 @@ export class AgentService extends TypedEventEmitter<AgentServiceEvents> {
     const mockNodeDir = this.setupMockNodeEnvironment(taskRunId);
     this.setupEnvironment(credentials, mockNodeDir);
 
+    // Preview sessions don't persist logs — no real task exists
+    const isPreview = taskId === "__preview__";
+
     // OTEL log pipeline or legacy S3 writer if FF false
-    const useOtelPipeline = await this.isFeatureFlagEnabled(
-      "twig-agent-logs-pipeline",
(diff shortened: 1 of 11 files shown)


charlesvien (approved):

charlesvien:
I'm having some problems running this locally, the model dropdown never populates and stays loading forever.

Screenshot 2026-02-10 at 10 51 38 AM

P.S.: I also haven't been able to get codex working in general but I added a seperate issue for that.
https://gist.github.com/charlesvien/c2aadc08f101aa2e0fc643a86adeca54

PR #888, opened 2026-02-10
feat: agent chunks coalescing

tatoalo:
We are currently storing every `agent_message_chunk` delta to S3, this doesn't spark joy.
Let's buffer'em and emit a single `agent_message` event per turn segment.

Keeping the buffer since resume saga works on top of this mechanism for now.


joshsny (approved):
🚢🚢🚢

PR #901, opened 2026-02-11
feat: update task run status during local session lifecycle

tatoalo:
Local task runs now show as "in progress" in the task list, added  signalComplete() to the agent package as infrastructure for future use


apps/twig/src/renderer/features/sessions/service/service.ts
```diff
@@ -674,6 +690,20 @@ export class SessionService {
         notifyPromptComplete(session.taskTitle, stopReason);
       }
 
+      // Update remote task run status when agent turn completes with no pending work
```
joshsny: How should this work for interactive runs in Twig? I think this belongs to the agent package, and that for interactive runs I am not sure how we decide to transition to "completed" since the agent will regularly stop and then I prompt it again, probably the only reliable sign is PR merged. Thoughts?

PR #902, opened 2026-02-11
refactor(agent): remove OTEL write path from SessionLogWriter

tatoalo:
Remove the OTEL transport path from `SessionLogWriter`, keeping only the S3 write path


joshsny (approved):
🚢🚢🚢

PR #918, opened 2026-02-13
feat: posthog plugin service

skoob13:
## TLDR:

This PR ships the embedded plugin for Claude Code and skills sync to Codex (as Codex doesn't have a plugin system yet). Skills are automatically pulled from `posthog/posthog`. The next step might be to reuse the same service for Wizard.

Also a few important changes:
- The MCP is pinned to version 2–it ships the new PostHog data discovery through a unified HogQL interface.
- Sets Claude Code's feature flag to offload all MCP tools to a virtual CLI, enabling progressive discovery. It works very well with skills.

Screenshot 2026-02-18 at 14 52 16

## Problem

Twig agents (Claude Code, Codex) need access to PostHog-specific skills, but currently there's no mechanism to deliver or update these skills. We need a plugin system that:

1. Ships a baseline set of skills with the app
2. Keeps skills up-to-date without requiring app releases
3. Supports local skill development for contributors
4. Works across both Claude Code (via plugins) and Codex (via `~/.agents/skills/`)

## Changes

**New PostHog plugin system** (`plugins/posthog/`) — a directory-based plugin format with `plugin.json`, MCP/LSP configs, and skill directories. Skills are markdown files (`SKILL.md`) that get passed to agents as context.

**Three-tier skill sourcing with priority layering:**
| Source | Location | Priority |
|---|---|---|
| Shipped | `plugins/posthog/skills/` (committed) | Lowest |
| Remote | Downloaded from GitHub releases (`skills.zip`) | Middle |
| Local dev | `plugins/posthog/local-skills/` (gitignored) | Highest (dev only) |

**`PosthogPluginService`** (`apps/twig/src/main/services/posthog-plugin/service.ts`) — injectable singleton that manages the full plugin lifecycle at runtime:
- On startup: copies bundled plugin to `userData`, overlays any cached remote skills, syncs to Codex directory, starts a 30-minute update timer
- On update: downloads `skills.zip`, extracts with atomic swap (`.new`/`.old` dirs to avoid partial states), re-assembles the runtime plugin, re-syncs to Codex
- Includes throttling, reentrance guards, and graceful error handling (download failures are non-fatal — existing skills are preserved)

**Build-time Vite plugin** (`copyPosthogPlugin` in `vite.main.config.mts`) — at bundle time, copies the plugin directory to the build output, downloads and overlays remote skills, and in dev mode overlays `local-skills/` on top. Watches all three sources for hot-reload.

**Agent integration** (`AgentService`) — injects `PosthogPluginService` and passes `getPluginPath()` to Claude Code sessions via the `plugins` option. Also adds `x-posthog-mcp-version: 2` header to the PostHog MCP server config, and sets `ENABLE_EXPERIMENTAL_MCP_CLI` env var.

**Dev tooling** — `pnpm skills:pull` script (`scripts/pull-skills.mjs`) downloads remote skills into `local-skills/` for local editing and iteration.

**Packaging** — updated Forge config to unpack `plugins/posthog/**` from ASAR so the plugin directory is accessible at runtime.

## Tests

Manual testing & service unit tests

### DIFF
--- a/apps/twig/vite.main.config.mts
+++ b/apps/twig/vite.main.config.mts
@@ -4,8 +4,8 @@ import {
   cpSync,
   existsSync,
   mkdirSync,
-  readFileSync,
   readdirSync,
+  readFileSync,
   statSync,
 } from "node:fs";
 import { cp, mkdir, readdir, rm, writeFile } from "node:fs/promises";


packages/agent/src/adapters/claude/session/options.ts
```diff
@@ -88,6 +88,7 @@ function buildEnvironment(): Record<string, string> {
     ...process.env,
     ELECTRON_RUN_AS_NODE: "1",
     CLAUDE_CODE_ENABLE_ASK_USER_QUESTION_TOOL: "true",
+    ENABLE_EXPERIMENTAL_MCP_CLI: "true",
```
joshsny: 💖

apps/twig/vite.main.config.mts
```diff
@@ -121,6 +131,177 @@ function copyClaudeExecutable(): Plugin {
...
+
+      // Download using curl (available on macOS/Linux, works in Node without extra deps)
+      await execFileAsync("curl", ["-fsSL", "-o", zipPath, SKILLS_ZIP_URL], {
+        timeout: 30_000,
+      });
+
+      // Extract
+      const extractDir = join(tempDir, "extracted");
+      await mkdir(extractDir, { recursive: true });
+      await execFileAsync("unzip", ["-o", zipPath, "-d", extractDir]);
```
joshsny: we should probably lean on packages here where we can, we will eventually want to support windows, but there are lots of cases of us doing similar things so it's not a blocker

PR #937, opened 2026-02-17
feat: cache logs locally

### LINKED ISSUE #570: Sometimes there is a delay on a new message showing up
E.g. I send a message, it doesn't appear immediately, but 2 seconds later

jonathanlab:
I _think_ this is an improvement over just directy querying S3, especially in situations where the connection is spotty. It's also the first step towards making the app available offline.

It might also resolve #570 (delay on user messages). But I'll keep that open until we confirm it's resolved

It also does a few other things:
- Makes workspace verification non-blocking
- Makes loading the logs and loading the agent two separate things (So UI does not block in the totally unthinkable situation where Anthropic somehow made an agent SDK that takes 3-4 seconds on average to spin up)

Also, if we don't have the logs cached locally on a device, we download them from S3 which stays source of truth


joshsny (commented):
this is maybe very naive, but can we cache trpc queries in the app more generally and persist those, in a similar way to what react query supports out of the box (but using a persistent cache rather than the default in memory one), rather than need to do something specific for logs?

PR #963, opened 2026-02-19
fix: More performance fixes + add dev-only timing instrumentation

charlesvien:


k11kirky (approved):
Are we going to keep this in for long? feels a little messy to wrap everything

PR #970, opened 2026-02-20
feat(twig): oauth versioning

tatoalo:
This adds OAuth scope versioning with an in-session re-authentication flow and also a quick guard to prevent scope/version drift.
When OAuth scopes change, existing users may keep tokens tied to old grants. Instead of forcing an immediate logout (bad UX) we now detect stale scope versions and prompt users to re-authenticate from within Twig

- Users with stale OAuth grants are guided to re-auth
- If re-auth cannot proceed, users can safely log out via the prompt
twig_reoauth


joshsny (approved):
noice

PR #1053, opened 2026-02-26
feat(twig): add user agent header to requests

robbie-c:
Made sure that requests are made with a hog.dev user agent.

We use this in the posthog API to add a `source` for `report_user_action()` calls, which lets us know how e.g. dashboards are created.

See:
* example dashboard https://us.posthog.com/project/2/insights/a5XEngMW?dashboard=636477&variables_override=%7B%7D&filters_override=%7B%7D&tile_filters_override=%7B%7D
* get_request_analytics_properties/get_event_source https://github.com/PostHog/posthog/blob/master/posthog/event_usage.py#L271

I made different user agents for desktop/mobile/cloud, so we can tell those apart too, if we need to

### DIFF
--- a/apps/mobile/src/features/conversations/api.ts
+++ b/apps/mobile/src/features/conversations/api.ts
@@ -1,42 +1,11 @@
 import { fetch } from "expo/fetch";
-import Constants from "expo-constants";
-import { useAuthStore } from "@/features/auth";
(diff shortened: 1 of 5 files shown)


robbie-c:
Easy enough to change! Just needs to be something :D

PR #1066, opened 2026-02-27
fix(auth): improved refresh token logic & token handling

skoob13:
## Problem

The PostHog MCP server disconnects after some time and can't recover. Apparently, the problem was a combination of factors:
- Refreshing of OAuth tokens was happening in the background even if we had active agent sessions. Since the OAuth token is deleted from the database when it is refreshed, it exposes two bugs:
  - The MCP server couldn't find the old token while the session was still in progress, so it would throw 401.
  - The Claude Code adapter was opening a second ghost session by retrieving tools using the MCP SDK, so Claude Code was actually using two connections (it explained why I saw both regular tools and mcp-cli commands in some screenshots). The ghost session was never cleaned up, so when Claude Code received the updated MCP credentials, it preferred to use an open session with a stale token (it's most likely a bug in CC).
- `mcp-cli` was too experimental and didn't handle reconnections. I'm not sure here 100% because I think it might have handled reconnections, but tools were injected into the context by the ghost session, so the agent was obviously preferring contextual. However, it was removed from the new version of CC☹️ 
- The OAuth token update was too close to expiration (5m). As it appears, it was not enough for most internal usage sessions, since we've received a few reports of broken connections to the MCP.
- Related PR https://github.com/PostHog/posthog/pull/49269 to update caching TTL of tokens, as the broken OAuth refresh was partially masked by the increased TTL of refreshed tokens.

## Changes

- Bumped token refresh from 5 minutes before expiration to 30 minutes. This gives enough time to find a window between active sessions and to propagate changes. 20 minutes might be fine, but sessions have become longer nowadays.
- Implemented smarter OAuth token refresh: If there are no active sessions, update. If there are active sessions, look for a window when they will all become inactive. Otherwise, update if we're very close to token expiration.
- Replaced the experimental MCP CLI with tool offloading. I'll additionally work on the MCP to increase the probability of tool discovery.
- Bumped CC to the new version, which required adding Sonnet 4.6: https://github.com/PostHog/posthog/pull/49310.
- Fixed the ghost MCP session (next PR in the stack).

## Tests

Feels like this requires an e2e test?

### DIFF
--- a/apps/twig/src/main/services/agent/service.ts
+++ b/apps/twig/src/main/services/agent/service.ts
@@ -374,11 +374,9 @@ export class AgentService extends TypedEventEmitter<AgentServiceEvents> {
   public hasActiveSessions(): boolean {
     for (const session of this.sessions.values()) {
       if (session.promptPending) {
-        log.info("Active session found", { sessionId: session.taskRunId });
         return true;
       }
     }
-    log.info("No active sessions found");
     return false;
   }
 
--- a/apps/twig/src/renderer/features/auth/stores/authStore.ts
+++ b/apps/twig/src/renderer/features/auth/stores/authStore.ts
@@ -118,36 +118,36 @@ async function attemptRefreshWithActivityCheck(
       return;
     }
 
+    // Refresh if there are no active sessions
     const hasActive = await trpcVanilla.agent.hasActiveSessions.query();
-
     if (!hasActive) {
       await getState().refreshAccessToken();
       return;
     }
 
-    // Sessions are busy — wait for the service to tell us they're idle
-    log.debug("Active sessions detected, waiting for idle event");
-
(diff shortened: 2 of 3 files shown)


skoob13:
* **#1067** <a href="https://app.graphite.com/github/pr/PostHog/Twig/1067?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1066** <a href="https://app.graphite.com/github/pr/PostHog/Twig/1066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/Twig/1066?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1081, opened 2026-03-01
fix: Surface api error details

okedeji:
I built and shared the [PostHog IDE Companion ](https://github.com/okedeji/posthog-ide-companion)with the team, then found out the VSCode extension direction was intentionally ditched for Twig. So I tried Twig out. OAuth flow, project pick, added a local repo, all good.

Then I sent my first message, but got an error: *"Failed to create task"* with zero detail. I popped open devTools (wasn't disabled in prod), saw a 403 in the network tab. Made sense since the project's still experimental, and I may just be bumped by internal ACL, but looks like we have a gap in error messaging not including readable details, could get confusing for users in a proper prod release without devTools

**What I changed:**

- Added an `ApiError` class in `fetcher.ts` with `status`, `body`, and `detail`. Follows the existing pattern (`LlmGatewayError`, `TypedStatusError`, etc.)
- Added optional `cause` to `SagaResult` in `saga.ts`. The saga was catching errors and flattening them to strings, losing all structure
- When the cause is an `ApiError`, the toast now shows the actual `detail` as description. Also fixed the toast import to use `@utils/toast` instead of importing `sonner` directly
- Set `devTools: isDev` in `window.ts`. Was defaulting to enabled. If that was intentional (for prod inspection in this experimental stage), happy to drop this commit

Other areas (`InboxSignalsTab`, `WorkspaceSetupPrompt`, `TaskLogsPanel`, `SessionService`) have similar generic error messages but I kept this PR focused on the path I could reproduce and verify. Easy fix if you'd like me to follow up.

I mean, if you guys are generous enough to grant me access, would love to keep having fun poking around and help close out some other pending issues.

**Test plan:**

- Create a task that triggers an API error, toast shows step label + API detail
- Create a task successfully, no change
- devTools inaccessible in packaged builds
- `pnpm typecheck` passes
- `pnpm lint` passes

Screenshot 2026-03-02 at 00 52 19


k11kirky:
@okedeji Can you email me @ peter@posthog.com and I can get you set up with access to the Alpha

PR #1091, opened 2026-03-03
feat: Upgrade Claude Code adapter (see desc)

charlesvien:
Closes https://github.com/PostHog/Twig/issues/1059
Closes https://github.com/PostHog/Twig/issues/1052

Hopefully next time this will be easier with agents. I created an [UPSTREAM.md](http://UPSTREAM.md) in the claude code adapter that has some useful information on our fork and upgrade points and upgrade notes from my agent sessions, so that future sessions can absorb some initial context for the upgrade.

Sync Claude Code adapter with upstream v0.19.2 (from v0.10.9 fork point). We are nearly 10 minor versions and countless patch versions out of date. I purposefully tried to make the code read the same in the hot paths as upstream.

What do we gain from this?
1. Add session management support (list, fork, resume, replay)
2. Add token usage tracking with accumulated usage notifications
3. Add SettingsManager for file-level permission rules via PreToolUse hook
4. Add prompt queueing and pending message ordering
5. Add terminal output streaming
6. Add edit tool diff rendering (structuredPatch parsing)
7. Many bug fixes....

### DIFF
--- a/packages/agent/src/adapters/claude/permissions/permission-handlers.ts
+++ b/packages/agent/src/adapters/claude/permissions/permission-handlers.ts
@@ -448,11 +448,11 @@ export async function canUseTool(
     return planFileResult;
   }
 
-  if (session.permissionMode === "dontAsk") {
-    const message = "Tool not pre-approved. Denied by dontAsk mode.";
-    await emitToolDenial(context, message);
-    return { behavior: "deny", message, interrupt: false };
-  }
+  // if (session.permissionMode === "dontAsk") {
+  //   const message = "Tool not pre-approved. Denied by dontAsk mode.";
+  //   await emitToolDenial(context, message);
+  //   return { behavior: "deny", message, interrupt: false };
+  // }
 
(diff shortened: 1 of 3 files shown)


apps/twig/src/main/services/agent/service.ts
```diff
@@ -577,42 +577,34 @@ export class AgentService extends TypedEventEmitter<AgentServiceEvents> {
           credentials,
           customInstructions,
         );
-        const resumeResponse = await connection.extMethod(
-          "_posthog/session/resume",
-          {
+        const resumeResponse = await connection.unstable_resumeSession({
```
k11kirky: only worry on this is some of this stuff is only compatible with the v2 and thats why its unstable_ but if this is working maybe i am wrong :shrug:

jonathanlab:
### Merge activity

* **Mar 4, 8:58 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/Twig/1091).

PR #1134, opened 2026-03-06
chore(agent): make PRs drafts by default

joshsny:
When the cloud agent opens a PR, we aren't yet confident this change is actionable or the fix is correct, we should default to it creating draft PRs here.


tatoalo (approved):
yesss

PR #1143, opened 2026-03-06
feat(twig): use base branch for changes panel diff comparison

tatoalo:
added `baseBranch` from task run state through to the gh compare API so the Changes panel diffs against the user-selected branch instead of always comparing against the repo default branch.


charlesvien:
### Merge activity

* **Mar 6, 9:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/Twig/1143).

Issue #642, opened 2026-01-27
Navigating prompt history is annoying with multiline inputs


jonathanlab:
>nitpicking but when pressing the up arrow to get previous prompts in the prompting window, if there's a multiline prompt the text cursor has to go through the lines before giving you the previous prompt. Felt weird to me when my muscle memory from terminal to get 3 prompts back ended up sticking in a multiline commit prompt

most agent CLIs also have this for multiline inputs, but I definitely agree there should be a better solution here. What I usually tend to do is press home then arrow up. Maybe something like shift+up  or another modifier + up would be nicer there.

PR #1170, opened 2026-03-10
feat: Add Shift+Up/Down for prompt history navigation

charlesvien:
Closes https://github.com/PostHog/code/issues/642

1. Add Shift+Up/Down keyboard shortcuts to navigate prompt history from any cursor position
2. Show history hint (⇧↑↓) in the message editor footer when a task is active
    1. The hint hides itself when the user types in the message editor.
3. Register prompt-history-prev/next in the keyboard shortcuts list


charlesvien:
> Happy - but maybe this clutters the ui a bit - would prefer we hide by default as this is an unlikely action

Kk I removed it, power users can find it in the keyboard shortcut sheet with the rest of the shortcuts

PR #1209, opened 2026-03-12
feat: Optimistically render conversation items before subscription confirms

charlesvien:
Adds a generic optimisticItems mechanism to the session store so any conversation item type can be displayed immediately before the real event arrives from the main process. Currently used for user messages, but the OptimisticItem union is designed to be extended for future use cases.

1. Add OptimisticItem type and optimisticItems array to AgentSession store
2. Append an optimistic user message when a prompt is sent, before the server round-trip
3. Clear optimistic items when the real session/prompt event arrives or on error
4. Merge optimistic items into the conversation view between confirmed items and queued messages
5. Add useOptimisticItemsForTask hook and store setters (appendOptimisticItem, clearOptimisticItems, appendEventAndClearOptimisticItems)

### DIFF
--- a/apps/code/src/renderer/features/sessions/service/service.test.ts
+++ b/apps/code/src/renderer/features/sessions/service/service.test.ts
@@ -52,6 +52,7 @@ const mockSessionStoreSetters = vi.hoisted(() => ({
   clearAll: vi.fn(),
   appendOptimisticItem: vi.fn(),
   clearOptimisticItems: vi.fn(),
+  replaceOptimisticWithEvent: vi.fn(),
 }));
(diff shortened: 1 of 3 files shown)


charlesvien:
### Merge activity

* **Mar 12, 10:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/code/1209).
* **Mar 12, 10:51 PM UTC**: [Graphite](https://app.graphite.com/github/pr/PostHog/code/1209) rebased this pull request as part of a merge.

Issue #1201, opened 2026-03-11
Trim <system-reminder> from read prompt


jonathanlab:
See `tool-use-to-acp.ts` and

<img width="1425" height="699" alt="Image" src="https://github.com/user-attachments/assets/ac02c3ef-e4da-482f-acfa-241835e574a0" />

PR #1214, opened 2026-03-12
Strip system reminder from Claude tool use responses

jonathanlab:
**TL;DR:** Replace hardcoded system reminder string matching with a regex-based function to more reliably remove system reminders from Claude tool use responses.


jonathanlab:
Closes #1201

PR #1248, opened 2026-03-15
feat(code): add /clear command to reset conversation history

littlekirkycode:
Screenshot 2026-03-15 at 15 48 43

 ## Summary

  Adds a `/clear` slash command that wipes conversation history and
  reconnects with a fresh session. Useful when context is cluttered or the
  agent is stuck in a loop.

  ### Implementation

  **Client-side command** in the message editor — delegates to
  `SessionService.resetSession()`:

  ```ts
  {
    name: "clear",
    description: "Clear conversation history and start fresh",
    async execute(_args, ctx) {
      const { getSessionService } = await
  import("@features/sessions/service/service");
      await getSessionService().resetSession(ctx.taskId, ctx.repoPath);
      toast.success("Conversation cleared");
    },
  }
  ```

  **`resetSession()`** calls `reconnectInPlace` with a new `clearHistory`
  flag. When true, log replay is skipped entirely — the session reconnects
  with empty event history instead of replaying from the log file:

  ```ts
  const prefetchedLogs = clearHistory
    ? { rawEntries: [], sessionId: undefined, adapter: undefined }
    : await this.fetchSessionLogs(logUrl, taskRunId);
  ```

  **SDK's built-in `/clear` is suppressed** — it doesn't work in our ACP
  context, so it's added to `UNSUPPORTED_COMMANDS` and replaced by our own.

  ### Changes

  | File | What |
  |------|------|
  | `commands.ts` (message-editor) | New `/clear` command definition |
  | `commands.ts` (session) | Add `"clear"` to `UNSUPPORTED_COMMANDS` |
  | `service.ts` | `resetSession()` passes `clearHistory=true`, `reconnectInPlace` skips log fetch when clearing |

  ## Test plan

  - [x] Type `/clear` in message editor → conversation history is cleared
  - [x] Session reconnects with clean state (no old messages replayed)
  - [x] Toast confirms "Conversation cleared"
  - [x] `/clear` appears in slash command autocomplete


apps/code/src/renderer/features/message-editor/commands.ts
```diff
@@ -51,6 +51,21 @@ const commands: CodeCommand[] = [
...
+    name: "clear",
+    description: "Clear conversation history and start fresh",
+    async execute(_args, ctx) {
+      if (!ctx.repoPath || !ctx.taskId) {
+        toast.error("Cannot clear: no active session");
+        return;
+      }
+      const { getSessionService } = await import(
+        "@features/sessions/service/service"
+      );
```
jonathanlab: nit: no inline imports please

littlekirkycode:
Both changes applied — removed the inline import and rewrote resetSession() to tear down the old session and create a fresh task run via the suggested createNewLocalSession(), so old messages won't come back after a restart.

Also rebased #1249 on top of this as its clear-and-continue flow goes through resetSession().

PR #1257, opened 2026-03-16
fix(code): remove archived tasks from command center

adboio:
when a task is archived, it should be removed from the command center automatically


charlesvien (approved):
LGTM

PR #1327, opened 2026-03-25
fix(code): Default to action button in discard changes dialog

jurajmajerik:
## Problem
When discarding changes, you have to click the action button every time.

## Changes
Default to the action button instead so that I can quickly hit Enter, same as VS Code does it.

|Before|After|
|----|----|
|image|image|


charlesvien (approved):
LGTM

PR #1359, opened 2026-03-30
Set explicit gcTime on QueryClient

Twixes:
## Problem

We set `staleTime` globally but never set `gcTime`. TanStack Query's default gcTime is 5 min which sounds fine, but queries that override with `staleTime: Infinity` (we have a bunch) stay "fresh" forever, meaning the GC never kicks in for them even when they go inactive.

## Changes

Set an explicit `gcTime: 30min` at the client level so inactive query data always gets cleaned up regardless of staleTime. 30 min is generous enough for a long-running desktop app where you're switching between tasks/files, but prevents multi-hour sessions from hoarding stale data indefinitely.

### DIFF
--- a/apps/code/src/renderer/utils/queryClient.ts
+++ b/apps/code/src/renderer/utils/queryClient.ts
@@ -4,6 +4,7 @@ export const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       staleTime: 1000 * 60 * 5,
+      gcTime: 1000 * 60 * 30,
       refetchOnWindowFocus: true,
     },
   },


jonathanlab (commented):
afaik `gcTime` and `staleTime` are fully independent. Garbage collection is only triggered once a query has zero observers right? So if a query with `staleTime: infinity` has no observers, it gets garbage collected after 5 minutes. If it has observers, it'll just stay around infinitely. 

I do not see how bumping it to 30 minutes helps in this scenario, as it would only increase cache retention, but I might be wrong :)

PR #1390, opened 2026-03-31
fix: pass custom instructions to agent when rejecting tool permissions

### LINKED ISSUE #1368: Custom instruction being completely ignored on "Type here to tell is what to do differently"
The agent tried to install some packages I wasn't sure about. I tried to give it some more context and directions, but it just ignored it all and kept saying that "I was not allowing the shell command to complete". I wasn't, indeed, but I was sending custom instructions.

The 2 messages at the bottom at the conversation were the instructions that most likely got queued as messages and got sent after the agent gave up on shell permission.

Image

CaioDGallo:
## Problem

- When rejecting tool permission with custom instructions, the agent was not receiving the custom input. Because of this, it completely ignored any instructions passed and sent them later as a queued message.

Closes #1368

## Changes

- Route reject-with-instruction custom input through the permission response instead of sending it as a separate prompt
- Extract and include user feedback in the tool denial message so the agent can act on it

## How did you test this?

- Manually

image

image

### DIFF
--- a/apps/code/src/renderer/features/sessions/components/SessionView.tsx
+++ b/apps/code/src/renderer/features/sessions/components/SessionView.tsx
@@ -252,7 +252,10 @@ export function SessionView({
       }
 
       if (customInput) {
-        if (isOtherOption(optionId)) {
+        if (
+          isOtherOption(optionId) ||
+          selectedOption?._meta?.customInput === true
+        ) {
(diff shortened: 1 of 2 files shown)


CaioDGallo:
yeah, I use custom instructions a lot, so it was really annoying also. Happy to help!

PR #1434, opened 2026-04-01
feat(code): persist PR URLs to backend & emit file activity events

adboio:
## Problem

when agents create PRs, we detect the URLs from bash output and store them on the backend

but when created via UI in posthog code, we do no such thing

also, we generally need a better way to correlate local tasks to branches (and therefore PRs)

## Changes

when PRs are created via UI, call `updateTaskRun` with the PR URL

i accidentally also combined this with the change to emit events on file changes (which helps build a "loose" task<>branch association)

## How did you test this?

created PR via UI, verified task in db:

```
{"pr_url": "https://github.com/PostHog/posthog/pull/53123"}
```


adboio:
### Merge activity

* **Apr 3, 12:56 AM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/1434).

PR #1466, opened 2026-04-02
feat: Add initial structured output for agent

ryans-posthog:
## Problem
Currently, users cannot execute a AI Task and expect a structured output from it. 

## Changes
The PR adds a new tool for accepting and verifying structured output and wires it to fire on the `onOutput` event. Currently it is only available for Anthropic Claude based executions. The implementation for Codex will follow as it requires a more complex method of interaction.

## How did you test this?
Build the agent and start the Posthog system locally. Rebuild the agent docker image with the new code and execute a task via the console with a json schema set. Verify that the task calls the structured output endpoint and the task completes successfully.


jonathanlab (commented):
The Claude SDK supports structured outputs via `output_config`, which can even be typed via Zod. Could we use that instead? Then we also would not need an output MCP server right? Also note the possibility to use `strict: true`. 


Source: https://platform.claude.com/docs/en/build-with-claude/structured-outputs

joshsny (approved):

packages/agent/src/adapters/claude/structured-output/create-output-server.ts
joshsny, line 1: @ryans-posthog i am a little confused why we need the MCP server, as I thought we were using an SDK property, could you elaborate on that at all?

joshsny: stamped so as not to block, if this makes sense, feel free to merge it 👍

ryans-posthog: Oh right, I didn't remove it because I was testing out the SDK based communication, I'll remove that and get this merged.

PR #1487, opened 2026-04-03
feat: Harden CodexAcpAgent session state and permission handling

charlesvien:
## Changes

1. Initialize sessionState and codexConnection eagerly in the constructor instead of using ! assertions
2. Pass permissionMode, taskRunId and taskId from meta into createSessionState for all session creation paths (load, new, fork)
3. Fix double-counting of token usage by relying on sessionUpdate notifications instead of also adding response.usage
4. Remove redundant cancel() override, use inherited interrupt() and base class cancellation state
5. Use POSTHOG_NOTIFICATIONS.USAGE_UPDATE constant instead of hardcoded string
6. Escape newlines in system instructions passed to codex CLI args
7. Abort the session's AbortController on closeSession for proper async cleanup

## How did you test this?

Manually

### DIFF
--- a/packages/agent/src/acp-extensions.ts
+++ b/packages/agent/src/acp-extensions.ts
@@ -64,4 +64,7 @@ export const POSTHOG_NOTIFICATIONS = {
 
   /** Marks a boundary for log compaction */
   COMPACT_BOUNDARY: "_posthog/compact_boundary",
+
+  /** Token usage update for a session turn */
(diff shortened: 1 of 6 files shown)


k11kirky:
### Merge activity

* **Apr 6, 9:47 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/code/1487).

PR #1513, opened 2026-04-07
feat(sig): add modifying report state via inbox

oliverb123:
## Problem

We want to be able to snooze, suppress, delete reports via the PHC inbox. We also want to be able to reingest them, although this is gated behind 1: being a staff user and 2: running a dev build.

This PR also contains some futzing with the keyboard handling, because it wasn't as good as I wanted it to be


Twixes (approved):
A few comments from me, but as we've chatted in Slack, each I tackled in the PR stacked on top

PR #1514, opened 2026-04-07
chore(cloud-agent): extract shared utilities for cloud attachments

tatoalo:
[ph_code_cloud_attachments.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/727aaffd-ea15-4c0d-b46a-4f328d7c2e11.mp4)

## Problem

Cloud attachments feature needs shared utilities across multiple modules

## Changes

- Extract `escapeXmlAttr`/`unescapeXmlAttr` into `xml.ts` (was duplicated in 3 files)
- Extract `getFileName`/`getFileExtension`
- Extract `extractPromptDisplayContent`, `makeAttachmentUri`, `parseAttachmentUri`
- Move cloud prompt codec (`serializeCloudPrompt`, `deserializeCloudPrompt`) into `@posthog/shared`
- Fix mid-file import in `content.ts`

## How did you test this?

Unit tests for cloud prompt codec and prompt content extraction

PR #1515, opened 2026-04-07
feat(cloud-agent): add cloud prompt builder and file persistence

tatoalo:
## Problem

Cloud tasks need to embed local file attachments (text + images) as ACP ContentBlocks for the agent server.

## Changes

- Add `cloud-prompt.ts`: builds ACP ContentBlock arrays from user prompts with file tags, handles text truncation (100K chars) and image size limits (5MB), supported format detection
- Add `persistFile.ts`: shared browser file persistence with chunked base64 encoding (replacing O(N^2) `btoa(reduce)` pattern)
- Add `os.ts`: `selectFiles` tRPC procedure (native Electron file picker), `createClipboardTempFilePath` helper to preserve original filenames via `mkdtemp`

**Stack: 2/4** — depends on #1514

### DIFF
--- a/apps/code/src/main/trpc/routers/os.ts
+++ b/apps/code/src/main/trpc/routers/os.ts
@@ -39,6 +39,7 @@ const expandHomePath = (searchPath: string): string =>
 
 const MAX_IMAGE_DIMENSION = 1568;
 const JPEG_QUALITY = 85;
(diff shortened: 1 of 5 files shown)


tatoalo:
Recreating with clean branch names

PR #1532, opened 2026-04-07
feat(sig): Report list toolbar as core control

Twixes:
## Problem

The inbox list toolbar introduced in #1513 added controls _only_ for bulk operations. But the same controls have been missing for _individual_ reports too. In the Cloud version, we had them under a dropdown in the report pane itself. I considered doing the same here, but there's no point when we have a great toolbar - the toolbar should be the individual report controls as well.

At the same time, checkboxes take away from the content when shown always - and we don't need them by default for _most_ user interactions here. (Ofc will validate this, but at least I think that's a safe assumption for now that the individual report level is the primary one.)

## Changes

Making the report list toolbar always-on, with intuitive logic for selection state synced with the currently-viewed-report state. Unifies the selection state, so that "currently viewed" and "selected" are the same thing. Checkboxes only show up when multiple reports are selected. For multi-select, you can use the "Select all" checkbox or standard desktop Shift/Cmd+click patterns. See in action:

### DIFF
--- a/apps/code/src/main/services/git/create-pr-saga.ts
+++ b/apps/code/src/main/services/git/create-pr-saga.ts
@@ -72,21 +72,21 @@ export class CreatePrSaga extends Saga<CreatePrSagaInput, CreatePrSagaOutput> {
     let { commitMessage, prTitle, prBody } = input;
 
     if (input.branchName) {
+      const branchName = input.branchName;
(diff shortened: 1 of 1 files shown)


Twixes:
### Merge activity

* **Apr 8, 10:43 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/code/1532).

PR #1559, opened 2026-04-08
feat(cloud): cloud task SSE streaming

tatoalo:
## Problem

https://github.com/user-attachments/assets/afa63e15-f088-4cf2-b2c8-8add09ca9f67

Cloud task log viewing used polling (500ms when viewing, 60s if in background) which was inefficient, had dedup complexity, and couldn't stream status changes in realtime. Also it was kinda bad. 
We also had missing comms when a remote sandbox was torn down before the terminal status event arrived, which resulted in ph-code to nevel learn about the run being completed and always show a broken "in progress" state which was not the reflection of the real one.

Added a ton of tests just to make sure this is robust and covers some edge cases.
Diid a lot of manual testing, would appreciate a couple of eyes testing this as well please 🙏🏻 

(https://github.com/PostHog/posthog/pull/53754)

## Changes

- Complete rewrite of `CloudTaskService` from polling to SSE streaming with bootstrap (fetch history + open SSE with `start=latest`), exponential backoff reconnect, and `handleStreamCompletion` that checks actual run status on any stream error
- aadded an sse parse to parse strem chunks with support for `id`, `event`, and multi-line `data` fields
- a bit of refactor to extract `TaskRunStatus`, `TERMINAL_STATUSES`, and `isTerminalStatus()` to `@shared/types` 
- rewrote `watchCloudTask` to pass `apiHost`/`teamId` directly, added `retryCloudTaskWatch`, and `handleCloudTaskUpdate` now handles SSE error events (moved the auth concern to the renderer which already has the auth state, keeping the main-process service simpler)
- sidebar now reads `cloudStatus` from session store (updated via SSE) instead of waiting for API task refetch
- status rename `"started"` → `"queued"` to match BE API

### DIFF
--- a/apps/code/src/main/services/cloud-task/service.ts
+++ b/apps/code/src/main/services/cloud-task/service.ts
@@ -8,18 +8,45 @@ import type { AuthService } from "../auth/service";
 import {
   CloudTaskEvent,
   type CloudTaskEvents,
+  isTerminalStatus,
   type SendCommandInput,
   type SendCommandOutput,
   type TaskRunStatus,
-  TERMINAL_STATUSES,
   type WatchInput,
 } from "./schemas";
+import { type SseEvent, SseEventParser } from "./sse-parser";
 
 const log = logger.scope("cloud-task");
 
-const LOG_POLL_INTERVAL_MS = 500;
-const STATUS_POLL_INTERVAL_MS = 60_000;
-const STATUS_POLL_INTERVAL_VIEWING_MS = 3_000;
+const MAX_SSE_RECONNECT_ATTEMPTS = 5;
+const SSE_RECONNECT_BASE_DELAY_MS = 2_000;
+const SSE_RECONNECT_MAX_DELAY_MS = 30_000;
+const EVENT_BATCH_FLUSH_MS = 16;
+const EVENT_BATCH_MAX_SIZE = 50;
+const SESSION_LOG_PAGE_LIMIT = 5_000;
+
+interface SessionLogsPage {
+  entries: StoredLogEntry[];
+  hasMore: boolean;
+}
+
+interface CloudTaskConnectionError {
+  title: string;
+  message: string;
+  retryable: boolean;
+  autoRetry?: boolean;
+}
+
+class CloudTaskStreamError extends Error {
+  constructor(
+    message: string,
+    public readonly details: CloudTaskConnectionError,
+    public readonly status?: number,
+  ) {
+    super(message);
+    this.name = "CloudTaskStreamError";
+  }
+}
 
 interface TaskRunResponse {
   id: string;
@@ -28,6 +55,19 @@ interface TaskRunResponse {
   output?: Record<string, unknown> | null;
   error_message?: string | null;
   branch?: string | null;
+  updated_at?: string;
+  completed_at?: string | null;
+}
+
+interface TaskRunStateEvent {
+  type: "task_run_state";
+  status?: TaskRunStatus;
+  stage?: string | null;
+  output?: Record<string, unknown> | null;
+  error_message?: string | null;
+  branch?: string | null;
+  updated_at?: string | null;
+  completed_at?: string | null;
 }
 
 interface WatcherState {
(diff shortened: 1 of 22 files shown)


apps/code/src/main/services/cloud-task/service.ts
```diff
@@ -28,31 +55,123 @@ interface TaskRunResponse {
...
+        {
+          title: "Cloud stream failed",
+          message: `The cloud stream request failed with status ${status}. Retry to reconnect.`,
+          retryable: true,
+          autoRetry: true,
+        },
+        status,
+      );
+  }
+}
```
jonathanlab, line 186: This is great, love the clear error handling

apps/code/src/main/services/cloud-task/service.ts
```diff
@@ -392,34 +659,291 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
...
+        taskId: watcher.taskId,
+        runId: watcher.runId,
+        kind: "logs",
+        newEntries: dedupedEntries,
+        totalEntryCount: watcher.totalEntryCount,
+      });
+    }
+
+    watcher.bufferedLogBatches = [];
+  }
```
jonathanlab, line 762: I wonder if deduplicating by exact content match is the right move or if we should rely purely on the SSE id here..? Not sure, you might have more context than I do.

apps/code/src/main/services/cloud-task/sse-parser.ts
```diff
@@ -0,0 +1,84 @@
...
+        id: this.currentEventId ?? undefined,
+        data,
+      };
+    } catch {
+      return null;
+    } finally {
+      this.currentEventName = null;
+      this.currentEventId = null;
+    }
+  }
```
jonathanlab, line 88: We probably want some log on parse failure here

apps/code/src/main/services/cloud-task/service.ts
```diff
@@ -222,165 +355,299 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
...
+
+  private handleSseEvent(key: string, event: SseEvent): void {
+    const watcher = this.watchers.get(key);
+    if (!watcher || watcher.failed) return;
+
+    if (event.id) {
+      watcher.lastEventId = event.id;
+    }
+
+    if (event.event === "error") {
```
jonathanlab, line 648: I see a lot of type casting going on. Could we use a type guard or a Zod schema for this? It makes all this message handling pretty hard to read

apps/code/src/main/services/cloud-task/service.ts
```diff
@@ -222,165 +355,299 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
...
+      const message =
+        typeof event.data === "object" &&
+        event.data !== null &&
+        "error" in event.data &&
+        typeof (event.data as { error?: unknown }).error === "string"
```
jonathanlab: I see a lot of type casting going on. Could we use a type guard or a Zod schema for this? It makes all this message handling pretty hard to read

apps/code/src/renderer/features/sidebar/components/items/TaskItem.tsx
```diff
@@ -203,6 +199,12 @@ export function TaskItem({
   const isWorktreeTask = workspaceMode === "worktree";
   const isCloudTask = workspaceMode === "cloud";
 
+  const isTerminalCloud =
+    isCloudTask &&
+    (taskRunStatus === "completed" ||
+      taskRunStatus === "failed" ||
+      taskRunStatus === "cancelled");
+
```
jonathanlab: We can just use the helper you created for this instead

apps/code/src/shared/types.ts
```diff
@@ -104,6 +124,9 @@ export interface CloudTaskUpdatePayload {
   output?: Record<string, unknown> | null;
   errorMessage?: string | null;
   branch?: string | null;
+  // Connection/error fields (present when kind is "error")
+  errorTitle?: string;
+  retryable?: boolean;
 }
```
jonathanlab: nit: You could probably turn `CloudTaskUpdatePayload` into a discriminated union to handle all these different cases.

jonathanlab (approved):
Great stuff. I left mostly nits that would help with keeping the code maintainable/readable.

apps/code/src/main/services/cloud-task/service.ts
tatoalo: Great point! So in this precise case it is intentional, basically the IDs we work with are redis stream IDs and they are assigned when those events enter the caching system, the historical entries come from S3 and have no stream IDs so the json payload is the only shared key between the sources

jonathanlab: ah nice, thanks for explaining

PR #1579, opened 2026-04-09
fix: improve inline code rendering in signal details

### LINKED ISSUE #1567: render markdown properly in signals details
markdown is not rendered properly, see backticks here:

Image

reverb256:
### Problem
Inline code (backtick content) in signal card details is barely visible. The `Code` component uses `variant="ghost"` which has no background, and the parent container applies `color: var(--gray-11)` which makes inline code blend into surrounding text.

### Fix
Added explicit CSS rules for `code` elements in the signal card body:
- `bg-gray-4` — subtle background to distinguish code from text
- `rounded` — rounded corners
- `px-1 py-0.5` — padding for visual separation
- `font-mono` — explicit monospace font

This makes inline code clearly distinguishable from surrounding text while maintaining the compact signal card layout.

Fixes #1567

### DIFF
--- a/apps/code/src/renderer/features/inbox/components/detail/SignalCard.tsx
+++ b/apps/code/src/renderer/features/inbox/components/detail/SignalCard.tsx
@@ -257,7 +257,7 @@ function CollapsibleBody({ body }: { body: string }) {
   return (
     <Box>
       <Box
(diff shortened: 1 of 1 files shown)


adboio (approved):
thank you! :heart:

PR #1592, opened 2026-04-09
chore(code): github branch picker loading all branches

VojtechBartos:
## Problem

  The branches API was capped at 2000 branches. Repositories with more than that were missing branches in the selector.

  ## Changes

  - Paginate the branches API call with offset/limit, fetching all pages progressively
  - Show the first page of branches immediately while the rest load in the background
  - Display a "Loading more" indicator with running count in the branch dropdown
  - Show a spinner on the branch selector trigger while pages are still loading
  - Disable the branch selector until the first page arrives

## Prerequisite

Posthog API changes https://github.com/PostHog/posthog/pull/53931

## Showcase 

https://github.com/user-attachments/assets/401d0235-12cd-4a73-bbf7-a571bd17ca72

### DIFF
--- a/apps/code/src/renderer/hooks/useIntegrations.ts
+++ b/apps/code/src/renderer/hooks/useIntegrations.ts
@@ -6,7 +6,7 @@ import {
   useIntegrationStore,
 } from "@features/integrations/stores/integrationStore";
 import { useQueries } from "@tanstack/react-query";
-import { useCallback, useEffect, useMemo } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { useAuthenticatedInfiniteQuery } from "./useAuthenticatedInfiniteQuery";
 import { useAuthenticatedQuery } from "./useAuthenticatedQuery";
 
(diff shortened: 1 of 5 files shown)


apps/code/src/renderer/hooks/useIntegrations.ts
```diff
@@ -67,18 +68,63 @@ function useAllGithubRepositories(githubIntegrations: Integration[]) {
...
+        if (!lastPage.hasMore) return undefined;
+        return allPages.reduce((n, p) => n + p.branches.length, 0);
+      },
+      staleTime: 0,
     },
-    { staleTime: 0, refetchOnMount: "always" },
   );
+
+  // Auto-fetch remaining pages in background once the first page arrives
+  useEffect(() => {
```
adboio: can we abort this query once the user selects a branch?

or if we keep this and fetch everything even after the user is done, we should at least be caching the results (see global comment)

VojtechBartos:
@adboio soooo :) ...

- we can fetch max 100 GH branches per page :(
- each page fetch takes ~400-500ms from my local, i hope it's going to be faster on prod, bud still
- for each page we were fetching default branch, not efficient

the changes i have made
- Posthog Code
	- we fetch first page with limit 100 so it's disabled fast and rest with limit 1000
	- we won't show spinner once collapsed
	- we will show only spinner when expanded
	- when you close/choose branch we pause fetching and resume when you open it again
	- changed query data stale time to default 5mins
- Posthog Backend
	- added caching for the fetching default branch (we can add branches fetching at some point in future if it's really pain)

WDYT?

skoob13 (approved):
LGTM. Can we add the search? How would we find master for PostHog's monorepo?

VojtechBartos:
@skoob13 there is UI based search, so you are going to find it. There is a small issue related to that on API side, so fixing it

PR #1595, opened 2026-04-09
feat(code): enable branch linking for local tasks

adboio:
## Problem

to build a good task wrap-up flow, we need to know which branch belongs to a given task

## Changes

adds automatic branch linking for local tasks. no UI/UX around this yet, nothing uses this data yet, just storing it

### agent scenarios:

| scenario | current linkedBranch | agent editing on... | result |
| --- | --- | --- | --- |
| agent edits on feature branch | null | feat-a | linkedBranch = feat-a |
| agent edits on same linked branch | feat-a | feat-a | no change |
| agent switches to a new feature branch | feat-a | feaet-b | linkedBranch = feat-b |
| agent edits on default branch | feat-a | main | no change |
| agent edits on default branch, never linked | null | main | no change |

### ui scenarios:

| scenario | result |
| --- | --- |
| "new branch" button | linkedBranch = new branch |
| "create PR" flow creates a branch | linkedBranch = new branch |
| "create PR" flow on existing branch | linkedBranch = new branch |
| branch creation fails | no change |
| PR creation fails | no change |

## How did you test this?

manually

PR #1596, opened 2026-04-09
feat(code): warn on local task branch mismatch

adboio:
## Problem

we want task isolation, but we can only do so much with local tasks, since they are simply not in an isolated environment

downstack #1594 + #1595 add automatic branch linking to tasks, so when an agent touches a file, or an agent/human creates a branch from posthog code, we associate that branch to that task

now, we need a way to keep the user in the loop on what's happening, and steer them in the right direction

## Changes

adds a UI warning if you send a prompt in a task that has a linked branch, but that branch is not currently checked out

details:

- hooks into tiptap with a new `onBeforeSend`so prompt is never lost if you cancel
- the warning is "blocking" in the sense that you have to make a decision. you can:
    - "continue anyway" -> ignore the warning, agent works in your current branch
        - automatic branch linking will still occur and maybe overwrite the current link, per the standard rules (see #1595 )
    - "switch branch" - checks out the task's linked branch
        - warns if there are uncommitted changes on current branch. this does not block, and surfaces errors from the switch (if any) in the dialog

## How did you test this?

manually

### DIFF
--- a/apps/code/src/main/services/workspace/service.ts
+++ b/apps/code/src/main/services/workspace/service.ts
@@ -369,6 +369,8 @@ export class WorkspaceService extends TypedEventEmitter<WorkspaceServiceEvents> 
       branchName,
     });
     trackAppEvent("branch_linked", {
+      task_id: taskId,
+      branch_name: branchName,
       source: source ?? "unknown",
     });
     log.info("Linked branch to task", { taskId, branchName, source });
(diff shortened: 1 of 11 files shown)


adboio:
* **#1596** <a href="https://app.graphite.com/github/pr/PostHog/code/1596?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/1596?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1595** <a href="https://app.graphite.com/github/pr/PostHog/code/1595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1594** <a href="https://app.graphite.com/github/pr/PostHog/code/1594?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jonathanlab (approved):
lgtm

PR #1597, opened 2026-04-09
fix: hyphenate wrapped text in agent thinking output

reverb256:
## Problem

Agent thinking output wraps long words mid-word without hyphenation marks, making it hard to read where words break.

Fixes #1593

## Solution

Changed the `<pre>` className from `break-all` to `break-words hyphens-auto` in `ThoughtView.tsx`. This enables CSS hyphenation with proper hyphen marks at word breaks.

- `break-words` prevents mid-word breaks without indication
- `hyphens-auto` adds proper hyphenation marks (the `<html lang="en">` attribute already enables English hyphenation rules)

## Screenshot

Before: words break mid-character with no indication
After: words wrap with proper hyphen marks (e.g., `hyphe-` at line end)

## Files Changed

- `apps/code/src/renderer/features/sessions/components/session-update/ThoughtView.tsx` — 1 line changed


adboio (approved):

adboio:
### Merge activity

* **Apr 9, 9:02 PM UTC**: [Graphite](https://app.graphite.com/github/pr/PostHog/code/1597) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

PR #1601, opened 2026-04-09
fix(mcp-store): disable MCPs from local config file

cvolzer3:
## Problem

Codex sessions in PostHog Code were inheriting `[mcp_servers.*]` entries from the user's personal `~/.codex/config.toml`, causing conflicts with the servers we intentionally supply via the MCP store.

## Changes

- Manually process the `mcp_servers.*` entries, and pass a command to disable each individually when spawning `codex-acp`
- Minor changes to `CodexSettingsManager` to load settings upon initialization

## How did you test this?

- Ran existing tests
- Tested locally

### DIFF
--- a/packages/agent/src/adapters/codex/settings.ts
+++ b/packages/agent/src/adapters/codex/settings.ts
@@ -67,7 +67,7 @@ export class CodexSettingsManager {
   }
 
   dispose(): void {
-    // No-op: no resources to release. Kept async to satisfy the BaseSettingsManager interface.
+    // No-op: no resources to release. Kept to satisfy the BaseSettingsManager interface.
   }
 }
 
--- a/packages/agent/src/adapters/codex/settings.ts
+++ b/packages/agent/src/adapters/codex/settings.ts
@@ -13,6 +13,8 @@ export interface CodexSettings {
   personality?: string;
   modelReasoningEffort?: string;
   trustLevel?: string;
+  // Names of every `[mcp_servers.<name>]` section declared in the user's config.toml
+  mcpServerNames: string[];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/agent/src/adapters/codex/settings.ts
```diff
@@ -24,32 +26,29 @@ export interface CodexSettings {
...
+    this.loadSettings();
   }
 
   async initialize(): Promise<void> {
-    if (this.initialized) {
-      return;
-    }
-    await this.loadSettings();
-    this.initialized = true;
+    // No-op: settings are loaded in the constructor. Kept async to
```
cvolzer3, line 37: I don't like this, but I couldn't find an elegant solution without jacking up the BaseSettingsManager interface (and subsequently, Claude adapter)

adboio:
to clarify, this will mean users can't use any of the MCPs they have configured with codex? how will they use MCPs with codex then? i think as a user it'd be frustrating if i had a bunch of MCPs all set up and ready to go with codex native and then they didn't work or i had to set them all up again in ph code

cvolzer3:
@adboio yeah, I was wondering what experience we wanted there. From how I understand it, we aren't importing the MCP servers from the user's Claude configs, so I mirrored that behavior with Codex. If we decide we want to bring their MCP integrations into PH Code, we would want to add a whole UI for that. Also, not sure how that'd work with sandbox agents.

packages/agent/src/adapters/codex/settings.ts
k11kirky: fine for now

packages/agent/src/adapters/codex/spawn.ts
```diff
@@ -28,6 +30,13 @@ function buildConfigArgs(options: CodexProcessOptions): string[] {
 
   args.push("-c", `features.remote_models=false`);
 
+  // Disable the user's local MCPs one-by-one so Codex only uses the MCPs we
+  // provide via ACP. We can't use `-c mcp_servers={}` because that makes Codex
+  // ignore MCPs entirely, including the ones we inject later.
```
k11kirky, line 35: hmm - im not sure we want to remove all locals right? only if they conflict we should favor ours

k11kirky (approved):
approving to unblock but added a question

adboio:
ahh i assumed we were bringing in claude MCPs, i don't use very many so i guess i never noticed lol

is this actively causing problems? if not, i think we should reconsider and try to find a way to bring in MCPs from both codex + claude instead of disabling them for both

cvolzer3:
discussion moved to slack: conclusion, let MCP Store own the MCP integrations & dont support local. future: add onboarding flow to combine/filter/dedupe from the local to populate the MCP store

PR #1644, opened 2026-04-14
feat(cloud-agent): support codex cloud runs

tatoalo:
## Problem

cloud runs let users pick`codex` via the UI but the selected runtime was not carried through the cloud-run path consistently

ph_code_codex_support

## Changes

- added explicit cloud runtime fields for adapter, provider, model, and reasoning effort
- updated the cloud agent server startup path to select the configured adapter explicitly instead of relying on Claude-oriented defaults.
- added shared reasoning-effort capability helpers so local and cloud flows use the same adapter/model support rules

Companion: https://github.com/PostHog/posthog/pull/54462

### DIFF
--- a/packages/agent/src/server/agent-server.ts
+++ b/packages/agent/src/server/agent-server.ts
@@ -19,6 +19,7 @@ import {
 } from "../adapters/acp-connection";
 import { selectRecentTurns } from "../adapters/claude/session/jsonl-hydration";
 import type { CodeExecutionMode } from "../execution-mode";
+import { DEFAULT_CODEX_MODEL } from "../gateway-models";
 import { PostHogAPIClient } from "../posthog-api";
 import {
   type ConversationTurn,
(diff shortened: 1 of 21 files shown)


packages/agent/src/server/agent-server.ts
```diff
@@ -638,6 +643,44 @@ export class AgentServer {
...
+        return null;
+      }),
+    ]);
+
+    const prUrl =
+      typeof (preTaskRun?.state as Record<string, unknown>)
+        ?.slack_notified_pr_url === "string"
+        ? ((preTaskRun?.state as Record<string, unknown>)
+            .slack_notified_pr_url as string)
+        : null;
```
jonathanlab: :/

jonathanlab (approved):
LGTM, thank you!

packages/agent/src/server/agent-server.ts
tatoalo: yeah fair, using an external helper to make it a bit less fugly

PR #1651, opened 2026-04-14
refactor: initial decoupling of electron

jonathanlab:
### What's going on?

When a user works on a task, the work happens somewhere, either locally on their machine or in a cloud sandbox. Right now, those two paths are different codepaths. Every feature that touches git, the filesystem, or the shell needs to know which mode it's in, which makes cloud parity with local very difficult.

That's one part of the problem. We also ideally want to be able to run PostHog Code on a bunch of different platforms:  
- Cloud  
- Mobile  
- Browser  
- Desktop

We also have a bunch of data/functionality dependencies:  
- Platform specific dependencies (Think clipboards, notifications, OS dialogues, etc, each platform thinks differently about these)  
- PostHog API (This one's easy to handle across different platforms)  
- Node server stuff; Think all our git handling, process spawning, file watchers etc.  Basically stuff we want to run and behave identically on both a sandbox and users local environment. Fortunately we know this will always run in a node environment, so we don't need a per platform implementation.

### What are we doing?

To make the above possible we need to split our node server (which we currently run inside electron) into two packages that can run anywhere. To achieve that, we need to get rid of the electron dependencies in the node server.

```
# This is node, running in a sandbox OR your local device, exposing TRPC procedures.
# It handles all workspace/task run specific logic, such as file changes.
packages/workspace-server

# This is node running on some web server/worker OR your local device, 
# and coordinates/talks to the workspace servers.
# It also handles non-task specific stuff.
# We can't run it in the sandbox because we might have multiple sandboxes 
# at the same time, plus they're ephemeral.
# This also exposes TRPC procedures
packages/app-server 

# Shared interfaces that all platforms must adhere to
# Toy example with clipboard usage, all electron related stuff goes here,
# it's also about stuff like path handling, file attachments, etc:
packages/platform <- interfaces such as (IClipboard, IExternalURL)

# These implement the behaviors described in packages/platform
apps/web/ <- WebClipboard
apps/desktop/ <- ElectronClipboard
apps/ui/ <- React DOM UI

# This one also implements it's own UI since we can't run React DOM here
apps/mobile <- MobileClipboard 
```

This makes the first step, by defining a package that all platforms must adhere to, in `packages/platform` .  
We make an initial start at implementing one of those interfaces via `apps/code/platform-adapters/electron-url-launcher.ts`. At runtime we inject this into our server. In a browser context we'd inject `web-url-launcher.ts etc.`. It's a toy example for the rest of the stack.

### That's great but aren't you just massively overengineering this?

Fully extracting electron will eventually allow us to split the node server we run in electron into `packages/server`, then expose that node server via the modal sandbox over hono with TRPC.

This is good because we can then route to either local or the sandbox based on whether the task run is cloud or local, which _IN THEORY_ could make cloud feel identical to local.  
  
It's a bigger lift, but IMO the right way to do this. As far as I can see it's the only way to prevent 50 (and soon more) `if workspace.mode === 'cloud'` statements scattered throughout our codebase. Without this, every feature that touches git/shell/files needs a local codepath and a cloud codepath, and that will become more painful as time goes on.
  
Instead we just call `trpc.git.getCurrentBranch` and route it to the right spot (either local or a sandbox) based on whether the task run is local or cloud.

It also open avenues to supporting other platforms such PostHog Code in the web :) (This is a nice side effect, definitely not the main concern). A more direct benefit is that we could share a lot of the code between PostHog Mobile and PostHog Code desktop   
  
---------  
  
I've spent quite a while thinking about this, and this architecture is loosely based on VSCode's internal structure. Happy to hear feedback

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,9 @@ jobs:
       - name: Build electron-trpc
         run: pnpm --filter @posthog/electron-trpc build
 
+      - name: Build platform
+        run: pnpm --filter @posthog/platform build
+
       - name: Build shared
         run: pnpm --filter @posthog/shared build
 
--- a/.github/workflows/code-release.yml
+++ b/.github/workflows/code-release.yml
@@ -85,6 +85,9 @@ jobs:
       - name: Build electron-trpc package
         run: pnpm --filter @posthog/electron-trpc run build
 
+      - name: Build platform package
+        run: pnpm --filter @posthog/platform run build
+
       - name: Build shared package
         run: pnpm --filter @posthog/shared run build
 
@@ -181,6 +184,9 @@ jobs:
 
       - name: Build electron-trpc package
         run: pnpm --filter @posthog/electron-trpc run build
+
+      - name: Build platform package
+        run: pnpm --filter @posthog/platform run build
 
       - name: Build shared package
         run: pnpm --filter @posthog/shared run build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,6 +78,7 @@ jobs:
       - name: Build packages
         run: |
           pnpm --filter @posthog/electron-trpc build &
+          pnpm --filter @posthog/platform build &
           pnpm --filter @posthog/shared build
           pnpm --filter @posthog/git build
           pnpm --filter agent build &
--- a/apps/code/package.json
+++ b/apps/code/package.json
@@ -131,6 +131,7 @@
     "@posthog/git": "workspace:*",
     "@posthog/hedgehog-mode": "^0.0.48",
     "@posthog/quill": "0.1.0-alpha.7",
+    "@posthog/platform": "workspace:*",
     "@posthog/shared": "workspace:*",
     "@radix-ui/react-collapsible": "^1.1.12",
     "@radix-ui/react-icons": "^1.3.2",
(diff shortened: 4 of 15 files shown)


jonathanlab:
* **#1654** <a href="https://app.graphite.com/github/pr/PostHog/code/1654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1653** <a href="https://app.graphite.com/github/pr/PostHog/code/1653?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1651** <a href="https://app.graphite.com/github/pr/PostHog/code/1651?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/1651?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

charlesvien (approved):

charlesvien (approved):

tatoalo (approved):

jonathanlab:
### Merge activity

* **Apr 20, 1:04 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/code/1651).
* **Apr 20, 1:04 PM UTC**: @jonathanlab merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/1651).

PR #1652, opened 2026-04-14
feat(cloud-agent): refine cloud PR publish policy

tatoalo:
## Problem

cloud runs PR behavior needed to be more predictable across origins and session states

We want:
- Slack-origin coding tasks to auto-create or auto-update PRs by default
- non-Slack cloud runs to stop for review unless the user explicitly asks to publish
- exploratory runs with `createPr=false` to reliably avoid publishing
- detected existing PRs in a live session to follow the same policy as the startup prompt

### DIFF
--- a/packages/agent/src/server/agent-server.ts
+++ b/packages/agent/src/server/agent-server.ts
@@ -1,4 +1,8 @@
-import type { ContentBlock } from "@agentclientprotocol/sdk";
+import type {
+  ContentBlock,
+  RequestPermissionRequest,
+  RequestPermissionResponse,
+} from "@agentclientprotocol/sdk";
 import {
   ClientSideConnection,
   ndJsonStream,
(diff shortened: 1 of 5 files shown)


packages/agent/src/server/agent-server.ts
```diff
@@ -1331,6 +1396,42 @@ ${attributionInstructions}
           }
         }
 
+        if (this.config.createPr === false) {
+          const meta =
```
joshsny: nit: this function is ugly and hard to parse, can we split it out in some way into some helpers?

joshsny (approved):
🚢🚢🚢

packages/agent/src/server/agent-server.ts
tatoalo: absolutely, thanks for flagging!

PR #1680, opened 2026-04-15
fix: prevent filepath wrapping in diff panel

reverb256:
## Problem

Long filepaths in the diff review panel wrap between the gray directory path and bold filename, breaking visual alignment.

Closes #1600

## Fix

Wrap both path segments in a single container with `nowrap` + `ellipsis` truncation:
- `whiteSpace: nowrap` keeps dir + filename on one line
- `overflow: hidden` + `text-overflow: ellipsis` truncates long paths
- `title` attribute shows full path on hover

## Tests

4 new tests in `ReviewShell.test.tsx` verifying filepath rendering and nowrap behavior.


.gitignore
```diff
@@ -59,3 +59,5 @@ plugins/posthog/local-skills/
 
 # Symlinked copies of posthog, to make developing against those APIs easier
 posthog-sym
+.pi-lens/
```
jonathanlab: Please gitignore these kind of files locally.

jonathanlab (changes requested):

jonathanlab:
CleanShot 2026-04-20 at 16 37 04@2x

I updated it to this instead, we mainly care about filename so let's show that first.

jonathanlab (approved):

PR #1676, opened 2026-04-15
fix: Prevent auto-generated title from overwriting manual rename

charlesvien:
## Problem

Auto-title generation silently bypassed the title_manually_set guard because getQueryData used exact key matching while task queries include filter objects in their keys.

Closes https://github.com/PostHog/code/issues/1669

## Changes

1. Fix cache lookup in useChatTitleGenerator to use getQueriesData (prefix match) instead of getQueryData (exact match)
2. Add title_manually_set guard to generateTaskTitle at task creation time

## How did you test this?

Manually

PR #1681, opened 2026-04-15
fix: preserve manual task rename from auto-generated title

### LINKED ISSUE #1669: Manual task rename gets overwritten by auto-generated name from latest message
nit: if I've renamed a task, sometimes it gets renamed back automagically to something to do with the most recent task. but you should keep a manual name and not replace it IMO

reverb256:
## Problem

When a user manually renames a task, the auto-generated title from the latest LLM message overwrites it. The `title_manually_set` check only ran *before* the async title generation call, so if the user renamed during generation, the stale check passed.

Closes #1669

## Fix

Extracted `shouldApplyAutoTitle()` pure function that reads `title_manually_set` from the React Query cache. Called both before and after the async generation to prevent the race condition.

## Tests

5 new tests in `shouldApplyAutoTitle.test.ts` covering all cases including the race condition.


jonathanlab:
Already handled by #1676

PR #1686, opened 2026-04-16
fix(cloud): show cloud run file content from session events instead of local file system

VojtechBartos:
## Problem

  When clicking a file link in an agent message during a cloud run, users see "Failed to load file." This happens because the app tries to read the file from the local filesystem (e.g., `/tmp/workspace/repos/.../src/App.js`), but that path only exists on the remote cloud sandbox — there's no local copy.

## Solution

  For cloud runs, extract file content from the session events instead of the local filesystem. The agent's tool call results already contain file content: read tool calls embed the file text, and write/edit tool calls include `newText` in their diffs.

## Showcase

https://github.com/user-attachments/assets/ae3f588e-d59b-48f6-b007-3a3c228e00ae

### DIFF
--- a/apps/code/src/renderer/features/code-editor/hooks/useCloudFileContent.ts
+++ b/apps/code/src/renderer/features/code-editor/hooks/useCloudFileContent.ts
@@ -1,6 +1,5 @@
-import { useSessionForTask } from "@features/sessions/hooks/useSession";
+import { useCloudEventSummary } from "@features/task-detail/hooks/useCloudEventSummary";
 import {
-  buildCloudEventSummary,
   type CloudFileContent,
   extractCloudFileContent,
 } from "@features/task-detail/utils/cloudToolChanges";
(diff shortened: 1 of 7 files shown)


apps/code/src/renderer/features/code-editor/hooks/useCloudFileContent.ts
```diff
@@ -0,0 +1,28 @@
...
+): CloudFileResult {
+  const session = useSessionForTask(enabled ? taskId : undefined);
+  const events = session?.events;
+  const isLoading = enabled && session === undefined;
+
+  return useMemo(() => {
+    if (!enabled || !events) {
+      return { content: null, touched: false, isLoading };
+    }
+    const summary = buildCloudEventSummary(events);
```
tatoalo: can we make this more efficient? users might have multiple tabs during active cloud runs, maybe we can have some hook there

apps/code/src/renderer/features/task-detail/utils/cloudToolChanges.ts
```diff
@@ -283,3 +284,48 @@ export function extractCloudToolChangedFiles(
...
+  content: string | null;
+  /** Whether the agent read, wrote, or deleted this file during the session. */
+  touched: boolean;
+}
+
+/**
+ * Combines read tool results and write/edit diffs to reconstruct the latest
+ * known content for a file from a cloud session's tool calls.
+ */
+export function extractCloudFileContent(
```
tatoalo, line 298: if the agent renames or moves a file how would this work? I guess we would end up in a broken state?

apps/code/src/renderer/features/code-editor/hooks/useCloudFileContent.ts
VojtechBartos: created share hook for the cloud event summary

apps/code/src/renderer/features/task-detail/utils/cloudToolChanges.ts
VojtechBartos: ups, add handle for it

tatoalo (approved):

PR #1687, opened 2026-04-16
feat: auto recover on disconnect

### LINKED ISSUE #1563: Agent should survive computer sleep and resume without requiring retry
I kicked off a query, my computer fell asleep. When I came back, it said there was a failure and whether to retry or new session.

https://posthog.slack.com/archives/C09G8Q32R6F/p1775678906619679

jonathanlab:
Closes #1563


tatoalo (approved):

apps/code/src/renderer/features/sessions/service/service.ts
```diff
@@ -73,6 +74,15 @@ import {
 } from "@utils/session";
 
 const log = logger.scope("session-service");
+const LOCAL_SESSION_RECONNECT_ATTEMPTS = 3;
```
tatoalo, line 77: Have observed any attempt at reconnect that failed? Not saying we don't need a counter here, just curios

jonathanlab: Mostly a safety net after coming up from sleep with other resources not being ready, bit of a guess since I don't really have a proof for it :)

PR #1707, opened 2026-04-17
feat(code): execute inline !bash commands in cloud sandbox

VojtechBartos:
## Problem

Inline `!command` bash was disabled for cloud runs — the handler used the local Shell service (node-pty) which can't reach the sandbox container.

## Solution

Added a `shell_execute` JSON-RPC method on the agent-server that spawns `bash -c <command>` inside the sandbox and streams stdout/stderr as SSE notifications. Renderer pre-generates an `executionId`, sends the command via `CloudTaskService.sendCommand`, and resolves when `shell_exit` arrives. Requires a paired backend PR to allowlist the new method.

## Prerequisites 

https://github.com/PostHog/posthog/pull/55064

## Showcase

https://github.com/user-attachments/assets/50c91c76-a77e-4bef-8f10-db35fef2a0b1

### DIFF
--- a/apps/code/src/renderer/features/sessions/service/service.ts
+++ b/apps/code/src/renderer/features/sessions/service/service.ts
@@ -185,6 +185,19 @@ export class SessionService {
   >();
   /** Maps toolCallId → cloud requestId for routing permission responses */
   private cloudPermissionRequestIds = new Map<string, string>();
+  private pendingCloudShells = new Map<
+    string,
+    {
+      taskRunId: string;
+      stdout: string;
+      stderr: string;
+      resolve: (result: {
+        stdout: string;
+        stderr: string;
+        exitCode: number;
+      }) => void;
+    }
+  >();
   private idleKilledSubscription: { unsubscribe: () => void } | null = null;
 
   constructor() {
@@ -555,6 +568,7 @@ export class SessionService {
     }
 
     this.unsubscribeFromChannel(taskRunId);
+    this.rejectPendingCloudShells(taskRunId, "Session ended");
     sessionStoreSetters.removeSession(taskRunId);
     if (session) {
       this.localRepoPaths.delete(session.taskId);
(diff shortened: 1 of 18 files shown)


apps/code/src/renderer/features/sessions/service/service.ts
```diff
@@ -2518,6 +2588,43 @@ export class SessionService {
...
+        executionId: update.executionId,
+        stream: update.stream,
+        chunkLen: update.chunk.length,
+        hasPending: !!pending,
+      });
+      if (pending) {
+        if (update.stream === "stdout") {
+          pending.stdout += update.chunk;
+        } else {
+          pending.stderr += update.chunk;
```
tatoalo, line 2645: can we bound this?

packages/agent/src/server/agent-server.ts
```diff
@@ -1980,6 +2006,113 @@ ${attributionInstructions}
...
+      executionId,
+      cwd,
+      command,
+    });
+
+    const child = spawn("bash", ["-c", command], {
+      cwd,
+      env: process.env,
+      stdio: ["ignore", "pipe", "pipe"],
+    });
```
tatoalo, line 2062: aren't we leaking `env`s here possibly?

apps/code/src/renderer/features/sessions/service/service.ts
```diff
@@ -2518,6 +2588,43 @@ export class SessionService {
...
+        exitCode: update.exitCode,
+        signal: update.signal,
+        hasPending: !!pending,
+      });
+      if (pending) {
+        this.pendingCloudShells.delete(update.executionId);
+        pending.resolve({
+          stdout: pending.stdout,
+          stderr: pending.stderr,
+          exitCode: update.exitCode ?? (update.signal ? 1 : 0),
```
tatoalo: nit for consistency, `signal` should not conflict with exit codes

tatoalo (commented):
some comments, thinking largely at how this could be abused 👀

PR #1710, opened 2026-04-17
Fix badges after redesign

Twixes:
## Problem

Badges got broken in the recent redesign:

Screenshot 2026-04-17 at 15 09 58@2x

## Changes

Making them a bit nicer. Also following the existing conventions where badges have a border, whereas buttons do not:

Screenshot 2026-04-17 at 15 09 27@2x


adamleithp (approved):
Lgtm

PR #1717, opened 2026-04-17
feat(phai): add refresh_session ACP method for sandboxes

skoob13:
## Problem

Sandbox agent-servers need a way to refresh a Claude session's MCP server config between turns — for example, to swap in freshly-minted OAuth credentials on a pre-prompt TTL check — without tearing down the whole session and losing conversation history.

## Changes

- Add `_posthog/refresh_session` custom ACP extension method (request/response, not notification) so the caller can await completion before sending the next prompt.
- Implement `ClaudeAcpAgent.extMethod()` for the Claude adapter: interrupts the running `Query`, ends the input stream, and builds a new `Query` with updated `mcpServers` using `resume` (not `sessionId`) so conversation history is preserved.
- Use the resume+rebuild pattern instead of `query.setMcpServers()` because the latter can non-deterministically surface plugin-installed MCPs over the CLI-supplied ones; passing `mcpServers` via `query()` options guarantees overwrite.
- Refactor `isNotification` in `acp-extensions.ts` to share matching logic with a new `isMethod` helper, and add a `POSTHOG_METHODS` registry.
- Reject refresh attempts while a prompt turn is in flight.

- New unit tests in `packages/agent/src/acp-extensions.test.ts` cover `isNotification`/`isMethod` including the `__posthog/` double-prefix case.
- New unit tests in `packages/agent/src/adapters/claude/claude-agent.refresh.test.ts` verify that `refresh_session`:
  - throws `methodNotFound` for unknown ext methods,
  - rejects while a prompt is running,
  - swaps `query`/`input`/`queryOptions` and preserves session-level state (accumulated usage, notification history),
  - builds the new `query()` call with `resume` instead of `sessionId` and the new MCP server set.

## Publish to changelog?

no

### DIFF
--- a/packages/agent/src/acp-extensions.test.ts
+++ b/packages/agent/src/acp-extensions.test.ts
@@ -1,0 +1,72 @@
+import { describe, expect, it } from "vitest";
+import {
+  isMethod,
+  isNotification,
+  POSTHOG_METHODS,
+  POSTHOG_NOTIFICATIONS,
+} from "./acp-extensions";
+
+describe("isNotification", () => {
+  it("matches the exact notification name", () => {
+    expect(
+      isNotification(
+        POSTHOG_NOTIFICATIONS.TURN_COMPLETE,
+        POSTHOG_NOTIFICATIONS.TURN_COMPLETE,
+      ),
+    ).toBe(true);
+  });
+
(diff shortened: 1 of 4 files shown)


tatoalo (approved):
lfg

PR #1723, opened 2026-04-17
fix: prevent token usage reset to 0 during /compact command

### LINKED ISSUE #1714: /compact breaks token usage ui
Image

compacting resets token usage to 0 incorrectly

reverb256:
## Problem
Fixes #1714

When using the `/compact` command, the token usage UI incorrectly showed 0 tokens used instead of preserving the actual token usage.

## Root Cause
`ClaudeAcpAgent` was sending `usage_update` with `used: 0` when handling `compact_boundary` system messages. This immediately reset the token count in the UI.

## Fix
Removed the `usage_update` call during `compact_boundary` events. Token usage is already properly updated on the next result message (which naturally happens after compaction). The `used: 0` call was a premature reset with no benefit.

## Changes
- `packages/agent/src/adapters/claude/claude-agent.ts` — removed 7 lines that sent `usage_update` with `used: 0` during compact_boundary

## Testing
- Existing test suite passes (3,539 tests)
- Manual verification: token usage persists correctly after `/compact`


charlesvien:
Going to bring this into the adapter upgrade today 👍

PR #1733, opened 2026-04-20
fix: add deeplinking to posthog code git setup

k11kirky:
## TLDR

The GitHub integration flow previously relied entirely on polling to detect when a user completed the GitHub App install in their browser. This meant the app had no reliable way to know when the flow succeeded or failed, and the UI couldn't react immediately or surface meaningful error states.

## Changes

- Added a deep link handler in `GitHubIntegrationService` that listens for `integration` callbacks from PostHog Cloud after the GitHub App install completes. The service emits typed events (`Callback`, `FlowTimedOut`) and queues callbacks that arrive before the renderer has subscribed.
- Added a 5-minute flow timeout in the service that fires `FlowTimedOut` if no deep link callback is received.
- Exposed three new tRPC endpoints: `onCallback` and `onFlowTimedOut` subscriptions, and a `consumePendingCallback` query to drain any callback that arrived before the renderer mounted.
- Added a `useGitHubIntegrationCallback` hook that subscribes to both events and drains any pending callback on mount, calling `onSuccess`, `onError`, or `onTimedOut` accordingly.
- Polling is now only used in dev mode (where the deep link protocol is not registered) and as a safety fallback timeout in production.
- The onboarding `GitIntegrationStep` now shows a timeout banner with a retry prompt, displays the connected GitHub account name when available, and surfaces a callout when GitHub is already connected on a different project.
- `GitHubIntegrationSection` in settings follows the same pattern, replacing polling-based detection with the new callback hook.
- The `next` redirect URL after GitHub authorization now points to `/account/social-connected` with query params instead of the project root, enabling the deep link callback to carry status, installation ID, and error details back to the app.


k11kirky:
* **#1733** <a href="https://app.graphite.com/github/pr/PostHog/code/1733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/1733?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/code/src/main/services/github-integration/service.ts
```diff
@@ -1,36 +1,152 @@
...
+    private readonly urlLauncher: IUrlLauncher,
+    @inject(MAIN_TOKENS.MainWindow)
+    private readonly mainWindow: IMainWindow,
+  ) {
+    super();
+
+    this.deepLinkService.registerHandler("integration", (_path, params) =>
+      this.handleCallback(params),
+    );
+    log.info("Registered integration deep link handler");
```
jonathanlab: nit: trying to keep log noise down, in general let's only log the failure cases

apps/code/src/main/services/github-integration/service.ts
```diff
@@ -1,36 +1,152 @@
...
+
+      log.info("Opening GitHub authorization URL in browser", { projectId });
       await this.urlLauncher.launch(authorizeUrl);
 
       return { success: true };
     } catch (error) {
+      this.clearFlowTimeout();
       return {
         success: false,
         error: error instanceof Error ? error.message : "Unknown error",
```
jonathanlab, line 85: nit: could use another log here just to know what went wrong in prod?

jonathanlab (approved):
LGTM

apps/code/src/main/services/github-integration/service.ts
k11kirky: Good call — removed the noisy success logs throughout this file and kept only failure/warn logs.

apps/code/src/main/services/github-integration/service.ts
k11kirky: Added a log.error in the catch block including the projectId and error message.

PR #1763, opened 2026-04-21
fix(signals): Use Zendesk ticket URLs instead of API URLs

sortafreel:
## Problem
- We linked to the API JSON, so clicking "Open" didn't allow to see actual human conversation

## Changes
- Now we link to the ticket


Twixes (approved):

Twixes (approved):
LGTM, although this specifically would make sense for me to inline

PR #1782, opened 2026-04-21
fix(agent): classify upstream prompt failures without replay

tatoalo:
## Problem

When the upstream Claude prompt stream terminated unexpectedly, the server surfaced a generic failure that did not preserve the underlying reason. Prompt-level replay also risked rerunning work after execution had already started, which is unsafe for non-idempotent ops

## Changes

* added structured upstream error classification in the Claude ACP conversion layer so terminated streams and connection failures can be identified without relying only on string matching
* removed prompt-level replay from the initial and resume send paths so transient upstream failures do not rerun a prompt after execution may already have started

### DIFF
--- a/packages/agent/src/adapters/claude/conversion/sdk-to-acp.ts
+++ b/packages/agent/src/adapters/claude/conversion/sdk-to-acp.ts
@@ -618,6 +618,41 @@ export type ResultMessageHandlerResult = {
   };
 };
 
+export type AgentErrorClassification =
+  | "upstream_stream_terminated"
+  | "upstream_connection_error"
+  | "agent_error";
+
+/**
+ * Classify an error string surfaced by the Claude CLI via `is_error: true`
+ * result messages. Transient upstream-stream terminations (e.g. the fetch body
+ * from the LLM gateway is torn down mid-stream) are retriable; most other
+ * errors are not.
+ */
+export function classifyAgentError(
+  result: string | undefined,
+): AgentErrorClassification {
+  if (!result) return "agent_error";
+  const text = result.trim();
+  // Anthropic SDK surfaces an undici fetch abort as "API Error: terminated".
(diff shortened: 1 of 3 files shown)


packages/agent/src/adapters/claude/conversion/sdk-to-acp.ts
```diff
@@ -608,6 +608,32 @@ export type ResultMessageHandlerResult = {
...
+ * from the LLM gateway is torn down mid-stream) are retriable; most other
+ * errors are not.
+ */
+export function classifyAgentError(
+  result: string | undefined,
+): AgentErrorClassification {
+  if (!result) return "agent_error";
+  const text = result.trim();
+  // Anthropic SDK surfaces an undici fetch abort as "API Error: terminated".
+  if (/API Error:\s*terminated\b/i.test(text)) {
```
joshsny, line 638: you love a little bit of regex error message matching

packages/agent/src/server/agent-server.ts
```diff
@@ -949,6 +953,50 @@ export class AgentServer {
...
+  private extractErrorClassification(error: unknown): {
+    classification: AgentErrorClassification;
+    message: string;
+  } {
+    const message =
+      error instanceof Error ? error.message : String(error ?? "");
+
+    // Prefer the structured `data` carried on RequestError if present.
+    const data = (error as { data?: unknown } | undefined)?.data;
+    if (
```
joshsny: this looks like zod would simplify it

joshsny (approved):

packages/agent/src/adapters/claude/conversion/sdk-to-acp.ts
tatoalo: if the claude code leak tought us anything right

PR #1747, opened 2026-04-21
feat(mcp-store): marketplace with installed-server rail and per-tool approval

cvolzer3:
Screenshot 2026-04-20 at 10 12 27 PM

## Summary

Rebuilds the MCP settings scene to match the new PostHog MCP store shape and the handoff prototype.

- Adopts the split **template** (shared OAuth) vs **custom** (per-user DCR) install model: `install_template/` for marketplace connects, `install_custom/` for user-supplied URLs with optional client creds.
- Adds per-tool **approval** UI (Approved / Requires approval / Blocked) with bulk set-all and a removed-tools toggle; backend proxy already enforces via JSON-RPC errors `-32001` / `-32002`.
- Restructures the section into a two-pane layout: a 256px **installed-server rail** with local search + status pulse dots, plus a right-hand pane that routes between marketplace / detail / add-custom.
- Introduces a `fullwidth` flag on Settings categories so MCP opts out of the standard 800px centered container; every other page is unchanged.
- Drops the dead `auth_type: "none"` branch in the main-process auth adapter — every installation routes through the MCP proxy.

## Test plan

- [ ] Settings → MCP servers renders the two-pane layout; other settings pages unchanged.
- [ ] Marketplace search + category chips filter the grid; "Clear filters" resets both.
- [ ] Rail search filters the installed list; "Active" count reflects filtered results.
- [ ] Rail installed list is alphabetical; status pulse shows green / amber / red correctly.
- [ ] Connecting an OAuth template opens the browser, auto-advances to the installation detail on callback, and populates tools with `needs_approval`.
- [ ] Per-tool toggle cycles Approved / Requires approval / Blocked; bulk set-all works; Refresh resyncs tools.
- [ ] Clicking an already-installed card in the marketplace routes to the installation detail (not the empty template view).
- [ ] Adding a custom OAuth server with optional client creds completes DCR + OAuth and lands on the new installation.
- [ ] Uninstalling from the detail view clears the rail row and returns to marketplace.
- [ ] `pnpm --filter code typecheck`, `pnpm lint`, and `pnpm --filter code test` all pass.

PR #1784, opened 2026-04-21
feat(mcp-store): agent request permission for tool calling

cvolzer3:
Screenshot 2026-04-21 at 2 47 43 PM

## Problem

When the agent uses a blocked or approval-required MCP tool, Claude incorrectly tells users to check "Claude Code settings." PostHog Code owns MCP installations, and needs_approval tools should show an interactive approval dialog instead of a static denial.

## Changes

- Intercept MCP tool calls in canUseTool(): `do_not_use` are denied and the user is directed to "Settings > MCP Servers"; `needs_approval` tools are shown permission dialog
- Fetch tool approval states from PostHog API at session start and thread them through `_meta` into the tool metadata cache
- When user approves a `needs_approval` tool, PATCH the backend to set `approved` before execution, otherwise the server-side proxy still rejects the call
- Sanitize server names to match the SDK's convention (non-alphanumeric → _)
- Skip read-only auto-approval for `needs_approval` tools
- Fix system prompt to stop Claude from suggesting Settings for generic MCP errors

## How did you test this?

- New unit tests in tool-metadata.test.ts (10 tests) and permission-handlers.test.ts (8 tests) covering all approval states, bypass mode, read-only override, and dialog title format
- Updated auth-adapter.test.ts and service.test.ts for new return types
- pnpm typecheck, pnpm --filter agent test (287 pass), pnpm --filter code test (770 pass)
- Manual: needs_approval shows dialog, do_not_use shows correct denial message


cvolzer3:
fyi, this builds on the other mcp store pr: https://github.com/PostHog/code/pull/1747

joshsny (approved):

PR #1812, opened 2026-04-22
fix(permissions): persist allow rules across worktrees of a repo

### LINKED ISSUE #1447: Permission approvals are not persistent
If I say yes and don't ask again, it only remembers for that particular worktree. If I go off to work on a different worktree later, I still have to approve the same commands again and again

richardsolomou:
## Problem

Closes #1447.

Clicking "Yes, don't ask again" in one worktree only remembered the approval for that specific worktree. Opening a sibling worktree of the same repository re-prompted the user for every command they had already approved. The root cause: Claude's `localSettings` scope resolves to `<cwd>/.claude/settings.local.json`, and each worktree has its own `cwd`.

## Changes

Anchor the `localSettings` scope in `SettingsManager` to the **primary worktree** (resolved via `listWorktrees` from `@posthog/git`) instead of the current working directory. Every worktree of a repo now reads and writes a single `.claude/settings.local.json` at the primary worktree's root. No new setting scope is introduced; the existing Claude settings hierarchy is preserved.

**`packages/agent/src/adapters/claude/session/repo-path.ts`** (new)
- `resolveMainRepoPath(cwd)` is a thin wrapper around `listWorktrees` from `@posthog/git/queries`. The primary worktree is always the first entry in `git worktree list --porcelain`. Falls back to `cwd` if not inside a git repo.

**`packages/agent/src/adapters/claude/session/settings.ts`**
- `SettingsManager` gains a `repoRoot` field, resolved in `loadAllSettings()`.
- `getLocalSettingsPath()` returns `<repoRoot>/.claude/settings.local.json` so every worktree shares a single file.
- New `addAllowRules(rules)` method persists rules to that file with:
  - `if (!this.initialized) await this.initialize()` guard so the method can't write to the wrong path if called pre-init.
  - An `AsyncMutex` (existing shared util) around the read/merge/write cycle so concurrent callers can't clobber each other.
  - A strict `readSettingsFileForUpdate` helper that throws on any read/parse error other than ENOENT — we refuse to overwrite a file we couldn't read, so a malformed `settings.local.json` is never wiped.
  - Atomic write via a `.tmp` file + `rename`, with cleanup on failure.
- `matchesRule` widened with a narrow catch-all (`rule.toolName === toolName && !rule.argument`) so argumentless rules persisted at the primary worktree auto-allow in secondary worktrees for non-ACP tools (TodoWrite, Task, WebSearch). The `!rule.argument` constraint prevents a deny rule like `Bash(rm -rf)` applied to a non-ACP Bash from matching any arbitrary command.
- ACP-prefix early-return in `checkPermission` dropped — our `SettingsManager` is now the only party that sees rules from the primary worktree in secondary-worktree sessions.

**`packages/agent/src/adapters/claude/permissions/permission-handlers.ts`**
- The `allow_always` flow now calls `settingsManager.addAllowRules(rules)` inside a try/catch (disk/parse failures fall back to session-scoped approval with a warning log) and returns `destination: "session"` to the SDK so it doesn't also persist to `<cwd>`.
- Non-`addRules` suggestions from the SDK (e.g. `addDirectories`) are forwarded with destination remapped to `session` via a `buildSessionPermissions` helper.
- `buildPermissionOptions` receives `settingsManager.getRepoRoot()` (new getter) instead of `session.cwd`, so the Bash "don't ask again" prompt reads e.g. *"don't ask again for `pnpm` commands in /Users/richard/dev/posthog"* — the primary worktree, consistent with where the rule actually lives.

**`packages/agent/src/adapters/claude/permissions/permission-options.ts`**
- Renamed the 3rd parameter `cwd` → `repoRoot` (the value now passed by the caller is the repo root, not the working directory).

**Cloud**: no-op. Cloud runs default to `bypassPermissions`, so the `allow_always` path is never entered.

## How did you test this?

Added `packages/agent/src/adapters/claude/session/settings.test.ts` with 8 tests that spin up a real `git init` + `git worktree add`:

- Allow rule persisted from a secondary worktree lands in the primary worktree's `.claude/settings.local.json`, not the secondary worktree's.
- A sibling worktree created *after* the write sees the rule on init and auto-allows.
- Name-based matching widens only for argumentless rules (argumentless `TodoWrite` allows; `UnknownTool(something)` with arbitrary input does not match).
- ACP-prefixed `mcp__acp__Bash` still auto-allows with a `Bash(pnpm test:*)` rule persisted at the primary worktree.
- `addAllowRules` refuses to overwrite a file whose contents are malformed JSON — throws, original bytes preserved.
- Three concurrent `addAllowRules` calls produce a file containing all three rules (no clobber).
- `resolveMainRepoPath` returns `cwd` for a non-git directory.

Ran:
- `pnpm --filter @posthog/agent test` — 208 passing.
- `pnpm lint` — no new warnings.
- `pnpm typecheck` — clean.

Manual: drove it through the desktop app with a repo that had two worktrees, clicked "don't ask again" for `pnpm` in worktree A, confirmed the file appeared only at the primary worktree, and confirmed worktree B ran `pnpm` without re-prompting.

---

### DIFF
--- a/packages/agent/src/adapters/claude/session/repo-path.ts
+++ b/packages/agent/src/adapters/claude/session/repo-path.ts
@@ -1,8 +1,4 @@
-import { execFile } from "node:child_process";
-import * as path from "node:path";
-import { promisify } from "node:util";
-
-const execFileAsync = promisify(execFile);
+import { listWorktrees } from "@posthog/git/queries";
 
 /**
  * Resolves the primary worktree (main repository) path for a given cwd.
@@ -12,25 +8,14 @@ const execFileAsync = promisify(execFile);
  * settings — such as "don't ask again" permission rules — to a single
  * location that every worktree of the same repository can read from.
  *
+ * `git worktree list --porcelain` always emits the primary worktree first.
  * Returns `cwd` when the directory is not inside a git repository or when
- * `git` is unavailable.
+ * git is unavailable.
  */
 export async function resolveMainRepoPath(cwd: string): Promise<string> {
   try {
-    const { stdout } = await execFileAsync(
-      "git",
-      ["rev-parse", "--path-format=absolute", "--git-common-dir"],
-      { cwd, timeout: 2_000 },
-    );
-    const commonDir = stdout.trim();
-    if (!commonDir) {
-      return cwd;
-    }
-    // `--git-common-dir` points to the primary worktree's `.git` directory,
-    // whose parent is the primary worktree root. For bare repositories the
-    // common dir is the repo itself; in that case `cwd` is the best guess.
-    const parent = path.dirname(commonDir);
-    return path.basename(commonDir) === ".git" ? parent : cwd;
+    const worktrees = await listWorktrees(cwd);
+    return worktrees[0]?.path ?? cwd;
   } catch {
     return cwd;
   }
--- a/packages/agent/src/adapters/claude/session/repo-path.ts
+++ b/packages/agent/src/adapters/claude/session/repo-path.ts
@@ -1,0 +1,37 @@
+import { execFile } from "node:child_process";
+import * as path from "node:path";
+import { promisify } from "node:util";
+
+const execFileAsync = promisify(execFile);
+
+/**
+ * Resolves the primary worktree (main repository) path for a given cwd.
+ *
+ * Secondary git worktrees share a `.git` common directory with the primary
+ * worktree. Returning the primary worktree path lets us scope per-repo
+ * settings — such as "don't ask again" permission rules — to a single
+ * location that every worktree of the same repository can read from.
+ *
+ * Returns `cwd` when the directory is not inside a git repository or when
+ * `git` is unavailable.
+ */
+export async function resolveMainRepoPath(cwd: string): Promise<string> {
+  try {
+    const { stdout } = await execFileAsync(
+      "git",
+      ["rev-parse", "--path-format=absolute", "--git-common-dir"],
+      { cwd, timeout: 2_000 },
+    );
+    const commonDir = stdout.trim();
+    if (!commonDir) {
+      return cwd;
+    }
+    // `--git-common-dir` points to the primary worktree's `.git` directory,
+    // whose parent is the primary worktree root. For bare repositories the
+    // common dir is the repo itself; in that case `cwd` is the best guess.
+    const parent = path.dirname(commonDir);
+    return path.basename(commonDir) === ".git" ? parent : cwd;
+  } catch {
+    return cwd;
+  }
+}
--- a/packages/agent/src/adapters/claude/session/settings.test.ts
+++ b/packages/agent/src/adapters/claude/session/settings.test.ts
@@ -110,6 +110,23 @@ describe("SettingsManager per-repo persistence", () => {
     expect(decision.decision).toBe("allow");
   });
 
+  it("refuses to overwrite the file when existing contents cannot be parsed", async () => {
+    const manager = new SettingsManager(worktree);
+    await manager.initialize();
+
+    const filePath = path.join(mainRepo, ".claude", "settings.local.json");
+    const original = "{ this is not valid json";
+    await fs.promises.mkdir(path.dirname(filePath), { recursive: true });
+    await fs.promises.writeFile(filePath, original);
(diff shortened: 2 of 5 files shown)


packages/agent/src/adapters/claude/session/settings.ts
```diff
@@ -319,6 +346,41 @@ export class SettingsManager {
...
+   * worktree of the same repository picks up the new rule on next load.
+   *
+   * Writes are serialised via `writeMutex` to prevent concurrent callers from
+   * clobbering each other, and use a temp-file + rename to keep the file
+   * consistent if the process dies mid-write.
+   */
+  async addAllowRules(rules: PermissionRuleValue[]): Promise<void> {
+    if (rules.length === 0) return;
+    if (this.initPromise) await this.initPromise;
+    await this.writeMutex.acquire();
```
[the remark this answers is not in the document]
richardsolomou: Good catch — fixed in 347295af. Switched to `if (!this.initialized) await this.initialize()` so it covers both the in-flight and never-started cases.

packages/agent/src/adapters/claude/session/settings.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    await manager.addAllowRules([
+      { toolName: "Bash", ruleContent: "pnpm test:*" },
+    ]);
+
+    const decision = manager.checkPermission("mcp__acp__Bash", {
+      command: "pnpm test --filter agent",
+    });
+    expect(decision.decision).toBe("allow");
+  });
```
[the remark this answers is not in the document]
richardsolomou: Done in 347295af — split into two focused `it` blocks (one asserting argumentless rules match, one asserting argumented rules don't).

packages/agent/src/adapters/claude/session/repo-path.ts
```diff
@@ -0,0 +1,37 @@
+import { execFile } from "node:child_process";
+import * as path from "node:path";
+import { promisify } from "node:util";
+
+const execFileAsync = promisify(execFile);
```
jonathanlab: afaik it's convention to use our `git` package for this, lmk if that doesn't work for your use case

jonathanlab: for this whole helper you added i mean, theres a git raw helper or something if i remember correctly

jonathanlab (approved):
This LGTM, thank you so much, this is way cleaner than a symlink.

packages/agent/src/adapters/claude/session/repo-path.ts
richardsolomou: Good call — switched to `listWorktrees` from `@posthog/git/queries` in 424608c8. Gives the same answer (the primary worktree is always first in `git worktree list --porcelain`) and routes through the managed git client like the rest of the codebase.

packages/agent/src/adapters/claude/session/settings.ts
```diff
@@ -319,6 +346,45 @@ export class SettingsManager {
...
+        ...(existing.permissions ?? {}),
+      };
+      const current = new Set(permissions.allow ?? []);
+      for (const rule of rules) {
+        current.add(formatRule(rule));
+      }
+      permissions.allow = Array.from(current);
+      const next: ClaudeCodeSettings = { ...existing, permissions };
+      await fs.promises.mkdir(path.dirname(filePath), { recursive: true });
+      await writeFileAtomic(filePath, `${JSON.stringify(next, null, 2)}\n`);
```
jonathanlab, line 399: AFAIK if we fail reading the settings in `loadingSettingsFile` due to some error, access failure, malformed JSON etc, this'll end up writing nothing back right, causing us to delete the contents? Let's make sure this never wipes anything please

richardsolomou: Excellent catch — fixed in dcb36b9e. Added a stricter `readSettingsFileForUpdate` helper used only by the write path: it still returns `{}` on ENOENT (safe to create a new file) but **throws** on any other read/parse error. The existing try/catch in the permission handler logs a warning and falls back to a session-scoped approval, so the file is never overwritten when we couldn't read it. Added a regression test covering a malformed-JSON file.

PR #1832, opened 2026-04-22
feat(code): space switcher for quick task navigation

adamleithp:
## Summary

Doesn't break cmd up/cmd down for going "home" when text is present in text box

- **Cmd+Up/Down** navigates between tasks (wraps around)
- **Hold Cmd 500ms** shows a vertical minimap overlay centered on active task
- Each item shows status icon + title + status text (Needs permission, Working, Completed, Failed, etc.)
- Top/bottom edges fade to transparent for long lists
- **"New task" item** at top of list, navigable via arrows
- **Cmd+0** added as alias for new task (alongside Cmd+N/T)
- Overlay unmounts from DOM when hidden (no backdrop-blur leak)

### Cmd+Up/Down behavior in inputs
- **Input with text** → native cursor movement (Cmd+Up to start, Cmd+Down to end)
- **Empty input or no input** → task navigation
- **Minimap visible** → always navigates, regardless of input content

### Minimap trigger
- Only appears from a pure Cmd hold (500ms) with no other keys pressed
- Cmd+Up/Down alone does NOT show the minimap
- Other shortcuts (Cmd+K, Cmd+B, etc.) cancel the hold timer

### Non-task views (inbox, settings, etc.)
- No item highlighted in minimap (uses -1 sentinel for no active slot)
- Arrow navigation starts from first/last item

## Test plan

- [x] Cmd+Up/Down switches tasks without showing overlay
- [x] Hold Cmd for ~500ms to see minimap appear
- [x] Arrow keys navigate within visible minimap (even with input focused)
- [x] Release Cmd to dismiss minimap
- [x] "New task" item at top, clickable and arrow-navigable
- [x] Cmd+0 opens new task view
- [x] Other Cmd shortcuts (Cmd+K, Cmd+B, etc.) don't trigger minimap
- [x] Status icons match task states (generating, completed, failed, needs permission)
- [x] Cmd+Up/Down in input with text → cursor moves to start/end (not task switch)
- [x] Cmd+Up/Down in empty input → switches task
- [x] On inbox/settings view, no item highlighted in minimap
- [x] With zero tasks, Cmd+Up/Down does nothing

### DIFF
--- a/apps/code/src/renderer/components/MainLayout.tsx
+++ b/apps/code/src/renderer/components/MainLayout.tsx
@@ -2,6 +2,7 @@ import { ConnectivityPrompt } from "@components/ConnectivityPrompt";
 import { HeaderRow } from "@components/HeaderRow";
 import { HedgehogMode } from "@components/HedgehogMode";
 import { KeyboardShortcutsSheet } from "@components/KeyboardShortcutsSheet";
+import { SpaceSwitcher } from "@components/SpaceSwitcher";
 
 import { ArchivedTasksView } from "@features/archive/components/ArchivedTasksView";
 import { CommandMenu } from "@features/command/components/CommandMenu";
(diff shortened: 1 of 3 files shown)


jonathanlab (approved):
Works like a charm! One tiny nit: Do you think we should make the space switcher itself show up faster? Right now it has a little delay on showing up, is that intentional? Maybe it'll feel more responsive if it shows up immediately when pressing CMD but maybe it just gets in the way, completely your call

PR #1786, opened 2026-04-21
feat: improve flag enricher

k11kirky:
## TLDR

Flag comments now include active/inactive status, 7-day evaluation stats (evals and unique users), and a direct link to the flag in PostHog.

## Problem

Flag annotations in code comments lacked context about whether a flag was actually active, how much traffic it was seeing, and where to find it in PostHog. Developers had to navigate to PostHog manually to get this information.

## Changes

- Added `getFlagEvaluationStats` to `PostHogApi`, which queries HogQL for `$feature_flag_called` events over a configurable lookback window (defaulting to 7 days) and returns per-flag evaluation counts and unique user counts
- Added `buildFlagUrl` helper that constructs a direct PostHog URL to a feature flag given the host, project ID, and flag ID
- Flag comments now include:
  - `active` / `inactive` status derived from the flag's `active` field
  - Evaluation stats formatted as `N evals / M users (7d)` when available
  - A direct URL to the flag in PostHog
  - `not in PostHog` when the flag key has no matching flag
- Flags not found in PostHog now short-circuit the comment formatter immediately rather than producing a partial comment
- `EnrichedFlag`, `EnrichedListItem`, and `EnrichmentContext` types extended to carry `url`, `evaluationStats`, `active`, `host`, and `projectId`
- `ParseResult.enrichFromApi` now fetches flag evaluation stats in parallel with other API calls and passes `host` and `projectId` through the enrichment context

PR #1837, opened 2026-04-22
feat: enrich wrapped functions and support jsx

k11kirky:
## TLDR

Adds cross-file PostHog wrapper detection so the enricher can recognize and annotate calls to user-defined functions that wrap PostHog SDK methods, and fixes JSX comment rendering to use `{/* */}` syntax instead of `//` when calls appear inside JSX elements.

## Problem

Previously, the enricher only detected direct PostHog SDK calls (e.g. `posthog.capture(...)`). If a codebase wrapped PostHog in a utility function like `track(event)` and imported it across files, those calls were invisible to the enricher. Additionally, inline comment annotations appended `// [PostHog] ...` suffixes unconditionally, which produces invalid syntax when the call site is inside a JSX tree.

## Changes

**Wrapper detection**
- Adds `wrapper-detector.ts` which analyzes function bodies to identify functions that directly call PostHog SDK methods, classifying them as either `fixed-key` (hardcoded event name) or `pass-through` (event name forwarded from a parameter). Supports JS/TS and Python.
- Adds `import-resolver.ts` which resolves relative import specifiers to absolute file paths for JS/TS (with extension probing and `index.*` fallback) and Python (relative dot-prefix syntax and `__init__.py` packages).
- `PostHogEnricher` gains `getWrappersForFile(absPath)` (with mtime-based LRU cache) and `findImportsInSource(source, langId, callerAbsPath)`.
- `file-enricher.ts` now detects relative imports in a file, resolves wrappers from those imported modules, and passes a `ParseContext` (containing `wrappersByLocalName` and `namespaceWrappers`) into `parse()`. Files with no direct `posthog` literal but with resolvable wrappers are still enriched.
- `call-detector.ts` consumes `ParseContext` to synthesize `PostHogCall` entries for bare and namespace wrapper calls, reusing the existing constant-map for identifier resolution.
- Import query strings added to `languages.ts` for JS/TS and Python.

**JSX-aware comment rendering**
- Adds `isInsideJsx()` helper in `ast-helpers.ts` that walks the AST upward to detect JSX element ancestry.
- `PostHogCall`, `CapturedEvent`, `FlagCheck`, `ListItem`, and `EnrichedListItem` gain an `inJsx` field.
- `comment-formatter.ts` now emits `{/* [PostHog] ... */}` for JSX-context items and handles mixed JSX/JS items on the same line by inserting a leading JSX-style block comment instead of an inline suffix.

**Other fixes**
- `FlagEvaluationStats` now includes `windowDays` so the comment formatter can render the actual query window instead of a hardcoded `7d`.
- Flag evaluation stat fetch failures are surfaced as `evaluationStatsError` and rendered as `"eval stats unavailable"` rather than silently omitting the stat.
- Flag URLs are now pre-computed in `parse-result.ts` (with trailing-slash normalization) and passed via `flagUrls` on `EnrichmentContext`, removing the `host`/`projectId` fields from that interface.
- `FunctionInfo` gains `bodyEndLine`.
- `ParserManager` uses a stable string key for query cache entries instead of `lang.toString()`.

### DIFF
--- a/packages/agent/src/enrichment/file-enricher.test.ts
+++ b/packages/agent/src/enrichment/file-enricher.test.ts
@@ -8,11 +8,15 @@ function makeDeps(overrides: {
   parseRejects?: Error;
   isSupported?: boolean;
   getApiKey?: () => string | Promise<string>;
+  findImportsInSource?: () => Promise<unknown[]>;
+  getWrappersForFile?: () => Promise<unknown[]>;
 }): {
   deps: FileEnrichmentDeps;
   parseSpy: ReturnType<typeof vi.fn>;
   enrichFromApiSpy: ReturnType<typeof vi.fn>;
   getApiKeySpy: ReturnType<typeof vi.fn>;
+  findImportsSpy: ReturnType<typeof vi.fn>;
+  getWrappersSpy: ReturnType<typeof vi.fn>;
 } {
   const enrichFromApiSpy = vi.fn(async () => ({
     toInlineComments: () =>
@@ -29,11 +33,19 @@ function makeDeps(overrides: {
   });
 
   const getApiKeySpy = vi.fn(overrides.getApiKey ?? (() => "phx_test"));
+  const findImportsSpy = vi.fn(
+    overrides.findImportsInSource ?? (async () => []),
+  );
+  const getWrappersSpy = vi.fn(
+    overrides.getWrappersForFile ?? (async () => []),
+  );
 
   const deps: FileEnrichmentDeps = {
     enricher: {
       isSupported: vi.fn(() => overrides.isSupported ?? true),
       parse: parseSpy,
+      findImportsInSource: findImportsSpy,
+      getWrappersForFile: getWrappersSpy,
     } as unknown as FileEnrichmentDeps["enricher"],
     apiConfig: {
       apiUrl: "https://test.posthog.com",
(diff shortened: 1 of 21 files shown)


k11kirky:
* **#1837** <a href="https://app.graphite.com/github/pr/PostHog/code/1837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/1837?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1786** <a href="https://app.graphite.com/github/pr/PostHog/code/1786?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jonathanlab (approved):
🚢 yes

k11kirky:
### Merge activity

* **Apr 23, 10:03 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/code/1837).
* **Apr 23, 10:04 AM UTC**: [Graphite](https://app.graphite.com/github/pr/PostHog/code/1837) rebased this pull request as part of a merge.
* **Apr 23, 10:15 AM UTC**: @k11kirky merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/1837).

PR #1858, opened 2026-04-23
fix(agent): allow bypassPermissions switch in sandboxed sessions

VojtechBartos:
## Problem

Approving a plan with "Yes, bypass all permissions" in a cloud run did nothing: the agent refused to proceed, the footer stayed on **Plan Mode**, and the next tool call re-prompted.

## Root cause

Three stacked gaps:

1. **SDK locked out of bypass.** The agent launches the Claude SDK with `allowDangerouslySkipPermissions: !IS_ROOT`. Cloud sandboxes run as root, so this was always `false` — the SDK rejected every bypass switch. The UI-side gate already handled this via `!IS_ROOT || IS_SANDBOX` (sandbox overrides root because the blast radius is a throwaway container). The SDK gate didn't. Aligned them.

2. **Cloud UI never saw the mode change.** Cloud runs receive `session/update` notifications through an SSE relay that only classifies `task_run_state` and `permission_request` — `config_option_update` fell through to the display log, never updating the store. The hydration path already knew to extract it; added the same scan to the live-update path.

3. **UI auto-reverted bypass → default.** A `SessionView` effect force-reset the mode when the user's local `allowBypassPermissions` preference was off. Correct for local sessions; wrong for cloud (sandboxed, the agent's `ALLOW_BYPASS` gate already said bypass is safe, user just explicitly picked it). Skip the revert when `isCloud`.

Also consolidated plan-approval mode-switch handlers onto the existing `applySessionMode + updateConfigOption` pattern used by `setSessionMode`, dropping a legacy `current_mode_update` emit the renderer no longer listens for.

### DIFF
--- a/apps/code/src/renderer/features/sessions/components/SessionView.tsx
+++ b/apps/code/src/renderer/features/sessions/components/SessionView.tsx
@@ -24,6 +24,7 @@ import {
 import { getFilePath } from "@utils/getFilePath";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { getSessionService } from "../service/service";
+import { flattenSelectOptions } from "../stores/sessionStore";
 import {
   useSessionViewActions,
   useShowRawLogs,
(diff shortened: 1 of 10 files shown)


joshsny (approved):

tatoalo:
agent-server still listens for that notification to update session.permissionMode, right?

`applySessionMode` or `updateConfigOption` are not emitting `current_mode_update`. So after a cloud plan approval, the agent-server's cached mode is frozen at whatever it was before plan mode?

I tried on a plan-initiated cloud run, then approved the plan with manual approval needed but then agent did the change autonomously 👀

tatoalo (approved):

PR #1859, opened 2026-04-23
fix(code): prevent sidebar from merging distinct repos into one group

VojtechBartos:
## Problem

In the sidebar's "by-project" mode, tasks from distinct repositories (e.g. `posthog/posthog` and `posthog/code`) collapsed into a single group labelled "posthog".

Three issues combined:

1. **Case inconsistency** — local tasks stored the repo verbatim from the git remote (`PostHog/code`), while cloud tasks lowercased it (`posthog/code`). Same repo, two different group keys.
2. **Legacy black-hole bucket** — `task.repository` values that didn't parse as `owner/repo` (e.g. a single segment like `"posthog"`) were used as the raw group key, pulling unrelated tasks into one group.
3. **Ambiguous labels** — groups were labelled with just the repo name, so two distinct repos sharing a short name looked identical.

## Solution

- Canonicalize the group key: `normalizeRepoKey(repository).toLowerCase()` so casing and `.git` suffix differences collapse into one group.
- Skip malformed repository values so they fall through to the folder path or `other` bucket instead of colliding with real repos.
- When two groups share a display name, prefix the organization (`PostHog/shared` vs `acme/shared`) so they stay visually distinct.
- Extracted the pure grouping helpers into `sidebar/utils/groupTasks.ts` with 15 unit tests covering case-merge, distinct repos, collision disambiguation, legacy values, folder-path fallback, folder order, and mixed buckets.

### DIFF
--- a/apps/code/src/renderer/features/sidebar/components/TaskListView.tsx
+++ b/apps/code/src/renderer/features/sidebar/components/TaskListView.tsx
@@ -337,8 +337,7 @@ export function TaskListView({
               const folder = folders.find(
                 (f) =>
                   (f.remoteUrl &&
-                    normalizeRepoKey(f.remoteUrl).toLowerCase() ===
-                      normalizeRepoKey(group.id).toLowerCase()) ||
+                    normalizeRepoKey(f.remoteUrl).toLowerCase() === group.id) ||
                   f.path === group.id,
               );
               const groupFolderId =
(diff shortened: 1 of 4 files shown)


jonathanlab (approved):
lgtm, thanks, make sure to check if the greptile comments are worth addressing

PR #1871, opened 2026-04-23
chore(code): close SpaceSwitcher on escape

VojtechBartos:
## Problem

Holding Cmd for 500ms opens the `SpaceSwitcher` minimap, but sometimes it stays stuck on screen with no way to dismiss it — the only way to close it was to release Cmd or blur the window, and in edge cases (e.g. key events getting lost, focus shifting) the switcher would remain visible indefinitely. Users had no escape hatch.

## Solution

Wire `Escape` up to `hide()` when the switcher is mounted. Reuses the existing `SHORTCUTS.BLUR` constant and the same hotkey options as the other `SpaceSwitcher` bindings so it works across form tags and contenteditable elements.


adboio (approved):
bless you

PR #1875, opened 2026-04-23
chore: install screen pizzaz

corywatilo:
I spent a while trying to get the text to render in white but apparently it's not possible with the way we build it. Thought about adding a solid background color behind where the text will render but there are so many variables (ie: color mode, if you're showing file extensions or not, etc). We can always come back to this later, but at least it's a step in the right direction.

## Before

image

## After

image

_[Figma](https://www.figma.com/design/IYG4CDj4Pq6C3bjUy3JBfm/PostHog-Code?node-id=194-962&t=phnKaUUYmiSEiw8E-1) in case we tweak later_


k11kirky (approved):
I LOVE IT

PR #1885, opened 2026-04-25
feat(agent): structured output for codex adapter via stdio MCP

ryans-posthog:
## Problem

The Codex adapter has no equivalent to Claude's `outputFormat` for structured output. Callers that pass `_meta.jsonSchema` through the agent server expecting a validated JSON payload back got plain-text completions instead — Codex simply ignored the schema.

## Changes

- New stdio MCP server (`packages/agent/src/adapters/codex/structured-output-mcp-server.ts`) that registers a single `create_output` tool whose input shape is generated from the caller-supplied JSON schema. AJV validates each invocation; invalid input is returned as a tool error so the model can retry.
- `CodexAcpAgent.applyStructuredOutput` injects this MCP server into `newSession` / `loadSession` / `unstable_resumeSession` / `unstable_forkSession` whenever `_meta.jsonSchema` is set and an `onStructuredOutput` callback is wired. The schema is passed to the child process as a base64-encoded env var to avoid shell escaping. A short system-prompt note is appended via `_meta.systemPrompt` instructing the model to call `create_output` with the final result.
- `createCodexClient` intercepts `tool_call` / `tool_call_update` notifications: it captures `rawInput` across status transitions and fires `onStructuredOutput(...)` exactly once per tool-call id when status reaches `completed`. State is cleared on fire so a re-emitted update can't double-fire.
- Resolver walks up from the bundled adapter location to find the compiled MCP server script across the various tsup entry points (agent.js, server/bin.cjs, server/harness/bin.js, …). `POSTHOG_STRUCTURED_OUTPUT_MCP_SCRIPT` overrides the lookup.
- Adds the script as a tsup entry, plumbs `onStructuredOutput` through `acp-connection.ts` → `CodexAcpAgent` → `createCodexClient`, and adds `@modelcontextprotocol/sdk` as a dependency.

## How did you test this?

- Reviewed the diff and confirmed `pnpm --filter agent typecheck` reports only pre-existing errors in `enrichment/file-enricher.ts` (same errors are present on `main`, unrelated to this PR).
- No unit tests added or run in this PR. The callback state machine in `codex-client.ts` (one-shot fire across `tool_call` + `tool_call_update`) and `applyStructuredOutput`'s MCP-server / system-prompt injection are both unit-testable and would be worth covering in a follow-up — the existing `codex-client.test.ts` already mocks the upstream connection and is the natural place to add a test.


ryans-posthog:
* **#1885** <a href="https://app.graphite.com/github/pr/PostHog/code/1885?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/1885?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/agent/src/adapters/codex/codex-agent.ts
```diff
@@ -152,6 +166,49 @@ function getCurrentPermissionMode(
...
+ * The child process validates tool input against the JSON schema with AJV.
+ * We pass the schema as a base64-encoded env var to avoid shell escaping.
+ *
+ * Path resolves relative to the compiled adapter location. When bundled into
+ * different entry points (dist/agent.js, dist/server/bin.cjs, dist/server/
+ * harness/bin.js, etc), `import.meta.dirname` sits at different depths. Walk
+ * up until we find the script so each bundle locates the shared dist asset.
+ */
+function resolveStructuredOutputMcpScript(): string {
+  const override = process.env.POSTHOG_STRUCTURED_OUTPUT_MCP_SCRIPT;
```
joshsny: what's the override here for, thsi seems a bit odd versus just importing the default script but maybe i am missing something here?

packages/agent/src/adapters/codex/structured-output-mcp-server.ts
joshsny, line 1: this would be nice to support as a general local MCP server that folks can extend with their own tools, rather than just for structured output, and then we can add the create_output tool here as the first one available - we'd want it available in claude too if we go that route, but we can do that in a later PR if someone asks for it

joshsny (commented):
wasn't able to run this locally, so the review is low context, but assuming you've tested this locally on a cloud run lets merge and see what it's like in prod

joshsny (approved):

packages/agent/src/adapters/codex/codex-agent.ts
ryans-posthog: This... might have been a remnant of some local testing I was doing, let me investigate

PR #1900, opened 2026-04-27
feat(mcp_store): MCP brand logos and auto-refresh after connect

cvolzer3:
## Problem

Currently, the MCP servers in the marketplace use a fallback icon.

## Changes

- Add ~30 new brand icon assets
- Adds icon_key normalization

## How did you test this?

- Tested locally


apps/code/src/renderer/features/settings/hooks/useMcpInstallationTools.ts
joshsny, line 1: is there a cleaner way to do this, the ref + effect logic seems quite complicated 🤔

apps/code/src/renderer/features/settings/components/sections/McpServersSettings.tsx
joshsny, line 1: this one is getting pretty unweildy with the effect calls, I think we could simplify it a bunch for some of those by moving the things in the effect calls to whatever makes modifications to the dependency array or moving state around a bit

joshsny (approved):
left a few small nits

PR #1904, opened 2026-04-28
fix(code): skip file-watcher flush after shutdown

VojtechBartos:
## Problem

We saw the following unhandled rejection during app shutdown:

```
Unhandled rejection Error: Database not initialized — call initialize() first
    at get db (DatabaseService)
    at WorkspaceRepository.findAll
    at WorkspaceService.getAllTaskAssociations
    at WorkspaceService.handleGitStateChanged
    at FileWatcherService.emit
    at FileWatcherService.flushPending
    at Timeout._onTimeout
```

`FileWatcherService.queueEvents` schedules a 500 ms `setTimeout(flushPending, …)` per repo whenever `@parcel/watcher` reports changes. During shutdown, `AppLifecycleService.doShutdown` runs `watcherRegistry.shutdownAll()` (sets `isShutdown = true`, unsubscribes parcel watchers) and then closes the database. A pending `flushPending` timer can fire **after** `db.close()` has nulled the connection — `flushPending` emits `GitStateChanged`, `WorkspaceService.handleGitStateChanged` calls `WorkspaceRepository.findAll`, and the DB getter throws.

The two parcel watcher callbacks already guard with `if (this.watcherRegistry.isShutdown) return;`, but the in-flight debounce timer that lands in `flushPending` slips past that guard.

## Solution

Add the same `isShutdown` early-return at the top of `FileWatcherService.flushPending` and clear the pending buffers/timer so the in-flight timer is a no-op once the registry is down. This mirrors the existing guards in `watchFiles` and `watchGit`, so behavior is uniform across all three places where the service decides whether to emit. No changes are needed in `WorkspaceService`, `DatabaseService`, or `AppLifecycleService` — the lifecycle ordering is fine, the file-watcher just needs to stop emitting after shutdown.

Added a unit test covering both branches (shut-down: silent + cleared; active: emits per-path events).

---

### DIFF
--- a/apps/code/src/main/services/file-watcher/service.test.ts
+++ b/apps/code/src/main/services/file-watcher/service.test.ts
@@ -1,0 +1,90 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { FileWatcherEvent } from "./schemas";
+
+vi.mock("../../utils/logger.js", () => ({
+  logger: {
+    scope: () => ({
+      info: vi.fn(),
+      error: vi.fn(),
+      warn: vi.fn(),
+      debug: vi.fn(),
+    }),
+  },
+}));
+
+vi.mock("@parcel/watcher", () => ({ subscribe: vi.fn() }));
+
+import type { WatcherRegistryService } from "../watcher-registry/service";
+import { FileWatcherService } from "./service";
+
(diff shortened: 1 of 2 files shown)


jonathanlab (approved):
Nice, good catch

PR #1905, opened 2026-04-28
fix(code): reliable cloud message-queue dispatch + optimistic prompt UX

VojtechBartos:
## Problem

Cloud follow-up messages typed during sandbox setup, death, or restore could be silently lost — a generic error toast and the prompt vanished. Users also had no visual anchor for the prompt they typed at task creation until the agent eventually echoed it back via SSE (many seconds while the sandbox provisions).

The deeper framing: **local runs already had a working version of all of this**. This PR closes the parity gap.

## Local–cloud parity at a glance

| Capability | Local (existing) | Cloud (before) | Cloud (this PR) |
|---|---|---|---|
| "Agent is ready" signal | `agent.reconnect.mutate()` resolving sets `session.status: "connected"` | None (heuristics on `cloudStatus` + `isPromptPending`) | `_posthog/run_started` flips `session.status: "connected"` |
| Queue gate when agent isn't ready | `sendLocalPrompt` checks `session.status` upfront, throws if not connected | Inferred from `cloudStatus` / `isPromptPending` / optimistic-pending heuristic — racy | `sendCloudPrompt` queues if `session.status !== "connected"` (same shape as local) |
| Drain trigger | Local doesn't queue across turns the same way (synchronous mutate) | Multiple racy auto-flushes | Single trigger: `_posthog/turn_complete` |
| Re-enqueue on failure | N/A (local mutate throws synchronously) | Old retry-loop dropped the prompt on max-retries | `prependQueuedMessages` rolls the queue back |
| Optimistic user bubble | `appendOptimisticItem` + `replaceOptimisticWithEvent` swap in place | None | Seeded in `hydrateCloudTaskSessionFromLogs`, pinned at top, content-deduped against echo |

## Changes

1. **`chore(agent)`: emit `_posthog/run_started`** — the long-declared lifecycle handshake, finally wired up. Persisted to the log so warm reconnects replay it.
2. **`fix(code)`: reliable queue dispatch** — peek-and-confirm dispatcher, single `turn_complete` drain trigger, `session.status === "connected"` queue gate, plus an immer-proxy fix in `dequeueMessages` that was the silent root cause behind "queue clears, message disappears".
3. **`feat(code)`: pin prompt at top during sandbox setup** — optimistic bubble seeded only when there's no prior history, pinned above the conversation for cloud, content-deduped against the agent's echo. Local sessions untouched.

## Showcase

https://github.com/user-attachments/assets/071cef80-e78a-4fab-8554-c1466f93b804

## Test plan

- [x] Initial cloud prompt visible immediately during sandbox boot
- [x] Follow-ups typed during boot / mid-turn / sandbox restore queue and dispatch correctly
- [x] Reopened task with prior conversation: no flash, no duplicate bubbles
- [x] Sandbox death → resume → queued messages dispatch after the resume turn completes
- [x] Local sessions unaffected
- [x] Typecheck, lint, all tests pass
- [x] Manually tested end-to-end

---

### DIFF
--- a/apps/code/src/renderer/features/sessions/stores/sessionStore.ts
+++ b/apps/code/src/renderer/features/sessions/stores/sessionStore.ts
@@ -376,18 +376,43 @@ export const sessionStoreSetters = {
   },
 
   dequeueMessages: (taskId: string): QueuedMessage[] => {
-    let queuedMessages: QueuedMessage[] = [];
+    // Read the queue from the frozen committed state BEFORE entering the
+    // immer draft, otherwise the items returned are proxies that get
+    // revoked when setState exits and any later access throws
+    // "Cannot perform 'get' on a proxy that has been revoked".
+    const state = useSessionStore.getState();
+    const taskRunId = state.taskIdIndex[taskId];
+    if (!taskRunId) return [];
+    const session = state.sessions[taskRunId];
+    if (!session || session.messageQueue.length === 0) return [];
+
+    const queuedMessages = [...session.messageQueue];
+
+    useSessionStore.setState((draft) => {
+      const trid = draft.taskIdIndex[taskId];
+      if (!trid) return;
+      const draftSession = draft.sessions[trid];
+      if (draftSession) {
+        draftSession.messageQueue = [];
+      }
+    });
+
+    return queuedMessages;
+  },
+
+  /**
+   * Splice messages back at the head of the queue. Used to roll back a
+   * dispatch attempt that drained the queue but failed before delivery.
+   */
+  prependQueuedMessages: (taskId: string, messages: QueuedMessage[]) => {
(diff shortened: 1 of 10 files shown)


apps/code/src/renderer/features/sessions/stores/sessionStore.ts
tatoalo, line 363: this is behaving like the old `dequeueMessages` so probably worth fixing as well to avoid further regressions in one go, what do you think?

tatoalo (approved):

apps/code/src/renderer/features/sessions/stores/sessionStore.ts
VojtechBartos: yup, fixed and pushed

PR #1921, opened 2026-04-28
fix(cloud-agent): improve cloud run reconnection

tatoalo:
## Problem

cloud task runs could look stale in ph-code after renderer refreshes or route churn. 
renderer could miss earlier log updates from an already-running watcher, remain behind after log count gaps, or keep showing a prompt as pending after the cloud agent had already completed the turn.


VojtechBartos (approved):
🚢 it

PR #1940, opened 2026-04-29
fix: use null auth store, use hook instead of zust store

k11kirky:
## Problem

Auth state was being read directly from `useAuthStore` in utility functions outside of React components, which is stale if this isnt the first session

## Changes

- Added `getCachedAuthState()` helper that reads the current `AuthState` directly from the React Query cache, falling back to `ANONYMOUS_AUTH_STATE` if no cached value exists
- Replaced `useAuthStore.getState()` calls in `posthogLinks.ts` and `urls.ts` with `getCachedAuthState()`, making the query cache the single source of truth for auth state in non-component contexts

### DIFF
--- a/apps/code/src/renderer/features/auth/hooks/authQueries.ts
+++ b/apps/code/src/renderer/features/auth/hooks/authQueries.ts
@@ -36,6 +36,13 @@ export async function fetchAuthState(): Promise<AuthState> {
   return await trpcClient.auth.getState.query();
 }
 
+export function getCachedAuthState(): AuthState {
+  return (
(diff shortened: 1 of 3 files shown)


k11kirky:
* **#1940** <a href="https://app.graphite.com/github/pr/PostHog/code/1940?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/1940?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

VojtechBartos (approved):

PR #1942, opened 2026-04-29
chore: tag all tasks runs under posthog code limits

joshsny:
we were splitting out cloud and local runs, let’s use posthog_code for everything now so that we can limit all task runs under a posthog code subscription


k11kirky (approved):
how is internal pretask set? just double checking this is safe

joshsny:
> how is internal pretask set? just double checking this is safe

it's fetched from the API, so we're all good

PR #1972, opened 2026-05-01
feat(handoff): reduce size of packs, update stale remoteURLs

jonathanlab:
- Git's handover pack files sent to S3 are now smaller. They used to dump your whole git history, now it's only the relevant parts, and it excludes irrelevant/remote commits
- Fixed a bug where remote URL wasn't updated if the upstreams remote URL changed

---


k11kirky (approved):
greptile comments seem valid

jonathanlab:
addressed both

PR #1991, opened 2026-05-04
feat(agent): gate destructive PostHog MCP exec sub-tools

skoob13:
## Problem

The PostHog MCP exposes a single `exec` dispatcher (`mcp__posthog__exec`) that runs subcommands like `call dashboard-update {...}`. Once a user approves `mcp__posthog__exec` once (e.g. via "always allow"), every subsequent `call` flows through silently — including destructive writes (`update`, `partial-update`, `delete`, `destroy`). This violates least-privilege expectations: an approval intended for read traffic ends up authorizing arbitrary live-data mutations.

## Changes

- New `posthog-exec-gate.ts` matches PostHog `exec` tools, extracts the sub-tool from the `command` string, and flags the destructive subset.
- `canUseTool` re-gates destructive sub-tools at sub-tool granularity. Approvals can be persisted per sub-tool (`posthogApprovedExecTools` in `.claude/settings.local.json`) via `SettingsManager.addPostHogExecApproval`. `auto`/`bypassPermissions` modes still skip the gate.
- `PreToolUseHook` now returns `permissionDecision: "ask"` when an existing settings allow rule would otherwise short-circuit a destructive PostHog sub-tool, forcing the SDK to invoke `canUseTool`. Returning `{ continue: true }` was insufficient — the SDK falls back to its default permission flow which re-checks the same allow rule.

- New unit tests: `posthog-exec-gate.test.ts` (regex coverage), `permission-handlers.test.ts` (gate flow), `settings.test.ts` (approval persistence), `hooks.test.ts` (deferral behavior).
- Manually reproduced in PostHog Code with a `mcp__posthog__exec` allow rule in `~/.claude/settings.local.json`: `call dashboard-update {...}` now triggers an approval prompt; `call experiment-get {...}` does not.

## Publish to changelog?

no

### DIFF
--- a/packages/agent/src/adapters/claude/hooks.test.ts
+++ b/packages/agent/src/adapters/claude/hooks.test.ts
@@ -7,7 +7,16 @@ vi.mock("../../enrichment/file-enricher", () => ({
   enrichFileForAgent: enrichFileMock,
 }));
 
-import { createReadEnrichmentHook, type EnrichedReadCache } from "./hooks";
+import { Logger } from "../../utils/logger";
+import {
+  createPreToolUseHook,
+  createReadEnrichmentHook,
+  type EnrichedReadCache,
+} from "./hooks";
+import type {
+  PermissionCheckResult,
+  SettingsManager,
+} from "./session/settings";
 
 const stubDeps = {} as FileEnrichmentDeps;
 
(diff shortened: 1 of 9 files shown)


tatoalo (approved):
ooof, this is great, thanks!

PR #2018, opened 2026-05-04
Add token refresh endpoints to agent server

ryans-posthog:
## Summary

- Adds a `set_token` command (also aliased as `posthog/set_token`) to the agent server that updates `GH_TOKEN` and `GITHUB_TOKEN` on the agent process so **Claude-based agents** pick up a refreshed token without restart.
- Adds a localhost-only `POST /gh` route that runs an arbitrary `gh` CLI invocation using the agent's current env. **Codex-based agents** run in an isolated sandbox, so a shell-script wrapper hits this endpoint to apply the new token through `gh`.
- Adds a small `gh-exec` helper (`runGh` + `isLoopbackAddress`) that spawns `gh` directly (no shell), inherits the agent's `process.env`, and supports timeouts. The `/gh` route reuses it; tests exercise both happy/error paths.

## Test plan

- [x] `pnpm --filter agent test` — full agent server suite passes (149/149), including 15 new `gh-exec` tests and 6 new `agent-server` tests covering both `set_token` aliases, `/gh` body validation, and that `/gh` does not require JWT.
- [x] `pnpm --filter agent typecheck` — clean.
- [x] Biome — clean (lint-staged ran on commit).
- [ ] Manual: send `POST /command` with `{"command":"set_token","params":{"token":"…"}}` — verify `GH_TOKEN` updates in the agent process.
- [ ] Manual: from inside a Codex sandbox, hit `POST /gh` with the wrapper script — verify `gh` runs with the new token.

### DIFF
--- a/packages/agent/src/server/gh-exec.ts
+++ b/packages/agent/src/server/gh-exec.ts
@@ -1,0 +1,103 @@
+import { spawn } from "node:child_process";
+import type { Logger } from "../utils/logger";
+
+export interface GhExecOptions {
+  cwd: string;
+  timeoutMs: number;
+  logger?: Logger;
+  binary?: string;
+}
+
+export interface GhExecResult {
+  stdout: string;
+  stderr: string;
+  exitCode: number | null;
+  signal: NodeJS.Signals | null;
+  timedOut: boolean;
+}
+
+const KILL_GRACE_MS = 2_000;
+
+export async function runGh(
+  args: string[],
+  options: GhExecOptions,
+): Promise<GhExecResult> {
+  const binary = options.binary ?? "gh";
+  const logger = options.logger;
+
+  logger?.debug("Running gh", {
+    binary,
+    args,
+    cwd: options.cwd,
+    timeoutMs: options.timeoutMs,
+  });
+
+  return new Promise<GhExecResult>((resolve, reject) => {
+    const child = spawn(binary, args, {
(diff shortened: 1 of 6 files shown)


ryans-posthog:
Make sure to not allow the caller to specify the binary.

joshsny (approved):
interestring trick to get this to work, lgtm from the code side, worth testing this properly in a modal sandbox with the modal docker setup before shipping it

packages/agent/src/server/gh-exec.ts
```diff
@@ -0,0 +1,112 @@
...
+  logger?.debug("Running command", {
+    binary,
+    args,
+    cwd: options.cwd,
+    timeoutMs: options.timeoutMs,
+  });
+
+  return new Promise<GhExecResult>((resolve, reject) => {
+    const child = spawn(binary, args, {
+      cwd: options.cwd,
```
tatoalo, line 48: do we really need to put all envs? we know what `gh` needs from us here right? would like to avoid subprocesses seeing everything if not strictly needed

ryans-posthog: Good point, I'll restrict the env to only the Github Token since that's really all we need here

k11kirky:
@ryans-posthog @joshsny is this pr still relevant since the change to tokens?

PR #2027, opened 2026-05-05
feat(code): GitHub settings page + cloud-mode missing-integration warning

VojtechBartos:
## Problem

Two adjacent gaps in the personal-GitHub UX for cloud tasks:

- If a user has no personal GitHub integration, the cloud repo picker silently renders empty — no signal that anything is wrong, no path to fix it.
- There's no dedicated place to manage personal GitHub installations. Users can have multiple (different orgs/accounts), but settings doesn't surface them and offers no Disconnect.

## Solution

- **Cloud-mode warning** — when workspace mode is "cloud" and the user has no personal GitHub integration, render an amber callout under the prompt input with a one-click Connect GitHub button (uses the existing `useGithubUserConnect` hook).
- **GitHub settings page** — new sidebar category that lists *every* linked personal GitHub installation, each with its own Disconnect confirmation. Single Connect / Connect-another-account button at the top.
- Cloud tasks still use the user's personal integration only — no team-integration fallback, so cloud PRs are always authored as the user.

Combines the spirit of #1989 and #2009 with adjustments: drop the team fallback, keep the warning, extend the settings page to handle multiple installations.

## Test plan

- [x] No personal GitHub: switch task input to **Cloud** → amber notice appears → Connect opens GitHub install URL.
- [x] One installation: Settings → GitHub shows one row → Disconnect → row disappears, cloud notice reappears in task input.
- [x] Multiple installations: Connect another account → two rows render → disconnect one leaves the other intact.
- [x] Saga unchanged: cloud task creation still sends \`github_user_integration\`; signal-report tasks still send \`github_integration\`.

---

### DIFF
--- a/apps/code/src/renderer/features/settings/components/sections/GitHubSettings.tsx
+++ b/apps/code/src/renderer/features/settings/components/sections/GitHubSettings.tsx
@@ -70,8 +70,7 @@ export function GitHubSettings() {
     <Flex direction="column" gap="3">
       <Flex align="center" justify="between" gap="3" wrap="wrap">
         <Text className="text-(--gray-11) text-[13px]">
-          Personal GitHub installations linked to your PostHog account. Cloud
-          task PRs are authored as you using these.
+          Personal GitHub installations linked to your PostHog account.
         </Text>
         <Button
           size="1"
--- a/apps/code/src/renderer/features/settings/components/sections/GitHubSettings.tsx
+++ b/apps/code/src/renderer/features/settings/components/sections/GitHubSettings.tsx
@@ -1,0 +1,339 @@
+import { useOptionalAuthenticatedClient } from "@features/auth/hooks/authClient";
+import { useAuthStateValue } from "@features/auth/hooks/authQueries";
+import {
+  describeGithubConnectError,
+  invalidateGithubQueries,
+  openUrlInBrowser,
+  useGithubUserConnect,
+} from "@features/integrations/hooks/useGithubUserConnect";
+import {
+  useUserGithubIntegrations,
+  useUserRepositoryIntegration,
+} from "@hooks/useIntegrations";
+import {
+  ArrowSquareOutIcon,
+  CaretDownIcon,
+  CaretRightIcon,
+  GearSixIcon,
+  GitBranchIcon,
+  GithubLogoIcon,
+  WarningIcon,
+} from "@phosphor-icons/react";
+import {
+  AlertDialog,
+  Box,
+  Button,
+  Flex,
+  IconButton,
+  Spinner,
+  Text,
+  Tooltip,
+} from "@radix-ui/themes";
+import type { UserGitHubIntegration } from "@renderer/api/posthogClient";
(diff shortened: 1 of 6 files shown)


apps/code/src/renderer/features/settings/components/sections/GitHubSettings.tsx
```diff
@@ -0,0 +1,347 @@
...
+  return `https://github.com/settings/installations/${integration.installation_id}`;
+}
+
+async function openInBrowser(url: string) {
+  try {
+    await trpcClient.os.openExternal.mutate({ url });
+  } catch {
+    window.open(url, "_blank", "noopener,noreferrer");
+  }
+}
```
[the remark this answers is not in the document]
joshsny: nit: can we pull this out into a shared util somewhere to avoid reimplementing

apps/code/src/renderer/features/settings/components/sections/GitHubSettings.tsx
```diff
@@ -0,0 +1,347 @@
...
+  };
+
+  const connectButtonLabel =
+    integrations.length === 0 ? "Connect GitHub" : "Connect another account";
+
+  return (
+    <Flex direction="column" gap="3">
+      <Flex align="center" justify="between" gap="3" wrap="wrap">
+        <Text className="text-(--gray-11) text-[13px]">
+          Personal GitHub installations linked to your PostHog account. Cloud
```
joshsny: general question (out of the scope of this PR) - how do we handle integration selection when multiple integrations have access to a repo?

apps/code/src/renderer/features/task-detail/components/CloudGithubMissingNotice.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Flex align="center" gap="2" justify="between" wrap="wrap">
+        <Flex align="start" gap="2" className="min-w-0 flex-1">
+          <Callout.Icon>
+            <InfoIcon size={14} />
+          </Callout.Icon>
+          <Callout.Text>
+            <Text size="1">
+              {hasError
+                ? describeGithubConnectError(error)
+                : "Connect your personal GitHub to create cloud tasks. Cloud PRs are authored as you."}
```
joshsny: we can probably drop "cloud PRs are authored as you" - since this is the default experience users would expect

joshsny (approved):
LGTM

apps/code/src/renderer/features/settings/components/sections/GitHubSettings.tsx
VojtechBartos: If 2 integrations expose same repo we pick whichever resolve first. Kinda good enough, but we could in future extend the picker and allow users choose which integration they would like to use

apps/code/src/renderer/features/task-detail/components/CloudGithubMissingNotice.tsx
VojtechBartos: Updated

PR #2031, opened 2026-05-05
feat(agent): expose agent version to UI for capability gating

VojtechBartos:
## Problem

Cloud sandbox agents can run a different (typically older) version than the desktop UI expects. We've had multiple incidents where the UI assumed a notification, field, or behaviour that the cloud agent was too old to emit, and the only fallback today is racing a cloud agent rollout to match desktop expectations.

There's currently no way for the renderer to know which agent version it's talking to — the agent only logs its version locally and stamps it onto outbound user-agent strings.

## Solution

Plumb the agent's semver from the agent process to the renderer and provide a small range-based gating helper.

- **Agent** emits `agentVersion` on the `_posthog/run_started` notification (and on the SSE `connected` event for symmetry). Putting it on `run_started` is deliberate — that notification is persisted to the session log and replayed on warm sandbox reconnects, so the renderer re-learns the version after the sandbox restarts.
- **Renderer** captures the version into the session store, exposes it via `useAgentVersion(taskId)`, and adds `useIsAgentVersion(taskId, range)` / `isAgentVersion(version, range)` for gating. Backed by the `semver` package so the full range grammar works:
  ```ts
  const canDoNewThing = useIsAgentVersion(taskId, ">=0.40.1");
  const canDoNewThing = useIsAgentVersion(taskId, ">=0.40.0 <1.0.0");
  ```
  Fails closed when the agent version is unknown, so an unknown agent never accidentally enables a newer code path.

## Local runs

This works for local sessions too — the version-capture branch is unconditional, only the cloud-specific `status: "connected"` flip stays gated on `isCloud`. So the same `useIsAgentVersion(taskId, …)` call works regardless of where the session runs, which gives us consistent gating behaviour across local/cloud (useful for dev builds where someone might be running a stale local agent).

---


pnpm-lock.yaml
```diff
@@ -5184,8 +5190,8 @@ packages:
   '@types/better-sqlite3@7.6.13':
     resolution: {integrity: sha512-NMv9ASNARoKksWtsq/SHakpYAYnhBrQgGD8zkLYk/jaK8jUGn08CfEdTRgYhMypUQAfzSP8W6gNLe0q19/t4VA==}
 
-  '@types/bun@1.3.12':
-    resolution: {integrity: sha512-DBv81elK+/VSwXHDlnH3Qduw+KxkTIWi7TXkAeh24zpi5l0B2kUg9Ga3tb4nJaPcOFswflgi/yAvMVBPrxMB+A==}
+  '@types/bun@1.3.13':
+    resolution: {integrity: sha512-9fqXWk5YIHGGnUau9TEi+qdlTYDAnOj+xLCmSTwXfAIqXr2x4tytJb43E9uCvt09zJURKXwAtkoH4nLQfzeTXw==}
```
tatoalo: is this needed?

tatoalo (approved):
"Agent emits agentVersion on the _posthog/run_started notification (and on the SSE connected event for symmetry)"

Can we remove it? no need to have symmetry if that's just dead code

VojtechBartos:
@tatoalo both things addressed, thanks!

PR #2032, opened 2026-05-05
fix(agent): keep idle event streams alive

tatoalo:
## Problem

cloud task event streams can sit idle while the sandbox is still healthy. The backend reader has an inactivity timeout, so idle `/events` sockets can disconnect and surface as `Cloud stream disconnected` even though retry immediately reconnects

## Changes

* emitting SSE ka-frames from the agent `/events` endpoint every 25 seconds, keeping the socket active at the transport layer and are ignored by SSE event parsers, so no task event or workflow heartbeat is created

### DIFF
--- a/packages/agent/src/server/agent-server.ts
+++ b/packages/agent/src/server/agent-server.ts
@@ -73,6 +73,8 @@ const errorWithClassificationSchema = z.object({
 
 type MessageCallback = (message: unknown) => void;
 
+const SSE_KEEPALIVE_INTERVAL_MS = 25_000;
+
 class NdJsonTap {
   private decoder = new TextDecoder();
   private buffer = "";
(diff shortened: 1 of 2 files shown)


packages/agent/src/server/agent-server.ts
```diff
@@ -329,41 +331,73 @@ export class AgentServer {
...
 
-          if (!this.session || this.session.payload.run_id !== payload.run_id) {
-            await this.initializeSession(payload, sseController);
-          } else {
-            this.session.sseController = sseController;
-            this.session.hasDesktopConnected = true;
-            this.replayPendingEvents();
-          }
+          keepaliveInterval = setInterval(() => {
+            enqueueSseFrame(": keepalive\n\n");
```
skoob13, line 375: I didn't know that it's conventional to omit the message prefix🤯

skoob13 (approved):
LGTM

packages/agent/src/server/agent-server.ts
tatoalo: ahaha yeah, SSE have specific rules there in the parser, here we basically want to send bytes over the idle socket to avoid read timeouts, if we were to use the other `data:` convention, that would create a real SSE data message instead

PR #2038, opened 2026-05-05
feat(code): add support for additional directories

jonathanlab:
We did not really support additional directories, now we do

`/add-dir` lets you specify additional directories for each task

Default directories can be managed via the settings screen.

image
image
image
image

---


adboio (approved):
thank you! stamping bc this looks good and no major concerns, but personally i think i'd prefer a UI button to do this, without actually sending a message, like a little folder icon in the chat box or something

it felt a little unintuitive with the slash command -> dialog -> dialog -> send message

but not blocking at all, and we can always add a UI button on top of the work you've done here if we want

good stuff :saluting_face:

jonathanlab:
> thank you! stamping bc this looks good and no major concerns, but personally i think i'd prefer a UI button to do this, without actually sending a message, like a little folder icon in the chat box or something
> 
> it felt a little unintuitive with the slash command -> dialog -> dialog -> send message
> 
> but not blocking at all, and we can always add a UI button on top of the work you've done here if we want
> 
> good stuff 🫡

yeah we should probably have an item in the attachment menu for this as well, we can do that in a follow up PR!

PR #2042, opened 2026-05-05
fix(agent): change logs to debug

joshsny:
these show in the renderer as they are info logs, change them to debug logs


charlesvien (approved):
LGTM

PR #2045, opened 2026-05-05
feat(code): don't show cloud as option if git integration not existing

k11kirky:
## Problem

Two related cloud session bugs needed fixing:

1. When a cloud run fails before the agent ever boots (e.g. sandbox provisioning failure), sending a new prompt would attempt to resume the run, spinning up another sandbox that hits the same failure in a loop. The error was never surfaced to the user.

2. When a user without a GitHub integration had previously used cloud mode, the workspace mode selector would briefly show "cloud" as selected on load before snapping to a local mode once the integration check resolved, causing a visible flicker.

## Changes

**Pre-boot cloud failure guard:** `sendPrompt` now checks whether a failed cloud session never reached a connected state. If so, it throws immediately with the stored error message (or a generic fallback) rather than attempting to resume, preventing the provisioning failure loop.

**Workspace mode flicker fix:** The `workspaceMode` value is now derived after the GitHub integration check resolves. While the integration list is still loading, the UI stays optimistic and keeps the current selection. Once loaded, if cloud is unavailable and the last-used mode was `"cloud"`, it falls back to the last-used local workspace mode. The cloud option in `WorkspaceModeSelect` is also hidden when `cloudAvailable` is `false`.

### DIFF
--- a/apps/code/src/renderer/features/sessions/service/service.test.ts
+++ b/apps/code/src/renderer/features/sessions/service/service.test.ts
@@ -2649,6 +2649,96 @@ describe("SessionService", () => {
       );
     });
 
+    const mockPreBootFailedSession = (overrides: Partial<AgentSession> = {}) =>
+      mockSessionStoreSetters.getSessionByTaskId.mockReturnValue(
+        createMockSession({
+          isCloud: true,
+          cloudStatus: "failed",
+          status: "disconnected",
+          ...overrides,
+        }),
+      );
+
+    it("refuses to resume when the previous run failed before the agent booted", async () => {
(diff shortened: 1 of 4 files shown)


k11kirky:
* **#2045** <a href="https://app.graphite.com/github/pr/PostHog/code/2045?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2045?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joshsny (approved):

PR #2046, opened 2026-05-05
Fix(mcp): "set all" button now follows active search

w3joe:
## Problem

When a user searched for tools by name and clicked "Set all", the bulk approval action applied to **all** tools in the installation, not just the filtered results. This made it impossible to bulk-approve a subset of tools by name prefix (e.g. all `create_*` tools from a Linear MCP server).

Closes #2043 

## Changes

- `useMcpInstallationTools`: `setBulkApproval` now accepts an optional `targetTools` array. When provided, only those tools are updated; otherwise falls back to all tools (preserving existing behaviour).
- `ServerDetailView`: when a search is active, the "Set all" buttons pass `filteredTools` as the target. Tooltips update to say "Approve filtered" / "Require approval for filtered" / "Block filtered" to make the scope clear to the user.

## How did you test this?

Manually tested in the app: Settings → MCP Servers → connected server with >5 tools → typed a prefix in the search box → clicked each "Set all" button and confirmed only the matching tools were updated.

## Publish to changelog?

No

### DIFF
--- a/apps/code/src/renderer/features/mcp-servers/components/parts/ServerDetailView.tsx
+++ b/apps/code/src/renderer/features/mcp-servers/components/parts/ServerDetailView.tsx
@@ -252,35 +252,58 @@ export function ServerDetailView({
               <Text color="gray" className="text-[13px]">
                 Set all:
               </Text>
-              <Tooltip content="Approve all">
+              <Tooltip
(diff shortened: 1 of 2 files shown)


jonathanlab (approved):
Great fix, thanks a lot!!

w3joe:
No probs!

PR #2054, opened 2026-05-06
chore(code): merge environments + cloud environments settings

VojtechBartos:
## Problem

Two sidebar entries — **Environments** and **Cloud environments** — that sound the same but configure different things (per-project setup recipes vs. account-wide sandbox profiles).

## Changes

- Merged into a single **Environments** page with a `Local` / `Cloud` SegmentedControl.
- Sharper per-segment copy explaining what each is.
- Both old setting IDs still route here, so existing deep links keep working.
- Form mode hides the page chrome (title + tabs) and the cloud form gets a back button to match the local one.

## Showcase

https://github.com/user-attachments/assets/739288bf-0d7e-4310-911b-a32f57f63dbf

---

### DIFF
--- a/apps/code/src/renderer/features/settings/components/SettingsDialog.tsx
+++ b/apps/code/src/renderer/features/settings/components/SettingsDialog.tsx
@@ -14,13 +14,12 @@ import {
   ArrowLeft,
   ArrowsClockwise,
   CaretRight,
-  Cloud,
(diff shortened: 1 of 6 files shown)


joshsny (approved):
LGTM

tatoalo (approved):

PR #2064, opened 2026-05-06
feat(code): add pganalyze as a signal source

rafaeelaudibert:
## Summary

- Mirrors [PostHog/posthog#57752](https://github.com/PostHog/posthog/pull/57752) into the code app so the new pganalyze warehouse source shows up in the inbox alongside GitHub, Linear, and Zendesk.
- New inline `PgAnalyzeIcon` SVG that inherits `currentColor` so it adapts to light/dark mode like the Phosphor icons (matches the PR's approach for the inbox surface).
- Added pganalyze to the source filter menu, the External Connections toggle column, the `SignalSourceConfig`/`SourceProduct` unions, the warehouse-source map, the `SignalCard` source-line label, and a new `PgAnalyzeSetup` form (org slug + API key) in `DataSourceSetup`.

## Test plan

- [x] `pnpm --filter code typecheck` clean
- [x] `pnpm lint` clean
- [x] `pnpm --filter code test --run inboxSignalsFilterStore` (10/10 pass)
- [x] `pnpm --filter code test --run filterReports` (17/17 pass)
- [x] Spot-check the toggle, filter chip, and setup form render correctly in light + dark mode against a project where the upstream PostHog PR is deployed


apps/code/src/renderer/features/inbox/hooks/useSignalSourceManager.ts
```diff
@@ -267,7 +273,12 @@ export function useSignalSourceManager() {
   );
 
   const handleSetup = useCallback((source: keyof SignalSourceValues) => {
-    if (source === "github" || source === "linear" || source === "zendesk") {
+    if (
+      source === "github" ||
+      source === "linear" ||
+      source === "zendesk" ||
+      source === "pganalyze"
+    ) {
```
jonathanlab, line 281: would be slick if we could extract this pattern into some `DATA_WAREHOUSE_SOURCES` and then use a predicate instead

apps/code/src/renderer/features/inbox/hooks/useSignalSourceManager.ts
```diff
@@ -295,7 +306,9 @@ export function useSignalSourceManager() {
       if (enabled && product in DATA_WAREHOUSE_SOURCES) {
         const hasExternalSource = !!findExternalSource(product);
         if (!hasExternalSource) {
-          setSetupSource(product as "github" | "linear" | "zendesk");
+          setSetupSource(
+            product as "github" | "linear" | "zendesk" | "pganalyze",
```
jonathanlab, line 310: and then we can also type this

jonathanlab (approved):
Note to maintainers: dont merge until BE is merged

PR #2075, opened 2026-05-06
fix: default reasoning to high and respect last-used choice

k11kirky:
## Summary

The default reasoning effort for the claude adapter had been lowered to "medium". This restores the prior "high" default and persists the user's last-selected reasoning so subsequent task creations respect that choice.

- Server-side preview config now returns `currentValue: "high"` for both adapters (`apps/code/src/main/services/agent/service.ts`).
- Adds `lastUsedReasoningEffort` to the settings store and persists it on every reasoning-level change in the task input.
- `usePreviewConfig` now reads `lastUsedReasoningEffort` on initial load and on model changes, validating it against the available options for the active model and falling back to "high" when the persisted value isn't supported (e.g. switching from a model that supports `max` to one that doesn't, or switching adapters).

## Test plan
- [x] `pnpm --filter code test` (1132 passed)
- [x] `pnpm lint`
- [ ] Manual: open task input, verify default reasoning is "High"
- [ ] Manual: change reasoning to "Low", create task, return to task input — should still show "Low"
- [ ] Manual: switch from claude (with effort = "max") to codex — should fall back to "high" since "max" isn't a codex option
- [ ] Manual: switch model within claude to one that doesn't support effort — selector should disappear cleanly

---

### DIFF
--- a/apps/code/src/main/services/agent/service.ts
+++ b/apps/code/src/main/services/agent/service.ts
@@ -1749,7 +1749,7 @@ For git operations while detached:
         id: adapter === "codex" ? "reasoning_effort" : "effort",
         name: adapter === "codex" ? "Reasoning Level" : "Effort",
         type: "select",
-        currentValue: adapter === "codex" ? "high" : "medium",
+        currentValue: "high",
         options: effortOpts,
         category: "thought_level",
         description:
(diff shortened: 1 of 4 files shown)


joshsny (approved):
looks okay, surprising amount of code to change to change this, but would be very happy to be rid of the medium default 🤔

PR #2078, opened 2026-05-07
fix(agent): Request-limit-unrecoverable-thread

lost-particles:
Fixes: #2048 

## Problem

Sometimes the prompt path was putting **whole file bodies into the user message payload** instead of only referencing paths:

- When an ACP **`resource`** chunk used a **`file://`** URI, our conversion logic was taking **`resource.text`** and embedding it as a `<context>...</context>` block in the Claude Agent SDK user message. That meant a large (or mistakenly huge) attachment could blow the **serialized HTTP request size** before the agent could even behave “normally.”
- The **`buildCloudPromptBlocks`** test/harness helper was also **reading text off disk** and emitting **`resource`** blobs with inlined (truncated) content — structurally the same problem, and inconsistent with how production cloud flow stays path/artifact-centric.

This is independent of **`READ_TOOLS` / permission labels** in our agent package: those only classify tool names like **`Read`** ([Claude Code tools reference — **`Read`](https://code.claude.com/docs/en/tools-reference)). They don’t fetch files; the model invokes **`Read`** at runtime.

Anthropic’s **Request too large** doc explains pressure from big payloads and recommends patterns like chunked reads and PDF **`pages`** rather than dumping everything into one turn ([Request too large](https://code.claude.com/docs/en/errors#request-too-large)).

Closes: #2048 

## Changes

1. **`promptToClaude` (`acp-to-sdk.ts`)**  
   For **`file://`** on **`resource_link`** and **`resource`**, I only emit short **path-first** copy that aligns with Claude Code’s **`Read`** **`FileReadInput`**: **`file_path`**, optional **`offset`** / **`limit`** for large text, and optional **`pages`** (string ranges, e.g. `"1-5"`) for PDFs ([Agent SDK TypeScript — **Read**](https://code.claude.com/docs/en/agent-sdk/typescript#read)).  
   For **`resource` + `file://`**, I **stop forwarding `resource.text`** into `<context>` so workspace files can’t be inlined that way anymore.

2. **`file://` `image` blocks without base64 `data`**  
   I map those to the same **path + Read** style text instead of dropping them or silently doing nothing useful.

3. **Non filesystem schemes**  
   Where we still rely on inlined payloads for tests/other flows (e.g. **`attachment://`**), I keep **link + `<context>`** behavior so those cases don’t regress.

4. **`buildCloudPromptBlocks` (harness)**  
   Text attaches become **`resource_link`** with **`file://`** URIs (same idea as production **`cloudPromptToBlocks`**), **without** **`readAbsoluteFile`** for TXT. Images in that harness stay **embedded `image`** (separate topic from `resource`). Production cloud stays **uploads + `artifact_ids`**.

5. **Tests**  
   I added/updated Vitest coverage so **`file://` + huge `resource.text`** never appears in the SDK message, **`attachment://`** inline behavior still works, and the cloud harness emits **`resource_link`** for text.

## How did you test this?

- `pnpm test`
- `pnpm --filter @posthog/agent test`
- `pnpm --filter code test`
- `pnpm --filter @posthog/git test`

### DIFF
--- a/apps/code/src/renderer/features/editor/utils/cloud-prompt.test.ts
+++ b/apps/code/src/renderer/features/editor/utils/cloud-prompt.test.ts
@@ -117,6 +117,27 @@ describe("cloud-prompt", () => {
     expect(mockFs.readAbsoluteFile.query).not.toHaveBeenCalled();
   });
 
+  it("encodes Windows drive paths as file URIs", async () => {
+    const blocks = await buildCloudPromptBlocks(
+      'read <file path="C:\\\\tmp\\\\100%\\\\a#b?.txt" />',
+    );
+
+    const uris = resourceLinksFrom(blocks);
+    expect(uris).toHaveLength(1);
+    // C:\tmp\100%\a#b?.txt → file:///C:/tmp/100%25/a%23b%3F.txt
+    expect(uris[0]).toBe("file:///C:/tmp/100%25/a%23b%3F.txt");
+  });
+
+  it("encodes Windows UNC paths as file URIs", async () => {
+    const blocks = await buildCloudPromptBlocks(
+      'read <file path="\\\\\\\\server\\\\share\\\\My Folder\\\\file.txt" />',
+    );
+
+    const uris = resourceLinksFrom(blocks);
+    expect(uris).toHaveLength(1);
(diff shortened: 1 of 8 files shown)


lost-particles:
@jonathanlab  please review this PR

charlesvien (approved):

PR #2081, opened 2026-05-07
fix: default cloud tasks to personal github auth

tatoalo:
## Problem

Manual cloud task creation could default to bot-authored GitHub writes when the client did not resolve a selected repository to a personal GitHub integration. That made cloud runs depend on the project-level GitHub installation even when the user had linked a personal installation for the repository.

### DIFF
--- a/apps/code/src/renderer/sagas/task/task-creation.test.ts
+++ b/apps/code/src/renderer/sagas/task/task-creation.test.ts
@@ -152,7 +152,7 @@ describe("TaskCreationSaga", () => {
       model: "gpt-5.4",
       reasoningLevel: "high",
       sandboxEnvironmentId: undefined,
-      prAuthorshipMode: "bot",
+      prAuthorshipMode: "user",
       runSource: "manual",
       signalReportId: undefined,
       initialPermissionMode: "auto",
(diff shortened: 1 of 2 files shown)


apps/code/src/renderer/features/task-detail/hooks/useTaskCreation.ts
```diff
@@ -94,7 +94,7 @@ function prepareTaskInput(
     sandboxEnvironmentId: options.sandboxEnvironmentId,
     cloudPrAuthorshipMode:
       options.signalReportId && options.workspaceMode === "cloud"
-        ? "user"
+        ? "bot"
```
VojtechBartos: why from user to bot?

VojtechBartos (approved):

apps/code/src/renderer/features/task-detail/hooks/useTaskCreation.ts
VojtechBartos: ignore me, seeing it's the signal stuff

tatoalo: it is signal yep but let's not change it here, reverting anyway

PR #2103, opened 2026-05-08
feat(code): project pickers on quill autocomplete

adamleithp:
## Summary
- Migrate `ProjectSwitcher` (sidebar "Change project" dialog) and `ProjectSelect` (onboarding inline picker) from `cmdk` (via the `Command.tsx` wrapper) to quill `Autocomplete*`.
- `ProjectSwitcher` now uses quill `Dialog` + `DialogContent` (matches `CommandMenu` / `FilePicker`); orgs render as labeled groups when there are 2+ orgs, otherwise unlabeled.
- `ProjectSelect` keeps the Radix `Popover` host (anchored to the inline "change" trigger), Autocomplete is rendered inline inside.
- Both: `inline + defaultOpen + autoHighlight="always"` pattern, custom `filter` for case-insensitive name matching, `AutocompleteStatus` empty content, `CommandKeyHints` footer on the dialog variant.
- Delete the now-dead `Command.tsx` wrapper + `Command.css`, plus the cmdk-targeted `ProjectSwitcher.css` / `ProjectSelect.css`.

Stacked on #2101 (`ux/file-picker-quill`) which is stacked on #2100 (`ux/cmd-menu`). Merge order: #2100 → #2101 → this.

## What's left on cmdk
- `components/ui/combobox/Combobox.tsx` + `useComboboxFilter.ts` (standalone reusable Combobox)
- `features/folder-picker/components/GitHubRepoPicker.tsx` (only imports `defaultFilter` from cmdk)

Both can move to a follow-up PR (likely against quill's `Combobox*` namespace, separate from `Autocomplete*`).

### `ProjectSwitcher` (sidebar bottom-left → "Change project")
- [x] Click avatar/project chip → dropdown opens.
- [x] Click "Change project" → dialog opens, input autofocused, first project highlighted (autoHighlight=always).
- [x] Type a partial project name → list filters case-insensitively; "X results" status updates.
- [x] Empty-search status renders nothing (it's the result count); `No projects match "xyz"` shows when filter is empty.
- [ ] When the user belongs to multiple orgs: each org renders as a labeled group; with one org, no header.
- [ ] Current project shows ✓ icon next to its name.
- [x] ↑/↓ moves highlight, Enter selects → mutation fires, dialog closes.
- [x] Clicking a project selects it.
- [x] Esc closes the dialog in a single press.
- [x] Reopening the dialog clears the previous query.

### `ProjectSelect` (onboarding "change" link next to project name)
- [ ] Visible only when there are 2+ projects available (single project: plain text, no trigger).
- [ ] Click "change" → popover opens below, input autofocused.
- [ ] Type → filters case-insensitively.
- [ ] ↑/↓ + Enter selects, popover closes, `onProjectChange` fires with the picked id.
- [ ] Clicking outside the popover closes it.
- [ ] Esc closes in a single press.
- [ ] Disabled state: `disabled` prop dims the trigger and prevents clicks.

### Regressions / general
- [x] CommandMenu (Cmd+K) still works.
- [x] FilePicker (Cmd+P) still works.
- [x] Sidebar dropdown menu (Settings, Logout, Discord, Learn more, etc.) renders normally — only the "Change project" path was touched.
- [x] `pnpm --filter code typecheck` clean.


jonathanlab (commented):
Nice! Seems like these components have a bunch of duplicated logic w.r.t. input handling (see `handleOpenChange` and `handleSelect`), could we deduplicate that?

apps/code/src/renderer/features/onboarding/components/ProjectSelect.tsx
```diff
@@ -68,39 +76,54 @@ export function ProjectSelect({
...
+              autoFocus
+              showClear
+            />
+            <AutocompleteStatus
+              emptyContent={
+                <span>
+                  No projects match <strong>"{query}"</strong>
+                </span>
+              }
+            />
```
jonathanlab, line 116: Can we make sure we handle the case where `{query}` is ""/nullish here? it'll just show as `No projects match ""` in that case

jonathanlab (approved):

PR #2107, opened 2026-05-08
feat: add per-task Revisit toggle and sidebar filter

### LINKED ISSUE #76245: add revisit toggle to come back to a chat
### Problem
Users managing multiple concurrent tasks often lose track of items requiring further review when interrupted. Incomplete thoughts are abandoned, and tasks are prone to getting missed or forgotten when attention shifts elsewhere.

### Proposed Solution
Add a toggle that says "revisit later" (or similar).

**Expected Behavior:**
* When activated, a visual indicator (e.g., a tiny yellow dot) displays on that particular task in the tasks panel.
* Users gain an easy way to identify and return to items requiring attention at a later time.
* Prevents tasks from being missed or forgotten after interruptions.

sidjainn:
## Summary

Closes PostHog/posthog#76245.

Adds a per-task "Revisit" annotation so users can mark tasks they want to come back to later, plus a sidebar filter to focus on those tasks.

- Sidebar right-click task options has a new Revisit toggle and `Cmd+Shift+M` shortcut (also listed in the keyboard shortcuts sheet).
- Sidebar chat-bubble icon fills yellow on tasks marked for revisit (does not override functional state icons — permission, generating, cloud, suspended, unread, PR/diff, pinned).
- Task list filter dropdown gains "All tasks / Revisit only" radio. Toggling back to "All tasks" returns to the default mode.
- State persists per-device via a Zustand `localStorage` store (`revisit-tasks-storage`). Mirrors the pattern of other persisted sidebar prefs; pinned tasks remain server-side-backed and untouched by this change.
- Filter narrowing applies to both pinned and unpinned task lists, so tasks pinned + marked for revisit still show under the pinned section, and non-revisit pinned tasks are hidden when the filter is on.

### Analytics

Two new PostHog events captured (also fire when toggled via keyboard shortcut):

- `Task revisit toggled` — `{ task_id, enabled }`
- `Task revisit list filter changed` — `{ filter_name, value, previous_value? }`

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm test` (1181 passed; 15 new tests across `revisitStore.test.ts` and `applyRevisitFilter.test.ts`)
- [ ] Manual (`pnpm dev:code`):
  - [ ] Open a task → Revisit toggle visible above the chat input on the right; off by default; tooltip shows the shortcut.
  - [ ] Flip on (Switch or `Cmd+Shift+M`) → sidebar chat bubble for that task turns yellow filled.
  - [ ] Reload the app → state persists.
  - [ ] Pinned / generating / suspended / PR-state tasks keep their existing icons (revisit does not override).
  - [ ] Open sidebar filter dropdown → choose "Revisit only" → list narrows to revisit-marked tasks. Choose "All tasks" → list restores.
  - [ ] Flip Revisit off on a task while the filter is "Revisit only" → task disappears from the list immediately.
  - [ ] PostHog events fire for both toggle (Switch + shortcut) and filter radio.

---


adboio:
@sidjainn thanks so much for opening this PR!!

i think this is great functionality - however, i wonder if it would be better slightly more "hidden" in the UI, so we don't have a persistent toggle on the tasks.

what do you think about instead of a toggle, adding this as an option in the right-click menu on tasks in the sidebar? so you could right click -> "mark as unread" or similar. same behavior, just without the toggle.

lmk your thoughts! :)

sidjainn:
@adboio yeah i think it is more graceful to keep the ux cleaner with all the non-critical knobs at 1 level down. i shall make the update, thanks for the feedback!

sidjainn:
@adboio  i have moved the revisit feature to right-click on the task in the sidebar. avoided the "mark as unread" copy because of 2 reasons - conflict with the green dot copy, and secondly mark as unread implies that clicking on the task will make it read but i feel revisit is more intentional and it is better to explicity remove a task from revisit list. lmk thoughts if any 
Screenshot 2026-05-08 at 11 22 25 PM

sidjainn:
There is one stylistic difference - there's a tick that shows when a task is marked for revisit. User can click on the option again to remove the task from revisit (it's quite intuitive). The other option would be to change copy, like pin and unpin, for example. i found the former better because it's easier to scan what is ticked, scanning through copy change entails higher effort.

jonathanlab (changes requested):
Hey! We've had a chat about this, and think we should just keep this scoped a bit smaller:
- A task can be marked as "Unread" (note the choice for "Unread" here instead of revisit which we find a tad confusing)
- We should not create a separate store/state for this, instead we can just update the conversation to be marked as unread. There should be logic to handle that already :) 

Thanks again for this PR, let me know if you want to adjust it yourself otherwise we're also happy to take a stab at it.

sidjainn:
hey @jonathanlab . in the instance of reusing the unread state

apps/code/src/renderer/features/sidebar/components/SidebarMenu.tsx:117 auto-runs markAsViewed(previousTaskId) on every view change. Marks viewed → clears unread instantly when user leaves. This is causing a major ux flaw - If user performs mark as unread on the current task, and moves to another task, the previous task becomes read again defeating the purpose.

If we chose to skip auto-mark-as-viewed for the "previous" task when leaving, it will interfere with the organic unread state.

What do you recommend? 
claude recommends to add a new column markedUnreadAt to the workspace row metadata: taskId, pinnedAt, lastViewedAt, lastActivityAt, etc. I am pushing an update with this approach. 

This is the behaviour now: 
- Right-click task → "Mark as unread" → yellow filled ChatCircle without any re-ordering. 
- Mark on currently-active task: persists. Yellow stays through nav-away. Clears on nav-back-to-task.
- Organic unread (agent activity, new prompt arrival on non-viewed task): green Circle, unchanged.
- Filter "Unread only" includes both.
- DB persistence cross-restart, single source of truth.

Ran pnpm typecheck, pnpm lint and pnpm test successfully. 
Let me know what you think? Thanks for reviewing!

PR #2110, opened 2026-05-08
feat(tasks): add task run event stream sender

tatoalo:
## Problem

agent server needs a resilient sender for task run live events before it can deliver events directly to the backend. It should preserve event order, tolerate retries, and avoid making agent execution wait on each individual live event write.

## Changes

- added `TaskRunEventStreamSender` for NDJSON event ingest requests to PostHog BE
- added sequence sync and rebase handling so retries can recover from already-accepted events
- added batching, request timeouts, stop-time draining, completion handshakes, oversized-event drops, and bounded backpressure
- added unit coverage tests for various scenarios

### DIFF
--- a/packages/agent/src/server/event-stream-sender.test.ts
+++ b/packages/agent/src/server/event-stream-sender.test.ts
@@ -2,39 +2,104 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import { Logger } from "../utils/logger";
 import { TaskRunEventStreamSender } from "./event-stream-sender";
 
+const STREAM_COMPLETE_CONTROL_TYPE = "_posthog/stream_complete";
+
+async function readRequestBody(init?: RequestInit): Promise<string> {
+  const body = init?.body;
+  if (!body) {
+    return "";
+  }
+  if (typeof body === "string") {
+    return body;
+  }
+  if (body instanceof Uint8Array) {
+    return new TextDecoder().decode(body);
+  }
+  if (body instanceof ReadableStream) {
+    const reader = body.getReader();
+    const chunks: Uint8Array[] = [];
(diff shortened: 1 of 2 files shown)


tatoalo:
* **#2111** <a href="https://app.graphite.com/github/pr/PostHog/code/2111?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2110** <a href="https://app.graphite.com/github/pr/PostHog/code/2110?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2110?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joshsny (commented):
I don't think we want to go the approach of batching into POST requests in production, the 50ms batching is okay-ish but the time it takes to form the connection, validate the auth in django and post it to the stream could add up to a few hundred ms, which will make the stream very choppy.  
  
I'd suggest we open a POST request and stream the body in and do that over some period (say 5 mins or until disconnect) and then we rely on the same logic of "ask for last event" when reconnecting again

tatoalo:
@joshsny good point, I changed this to avoid relying on a long-lived in-flight POST wo we have the agent server still posts directly to Django with ordered seqs, but it now uses short sequenced NDJSON request windows. That should maintain the reliability/ordering properties while avoiding proxy/tunnel resets and latency concerns from multi steps 🙏🏻

joshsny (approved):
okay looks like this is the right structure, will help us a lot, can we simplify things at all, there are quite a lot of moving pieces in this class and i think simplifying the logic as much as we can will help us

PR #2111, opened 2026-05-08
feat(tasks): wire agent server event ingest

tatoalo:
## Problem

once the backend provides a sandbox event ingest token, the agent server needs to opt into the new sender while preserving the existing local SSE behavior and task completion signaling

## Changes

- read `POSTHOG_TASK_RUN_EVENT_INGEST_TOKEN` from the agent server env
- construct the task run event stream sender when the ingest token is present
- enqueue broadcast and terminal task events into the sender while still serving existing SSE clients
- drain the sender during agent shutdown and task completion paths

### DIFF
--- a/packages/agent/src/server/bin.ts
+++ b/packages/agent/src/server/bin.ts
@@ -32,6 +32,15 @@ const envSchema = z.object({
   POSTHOG_CODE_REASONING_EFFORT: z
     .enum(["low", "medium", "high", "xhigh", "max"])
     .optional(),
+  POSTHOG_TASK_RUN_EVENT_INGEST_TOKEN: z.string().min(1).optional(),
+  POSTHOG_TASK_RUN_EVENT_INGEST_STREAM_WINDOW_MS: z
+    .string()
+    .regex(
+      /^[1-9]\d*$/,
+      "POSTHOG_TASK_RUN_EVENT_INGEST_STREAM_WINDOW_MS must be a positive integer",
+    )
+    .transform((value) => parseInt(value, 10))
+    .optional(),
 });
 
 const program = new Command();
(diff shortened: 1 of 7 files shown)


tatoalo:
* **#2111** <a href="https://app.graphite.com/github/pr/PostHog/code/2111?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2111?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2110** <a href="https://app.graphite.com/github/pr/PostHog/code/2110?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/agent/src/server/bin.ts
```diff
@@ -32,6 +32,7 @@ const envSchema = z.object({
   POSTHOG_CODE_REASONING_EFFORT: z
     .enum(["low", "medium", "high", "xhigh", "max"])
     .optional(),
+  POSTHOG_TASK_RUN_EVENT_INGEST_TOKEN: z.string().min(1).optional(),
```
joshsny, line 35: why don't we use the existing oauth token here, is there a reason to split this out into it's own auth?

tatoalo: Mainly to have cleaner mindset against the token and what it does/needs to do.
I added ^ just for event-ingestion since it has a narrower job for exactly one task run, if we were to re-use the existing oauth token, we would need to make the ingest endpoint added in the [BE stack](https://app.graphite.com/github/pr/PostHog/posthog/58135?mode=tour) also accept a broader user-scope token, we only authorize event ingestion for the single run, if you feel strongly about it we could reuse the existing one but we would then have duplicate oauth validation and some extra checks bits, no biggie though

joshsny (approved):
don't feel strongly about the token - if we're running the agent we should have a valid oauth token already so don't see the need for it but scoping to the task run is nice

PR #2122, opened 2026-05-11
fix/macos-packaged-readonly-startup

lost-particles:
Fixes: #2098

## Problem

Packaged PostHog Code users on macOS can end up launching from **App Translocation** or a **read-only volume** (e.g. DMG). The app may start partially, then updates, local tasks, or cloud flows misbehave with little in-app explanation. This hurts anyone installing from Downloads, disk images, or other flows that Gatekeeper exposes as a restricted runtime path—not only power users reading `main.log`.

Closes [#2098](https://github.com/PostHog/code/issues/2098)

## Changes

- **Early startup gate** (first step inside `app.whenReady()`, before logs, window creation, or `initializeServices()`): **packaged macOS only**. If we detect an unsafe bundle location, show a **blocking** `dialog.showMessageBoxSync` explaining read-only/translocation-style behavior, then `app.quit()`.
- **`macos-packaged-install-guard.ts`**: consolidated helpers—
  - **`AppTranslocation`** in `app.getAppPath()` or `process.execPath`.
  - **`/sbin/mount`** output: longest prefix match for resolved `exePath`; unsafe if mount options contain **`read-only`** and mount point is **not `/`** (avoids sealed-root false positives).
- **`macos-packaged-install-guard.test.ts`**: table-driven tests for mount parsing and combined `isMacosPackagedUnsafeBundleLocation`.

No renderer UI screenshots (native Electron dialog only).

## How did you test this?

- **Automated:** `pnpm --filter code exec vitest run src/main/utils/macos-packaged-install-guard.test.ts` (all tests passing when run).
- **Manual (recommended):** `pnpm --filter code package`; open **`PostHog Code.app`** from a normal writable path (should behave as before); open from a **read-only-mounted** DMG/volume or a **translocated** path—expect the blocking dialog and exit after OK. Dev **`pnpm dev`** should be unchanged (`app.isPackaged`).

### DIFF
--- a/apps/code/src/main/index.ts
+++ b/apps/code/src/main/index.ts
@@ -1,6 +1,6 @@
 import "reflect-metadata";
 import os from "node:os";
-import { app, BrowserWindow } from "electron";
+import { app, BrowserWindow, dialog } from "electron";
 import log from "electron-log/main";
 import "./utils/logger";
 import "./services/index.js";
@@ -34,6 +34,7 @@ import {
   getLogFilePath,
   readChromiumLogTail,
 } from "./utils/logger";
+import { isMacosPackagedUnsafeBundleLocation } from "./utils/macos-packaged-install-guard";
 import { createWindow } from "./window";
 
 // Single instance lock must be acquired FIRST before any other app setup
@@ -180,6 +181,31 @@ registerDeepLinkHandlers();
 initializePostHog();
 
 app.whenReady().then(async () => {
+  if (
+    process.platform === "darwin" &&
+    app.isPackaged &&
+    isMacosPackagedUnsafeBundleLocation(app.getAppPath(), process.execPath)
+  ) {
+    const appPath = app.getAppPath();
+    const exePath = process.execPath;
+    log.warn(
+      "Refusing to start: packaged app is on App Translocation or a read-only non-root volume",
+      { appPath, exePath },
+    );
+    dialog.showMessageBoxSync({
+      type: "warning",
+      title: "Read-only install location",
+      message:
+        "PostHog Code is running from a location with read-only access (for example App Translocation or a read-only disk image). The exact folder can differ depending on how you opened the app.",
+      detail:
+        "This prevents updates and tasks from running correctly. Move PostHog Code to the Applications folder (or another folder on a writable disk), quit the app completely, then open it again from the new location.",
+      buttons: ["OK"],
+      defaultId: 0,
+    });
+    app.quit();
+    return;
+  }
+
   const commit = __BUILD_COMMIT__ ?? "dev";
   const buildDate = __BUILD_DATE__ ?? "dev";
   log.info(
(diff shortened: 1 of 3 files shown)


lost-particles:
@jonathanlab Please review these changes

apps/code/src/main/utils/macos-packaged-install-guard.test.ts
jonathanlab, line 1: Please parametrize these tests <3

apps/code/src/main/index.ts
```diff
@@ -180,6 +181,31 @@ registerDeepLinkHandlers();
...
+      { appPath, exePath },
+    );
+    dialog.showMessageBoxSync({
+      type: "warning",
+      title: "Read-only install location",
+      message:
+        "PostHog Code is running from a location with read-only access (for example App Translocation or a read-only disk image). The exact folder can differ depending on how you opened the app.",
+      detail:
+        "This prevents updates and tasks from running correctly. Move PostHog Code to the Applications folder (or another folder on a writable disk), quit the app completely, then open it again from the new location.",
+      buttons: ["OK"],
```
jonathanlab: Please update "OK" to "Quit"

apps/code/src/main/index.ts
```diff
@@ -180,6 +181,31 @@ registerDeepLinkHandlers();
...
+  ) {
+    const appPath = app.getAppPath();
+    const exePath = process.execPath;
+    log.warn(
+      "Refusing to start: packaged app is on App Translocation or a read-only non-root volume",
+      { appPath, exePath },
+    );
+    dialog.showMessageBoxSync({
+      type: "warning",
+      title: "Read-only install location",
```
jonathanlab: Read-only install location > Move PostHog Code to Applications

apps/code/src/main/index.ts
```diff
@@ -180,6 +181,31 @@ registerDeepLinkHandlers();
...
+      message:
+        "PostHog Code is running from a location with read-only access (for example App Translocation or a read-only disk image). The exact folder can differ depending on how you opened the app.",
```
jonathanlab: Let's just slim it down to 
PostHog Code is running from a location with read-only access + The actual bundle root, no need to speculate.

apps/code/src/main/index.ts
```diff
@@ -180,6 +181,31 @@ registerDeepLinkHandlers();
...
+      detail:
+        "This prevents updates and tasks from running correctly. Move PostHog Code to the Applications folder (or another folder on a writable disk), quit the app completely, then open it again from the new location.",
```
jonathanlab: Lets just turn this into "After quitting, move PostHog Code to your Applications folder, then open it from there."

jonathanlab (commented):
Thanks for addressing this issue, I've left some comments!

jonathanlab (approved):
Thanks for addressing my comments, I'll get this merged ❤️

PR #2125, opened 2026-05-13
Bento onboarding splash screen

Twixes:
## Problem

I've watched 3 people onboard to PH Code in person with the current splash screen below – which is already better than before, but still each person broadly went past this without much understanding. Just results in our points - including the self-driving angle - landing weakly in the product itself (whereas [posthog.com/code](http://posthog.com/code) actually seems quite clear).

I feel very much this comes down to the intro being this raw list - it doesn't have anchor points for scanning.

Screenshot 2026-05-10 at 18 04 32@2x

## Changes

Screenshots tell a thousand words, so proposing we turn this into a bento-like splash screen. Crucially, screenshots coming in a follow-up (I'll need you guys to add some in @PostHog/team-code). Here, setting the stage with the sleek layout.

Screenshot 2026-05-12 at 16 41 28@2x

(also, tweaked the headline at the top so the font is exactly aligned with the logo)


adboio:
SEXYYYYYY

charlesvien (approved):

PR #2129, opened 2026-05-13
feat(mobile): move "New task" button to the tab bar

oliverb123:
## Summary

- Replace the text "New task" button in the Code screen header with a plus icon in the native tab bar, alongside **Code** and **Settings**.
- Wire the plus trigger to the existing `/task` modal via a focus-only trampoline route at `(tabs)/new-task.tsx`. `NativeTabs.Trigger` requires a real file-system route and doesn't expose a `tabPress` listener, so the trampoline `replace`s back to `/tasks` then `push`es the modal so dismissal returns the user to the Code tab.
- Picks up the previous TODO/commented-out `NativeTabs.Trigger` block in `(tabs)/_layout.tsx` and replaces it with a working version.
- The empty-state "Create task" button inside `TaskList` is intentionally left in place — only the header button moves.

## Test plan

- [ ] `pnpm install && pnpm --filter @posthog/mobile lint` passes
- [ ] iOS simulator: tab bar shows **Code**, **Settings**, **+**; tapping **+** opens the new-task modal; dismissing returns to the Code tab; tapping **+** a second time still works
- [ ] Android emulator: same as above (drawable `ic_menu_add` renders)
- [ ] Code screen header no longer has the "New task" text button
- [ ] Empty-state "Create task" button inside `TaskList` still navigates correctly
- [ ] Toggling the `aiChatEnabled` preference still shows/hides the Chats tab alongside the new plus trigger

### DIFF
--- a/apps/mobile/src/app/(tabs)/_layout.tsx
+++ b/apps/mobile/src/app/(tabs)/_layout.tsx
@@ -66,7 +66,7 @@ export default function TabsLayout() {
 
       {/* New task — opens the new-task modal via a trampoline route */}
       <NativeTabs.Trigger name="new-task">
-        <Label hidden />
+        <Label>New task</Label>
         <Icon
           sf={{ default: "plus", selected: "plus" }}
           drawable="ic_menu_add"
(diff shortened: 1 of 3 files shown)


dmarticus (approved):
i have a lot of problems with this but i'm not one to call the kettle black

charlesvien:
@oliverb123 do we wanna get this one in still

charlesvien (approved):

PR #2139, opened 2026-05-13
fix(code): Restore named leading-* utilities to fix oversized badges

Twixes:
## What

Tag badges (`[Alpha]`, `P1`, etc.) ballooned from ~17px to ~26px tall.

## Why

`radix-themes-tw`'s preset *replaces* (not extends) Tailwind's `lineHeight` scale, so `leading-tight` / `leading-snug` / `leading-normal` / `leading-relaxed` / `leading-loose` silently stopped compiling. `Badge`'s `!leading-tight` override became a no-op, and Radix's `--line-height-1` (overridden to 20px in `globals.css`) took over.

## Fix

Re-add the named tokens via `theme.extend.lineHeight` so they merge with the Radix scale instead of replacing it.

Also: one-line `SessionView` cleanup to unblock `pnpm typecheck` (TS2451 on `main` from #1971).

---

### DIFF
--- a/apps/code/src/renderer/features/sessions/components/SessionView.tsx
+++ b/apps/code/src/renderer/features/sessions/components/SessionView.tsx
@@ -250,7 +250,6 @@ export function SessionView({
     [onSendPrompt],
   );
 
-  const { isOnline } = useConnectivity();
   const handleBeforeSubmit = useCallback(
     (text: string, clearEditor: () => void): boolean => {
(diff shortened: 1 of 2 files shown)


adamleithp (approved):
I think this happened because of Quill, and were competing for tailwind defaults. I'll approve, but I don't think this is the answer, we should move away from these badges and use quill (and modify quill for these variants you want)

PR #2143, opened 2026-05-13
feat(code): render PNGs and image data URLs in previews

pauldambra:
## Summary

- PNG files read through the **Read** tool no longer get dumped into CodeMirror as binary garbage — the agent returns image content blocks and `ReadToolView` now renders them inline.
- File previews and inline code previews now detect when content is a single image data URL (`data:image/png;base64,…`, etc.) and render it as an image instead of as raw text.
- All rendering goes through a new `SafeImagePreview` that validates the mime type against an allowlist (no SVG, since SVG can carry script payloads), caps the payload at 20 MB, and falls back to a friendly message via an `ErrorBoundary` + `` so a malformed or malicious value can't crash the preview.

## Changes

- `apps/code/src/shared/utils/imageDataUrl.ts` — strict data-URL parser with mime allowlist + size guard, plus unit tests.
- `apps/code/src/renderer/components/ui/SafeImagePreview.tsx` — safe `` wrapper with onError + ErrorBoundary fallback.
- `apps/code/src/renderer/features/sessions/components/session-update/toolCallUtils.tsx` — new `getContentImage` helper to extract ACP image content blocks.
- `apps/code/src/renderer/features/sessions/components/session-update/ReadToolView.tsx` — renders image content blocks via `SafeImagePreview` when present.
- `apps/code/src/renderer/features/sessions/components/session-update/CodePreview.tsx` — renders a data-URL image when the content is a single image data URL.
- `apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx` — uses `SafeImagePreview` for both PNG files and files whose content is a data URL.

## Test plan

- [x] `pnpm --filter code test --run` — all 1200 tests pass, including 16 new parser tests.
- [x] `pnpm --filter code typecheck` — clean.
- [x] `pnpm lint` — clean.
- [ ] Manual: open a `.png` file in the file viewer → renders as image.
- [ ] Manual: open a text file whose only content is \`data:image/png;base64,…\` → renders as image.
- [ ] Manual: have the agent run \`Read\` on a PNG → expanded tool view shows the image.
- [ ] Manual: corrupt a data URL (bad base64 / disallowed mime / SVG) → falls back gracefully, no crash.

---

### DIFF
--- a/apps/code/src/renderer/components/ui/SafeImagePreview.tsx
+++ b/apps/code/src/renderer/components/ui/SafeImagePreview.tsx
@@ -1,0 +1,67 @@
+import { ErrorBoundary } from "@components/ErrorBoundary";
+import { Flex, Text } from "@radix-ui/themes";
+import {
+  buildImageDataUrl,
+  isAllowedImageMimeType,
+} from "@shared/utils/imageDataUrl";
+import { useState } from "react";
+
+interface SafeImagePreviewProps {
+  /** Base64-encoded image data (no data URL prefix). */
+  base64: string;
+  mimeType: string;
+  alt?: string;
+  className?: string;
+  /** Rendered when the image fails to decode or has a disallowed mime type. */
+  fallback?: React.ReactNode;
+}
+
+function DefaultFallback() {
+  return (
+    <Flex
+      align="center"
+      justify="center"
+      className="size-full min-h-12 p-3 text-(--gray-11)"
+    >
+      <Text className="text-[13px]">Unable to render image preview</Text>
+    </Flex>
+  );
+}
+
+function SafeImagePreviewInner({
+  base64,
+  mimeType,
+  alt,
+  className,
+  fallback,
+}: SafeImagePreviewProps) {
+  const [hasError, setHasError] = useState(false);
+
+  if (!isAllowedImageMimeType(mimeType) || base64.length === 0) {
+    return <>{fallback ?? <DefaultFallback />}</>;
+  }
+
+  if (hasError) {
+    return <>{fallback ?? <DefaultFallback />}</>;
+  }
+
+  return (
+    <img
+      src={buildImageDataUrl(mimeType, base64)}
+      alt={alt ?? "image preview"}
+      className={className ?? "max-h-full max-w-full object-contain"}
+      onError={() => setHasError(true)}
+    />
+  );
+}
+
+export function SafeImagePreview(props: SafeImagePreviewProps) {
+  return (
+    <ErrorBoundary
+      name="safe-image-preview"
+      fallback={props.fallback ?? <DefaultFallback />}
+    >
+      <SafeImagePreviewInner {...props} />
+    </ErrorBoundary>
+  );
+}
--- a/apps/code/src/renderer/components/ui/SafeImagePreview.tsx
+++ b/apps/code/src/renderer/components/ui/SafeImagePreview.tsx
@@ -1,8 +1,8 @@
-import { ErrorBoundary } from "@components/ErrorBoundary";
 import { Flex, Text } from "@radix-ui/themes";
 import {
   buildImageDataUrl,
   isAllowedImageMimeType,
+  MAX_IMAGE_BASE64_LENGTH,
 } from "@shared/utils/imageDataUrl";
 import { useState } from "react";
 
@@ -28,7 +28,7 @@ function DefaultFallback() {
   );
 }
 
-function SafeImagePreviewInner({
+export function SafeImagePreview({
   base64,
   mimeType,
   alt,
@@ -36,12 +36,19 @@ function SafeImagePreviewInner({
   fallback,
 }: SafeImagePreviewProps) {
   const [hasError, setHasError] = useState(false);
+  const [lastSource, setLastSource] = useState({ base64, mimeType });
 
-  if (!isAllowedImageMimeType(mimeType) || base64.length === 0) {
-    return <>{fallback ?? <DefaultFallback />}</>;
+  if (lastSource.base64 !== base64 || lastSource.mimeType !== mimeType) {
+    setLastSource({ base64, mimeType });
+    setHasError(false);
   }
 
-  if (hasError) {
+  const isPayloadValid =
+    base64.length > 0 &&
+    base64.length <= MAX_IMAGE_BASE64_LENGTH &&
+    isAllowedImageMimeType(mimeType);
+
+  if (!isPayloadValid || hasError) {
     return <>{fallback ?? <DefaultFallback />}</>;
   }
 
@@ -52,16 +59,5 @@ function SafeImagePreviewInner({
       className={className ?? "max-h-full max-w-full object-contain"}
       onError={() => setHasError(true)}
     />
-  );
-}
-
-export function SafeImagePreview(props: SafeImagePreviewProps) {
-  return (
-    <ErrorBoundary
-      name="safe-image-preview"
-      fallback={props.fallback ?? <DefaultFallback />}
-    >
-      <SafeImagePreviewInner {...props} />
-    </ErrorBoundary>
   );
 }
--- a/apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
+++ b/apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
@@ -1,4 +1,5 @@
 import { PanelMessage } from "@components/ui/PanelMessage";
+import { SafeImagePreview } from "@components/ui/SafeImagePreview";
 import { Tooltip } from "@components/ui/Tooltip";
 import { CodeMirrorEditor } from "@features/code-editor/components/CodeMirrorEditor";
 import { EnrichmentPopover } from "@features/code-editor/components/EnrichmentPopover";
@@ -16,6 +17,7 @@ import { Box, Flex, IconButton, Text } from "@radix-ui/themes";
 import { trpcClient, useTRPC } from "@renderer/trpc/client";
 import { getImageMimeType, isImageFile } from "@shared/constants/image";
 import type { Task } from "@shared/types";
+import { parseImageDataUrl } from "@shared/utils/imageDataUrl";
 
 import { useQuery } from "@tanstack/react-query";
 import { useCallback, useMemo, useState } from "react";
@@ -153,10 +155,16 @@ export function CodeEditorPanel({
         p="4"
         className="overflow-auto"
       >
-        <img
-          src={`data:${mimeType};base64,${imageQuery.data}`}
+        <SafeImagePreview
+          base64={imageQuery.data}
+          mimeType={mimeType}
           alt={filePath}
           className="max-h-[100%] max-w-[100%] object-contain"
+          fallback={
+            <PanelMessage detail={absolutePath}>
+              Failed to render image
+            </PanelMessage>
+          }
         />
       </Flex>
     );
@@ -190,6 +198,31 @@ export function CodeEditorPanel({
 
   if (fileContent.length === 0) {
     return <PanelMessage>File is empty</PanelMessage>;
+  }
+
+  const dataUrlImage = parseImageDataUrl(fileContent);
+  if (dataUrlImage) {
+    return (
+      <Flex
+        align="center"
+        justify="center"
+        height="100%"
+        p="4"
+        className="overflow-auto"
+      >
+        <SafeImagePreview
+          base64={dataUrlImage.base64}
+          mimeType={dataUrlImage.mimeType}
+          alt={filePath}
+          className="max-h-[100%] max-w-[100%] object-contain"
+          fallback={
+            <PanelMessage detail={absolutePath}>
+              Failed to render image
+            </PanelMessage>
+          }
+        />
+      </Flex>
+    );
   }
 
   if (isMarkdown) {
--- a/apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
+++ b/apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
@@ -31,6 +31,40 @@ interface CodeEditorPanelProps {
   absolutePath: string;
 }
 
+function FilePanelImagePreview({
+  base64,
+  mimeType,
+  filePath,
+  absolutePath,
+}: {
+  base64: string;
+  mimeType: string;
+  filePath: string;
+  absolutePath: string;
+}) {
+  return (
+    <Flex
+      align="center"
+      justify="center"
+      height="100%"
+      p="4"
+      className="overflow-auto"
+    >
+      <SafeImagePreview
+        base64={base64}
+        mimeType={mimeType}
+        alt={filePath}
+        className="max-h-[100%] max-w-[100%] object-contain"
+        fallback={
+          <PanelMessage detail={absolutePath}>
+            Failed to render image
+          </PanelMessage>
+        }
+      />
+    </Flex>
+  );
+}
+
 export function CodeEditorPanel({
   taskId,
   task: _task,
@@ -130,6 +164,12 @@ export function CodeEditorPanel({
     content: isImage ? null : fileContent,
   });
 
+  const dataUrlImage = useMemo(
+    () =>
+      isImage || fileContent == null ? null : parseImageDataUrl(fileContent),
+    [isImage, fileContent],
+  );
+
   if (isImage) {
     if (isCloudRun) {
       return (
@@ -146,27 +186,13 @@ export function CodeEditorPanel({
         <PanelMessage detail={absolutePath}>Failed to load image</PanelMessage>
       );
     }
-    const mimeType = getImageMimeType(absolutePath);
     return (
-      <Flex
-        align="center"
-        justify="center"
-        height="100%"
-        p="4"
-        className="overflow-auto"
-      >
-        <SafeImagePreview
-          base64={imageQuery.data}
-          mimeType={mimeType}
-          alt={filePath}
-          className="max-h-[100%] max-w-[100%] object-contain"
-          fallback={
-            <PanelMessage detail={absolutePath}>
-              Failed to render image
-            </PanelMessage>
-          }
-        />
-      </Flex>
+      <FilePanelImagePreview
+        base64={imageQuery.data}
+        mimeType={getImageMimeType(absolutePath)}
+        filePath={filePath}
+        absolutePath={absolutePath}
+      />
     );
   }
 
@@ -200,28 +226,14 @@ export function CodeEditorPanel({
     return <PanelMessage>File is empty</PanelMessage>;
   }
 
-  const dataUrlImage = parseImageDataUrl(fileContent);
   if (dataUrlImage) {
     return (
-      <Flex
-        align="center"
-        justify="center"
-        height="100%"
-        p="4"
-        className="overflow-auto"
-      >
-        <SafeImagePreview
-          base64={dataUrlImage.base64}
-          mimeType={dataUrlImage.mimeType}
-          alt={filePath}
-          className="max-h-[100%] max-w-[100%] object-contain"
-          fallback={
-            <PanelMessage detail={absolutePath}>
-              Failed to render image
-            </PanelMessage>
-          }
-        />
-      </Flex>
+      <FilePanelImagePreview
+        base64={dataUrlImage.base64}
+        mimeType={dataUrlImage.mimeType}
+        filePath={filePath}
+        absolutePath={absolutePath}
+      />
     );
   }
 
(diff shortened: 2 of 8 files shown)


apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
```diff
@@ -153,10 +155,16 @@ export function CodeEditorPanel({
         p="4"
         className="overflow-auto"
       >
-        <img
-          src={`data:${mimeType};base64,${imageQuery.data}`}
+        <SafeImagePreview
```
pauldambra: > [!NOTE]

**[convergent: qa-team/generalist-a + xp]** 🟡 MEDIUM

SVG file previews silently regress. `getImageMimeType` returns `image/svg+xml` for `.svg` files (via `shared/constants/image.ts`), and `isImageFile("a.svg")` is true — so the file routes into this `SafeImagePreview` branch, where the allowlist in `imageDataUrl.ts` rejects `image/svg+xml` and the user sees the "Failed to render image" fallback. Two modules disagree about what's an image and the user pays.

Fix one of:
- Drop `svg` from `IMAGE_MIME_TYPES` in `shared/constants/image.ts` so it falls through to the text/code path (simplest, safe).
- Render SVG through a sanitiser (DOMPurify) and an explicit non-`SafeImagePreview` path.
- Keep the previous unsafe `` path for `.svg` files only, with a clear comment about the threat model.

Deny-by-default for tool-output data URLs (the deliberate goal here) is right; the regression is specifically for explicitly-opened `.svg` files from disk.

apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
```diff
@@ -192,6 +200,31 @@ export function CodeEditorPanel({
     return <PanelMessage>File is empty</PanelMessage>;
   }
 
+  const dataUrlImage = parseImageDataUrl(fileContent);
+  if (dataUrlImage) {
+    return (
```
pauldambra, line 230: > [!NOTE]

**[xp]** 🟠 HIGH

Lines 150–170 and 205–225 are nearly byte-for-byte identical — same `Flex` wrapper, same `align/justify/height/p/className`, same `SafeImagePreview` props except for `base64` + `mimeType`, same `PanelMessage` fallback. Textbook DuplicatedCode in a single file.

Extract a local helper, something like:

```tsx
function FilePanelImagePreview({ base64, mimeType, filePath, absolutePath }) {
  return (
    <Flex align="center" justify="center" height="100%" p="4" className="overflow-auto">
      <SafeImagePreview
        base64={base64}
        mimeType={mimeType}
        alt={filePath}
        className="max-h-[100%] max-w-[100%] object-contain"
        fallback={<PanelMessage detail={absolutePath}>Failed to render image</PanelMessage>}
      />
    </Flex>
  );
}
```

…and call it from both branches. Don't try to unify across the three feature files (`CodeEditorPanel`, `CodePreview`, `ReadToolView`) — they legitimately differ at the wrapper layer.

apps/code/src/renderer/components/ui/SafeImagePreview.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function SafeImagePreviewInner({
+  base64,
+  mimeType,
+  alt,
+  className,
+  fallback,
+}: SafeImagePreviewProps) {
+  const [hasError, setHasError] = useState(false);
```
pauldambra, line 38: > [!NOTE]

**[qa-team/generalist-a]** 🟡 MEDIUM

`hasError` is initialised once via `useState(false)` and never reset when `base64` or `mimeType` change. If the same `SafeImagePreview` instance is reused with new image data (e.g. `CodeEditorPanel` switches between two image files that mount this component at the same slot), the fallback from the prior image will be shown even though the new src is valid — until the component is unmounted/remounted.

Fix options:
- Pass a stable `key={`${mimeType}:${base64.slice(0, 32)}`}` from callers (cheapest).
- Reset internally with `useEffect(() => setHasError(false), [base64, mimeType])`.
- Use a derived-state pattern (compare previous src in render and reset).

apps/code/src/renderer/features/sessions/components/session-update/toolCallUtils.tsx
```diff
@@ -68,6 +68,26 @@ export function getContentText(
   return undefined;
 }
 
+export interface ImageContentData {
```
pauldambra, line 71: > [!NOTE]

The same idea — "base64-encoded image bytes plus its mime type" — is expressed in two different shapes:
- `ImageContentData = { data: string; mimeType: string }` here
- `ParsedImageDataUrl = { base64: string; mimeType: string }` in `imageDataUrl.ts`

Pick one shape and one field name. Strong preference for `base64` — `data` doesn't say *how* it's encoded, and you'll be wiring these together across files for a while. Then `SafeImagePreview`'s props match both producers and the per-call-site re-mapping (`base64={imageContent.data}`) disappears.

apps/code/src/renderer/features/sessions/components/session-update/ReadToolView.tsx
```diff
@@ -27,9 +29,10 @@ export function ReadToolView({
 
   const filePath = locations?.[0]?.path ?? "";
   const startLine = locations?.[0]?.line ?? 0;
-  const fileContent = getReadToolContent(content);
+  const imageContent = getContentImage(content);
+  const fileContent = imageContent ? undefined : getReadToolContent(content);
```
pauldambra, line 33: > [!NOTE]

**[xp]** 🟡 MEDIUM

`const fileContent = imageContent ? undefined : getReadToolContent(content);` plus the title ternary, plus two separate `isExpanded && X &&` blocks, is the start of shotgun surgery — every new content kind (audio? pdf?) will add another `imageX ? undefined : ...` rung and another conditional block.

Replace the two getters with one that returns a discriminated union:

```ts
type ReadToolPayload =
  | { kind: "image"; mimeType: string; base64: string }
  | { kind: "text"; text: string; lineCount: number };

function getReadToolPayload(content): ReadToolPayload | undefined { ... }
```

Then the title, the expand toggle, and the body all switch on `payload.kind` once. Same total LOC, one idea expressed once.

apps/code/src/renderer/features/sessions/components/session-update/ReadToolView.tsx
```diff
@@ -53,12 +56,27 @@ export function ReadToolView({
           isExpanded={isExpanded}
         />
         <ToolTitle className="shrink-0 whitespace-nowrap">
-          Read{lineCount !== null ? ` ${lineCount} lines in` : ""}
+          {imageContent
+            ? "Read image"
+            : `Read${lineCount !== null ? ` ${lineCount} lines` : ""} in`}
```
pauldambra: > [!NOTE]

**[convergent: qa-team/copy + paul]** 🟢 LOW

Copy regression: the trailing ` in` moved outside the ternary. Before this PR:
```js
`Read${lineCount !== null ? ` ${lineCount} lines in` : ""}`
// lineCount = null -> "Read"
```
Now:
```js
`Read${lineCount !== null ? ` ${lineCount} lines` : ""} in`
// lineCount = null -> "Read in" (followed by FileMentionChip)
```

Probably impossible to actually hit in practice (no content -> not expandable), but the original was nicer. Easiest fix is keeping ` in` inside the ternary, or restructuring to:

```tsx
{imageContent
  ? "Read image"
  : lineCount !== null ? `Read ${lineCount} lines in` : "Read"}
```

apps/code/src/shared/utils/imageDataUrl.ts
```diff
@@ -0,0 +1,47 @@
...
+  "image/x-icon",
+  "image/vnd.microsoft.icon",
+  "image/tiff",
+  "image/avif",
+]);
+
+const DATA_URL_PATTERN =
+  /^data:([a-zA-Z]+\/[a-zA-Z0-9.+-]+)(?:;[a-zA-Z0-9-]+=[^;,]+)*;base64,([A-Za-z0-9+/=\s]+)$/;
+
+const MAX_BASE64_LENGTH = 20 * 1024 * 1024;
```
pauldambra: > [!NOTE]

**[convergent: qa-team/performance + xp]** 🟢 LOW

`MAX_BASE64_LENGTH` is misleading — the check at line 26 runs against the *whole trimmed data URL string* (`data:image/png;base64,` prefix + any whitespace + payload), not the base64 length. The name and the location of the check disagree.

Two cheap fixes:
- Rename to `MAX_DATA_URL_LENGTH` so the name matches the check, or
- Move the length check to *after* `match[2]` is extracted and whitespace-stripped, so the constant actually constrains decoded payload size.

While you're there: the `trim()` on line 25 allocates up to ~20 MiB before any cheap guard runs. Consider doing `startsWith("data:")` on the raw value first and only trimming if it might be a data URL. (Real previews here are screenshots, well under 5 MiB base64 — the cap could probably come down too.)

apps/code/src/renderer/components/ui/SafeImagePreview.tsx
```diff
@@ -0,0 +1,67 @@
...
+      alt={alt ?? "image preview"}
+      className={className ?? "max-h-full max-w-full object-contain"}
+      onError={() => setHasError(true)}
+    />
+  );
+}
+
+export function SafeImagePreview(props: SafeImagePreviewProps) {
+  return (
+    <ErrorBoundary
```
pauldambra: > [!NOTE]

**[convergent: qa-team/generalist-b + paul]** ⚪ NIT

Does the `ErrorBoundary` here actually catch anything? `` decode failures fire the `onError` event, they don't throw — and that's already handled by the `hasError` state below. The boundary's `captureException` will only fire if something inside `SafeImagePreviewInner` throws during render, which is unlikely for what's essentially ``.

Not blocking — defence in depth is fine — but worth either a comment explaining what it's defending against, or dropping it. As-is it's likely just extra nesting + dead defence.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments. Top-level summary follows in a separate comment.

apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
```diff
@@ -192,6 +200,31 @@ export function CodeEditorPanel({
     return <PanelMessage>File is empty</PanelMessage>;
   }
 
+  const dataUrlImage = parseImageDataUrl(fileContent);
```
pauldambra: > [!NOTE]

**[qa-team/performance]** 🟢 LOW

`parseImageDataUrl(fileContent)` runs on every render of `CodeEditorPanel` after the file loads. The common case is cheap (the `startsWith("data:")` guard short-circuits), but for any file whose content happens to start with `data:` the full regex + `replace(/\s+/g, "")` re-runs every render.

`useMemo(() => parseImageDataUrl(fileContent), [fileContent])` for consistency with how the rest of the file memoises derived state. Same issue in `CodePreview.tsx:48`.

apps/code/src/shared/utils/imageDataUrl.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export interface ParsedImageDataUrl {
+  mimeType: string;
+  base64: string;
+}
+
+export function parseImageDataUrl(value: string): ParsedImageDataUrl | null {
+  if (typeof value !== "string") return null;
+  const trimmed = value.trim();
+  if (trimmed.length === 0 || trimmed.length > MAX_BASE64_LENGTH) return null;
```
pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

Every failure path here returns `null` silently. When the size cap, mime allowlist, or regex rejects a payload, there's no log line — "why doesn't my image render?" is unanswerable from the app logs.

A `logger.debug` (or `logger.warn` for the size cap, which suggests something weird upstream) would mean we can grep when a user reports it. Not blocking, but i'm slightly allergic to silent fall-throughs in user-facing code.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (security, frontend, performance, copy, two generalists), paul-reviewer, xp-reviewer

Security posture is solid (SVG explicitly blocked, mime allowlist, length cap, no `javascript:` injection path, no ReDoS). The blockers are user-visible behavioural bugs introduced by the change: SVG file previews regress, and `SafeImagePreview`'s error state doesn't reset across image swaps. The HIGH from xp is straight-line in-file duplication that's worth fixing while you're touching this code anyway.

### Key findings

**🟠 HIGH**
- `CodeEditorPanel.tsx` lines 150–170 vs 205–225 are nearly byte-for-byte identical — extract a `FilePanelImagePreview` helper.

**🟡 MEDIUM**
- SVG file previews regress to "Failed to render image" because `shared/constants/image.ts` still maps `.svg` to `image/svg+xml`, which the new allowlist rejects.
- `SafeImagePreview` `hasError` state doesn't reset when `base64`/`mimeType` change — stale failure can stick to a new image.
- Naming drift: `ImageContentData.data` vs `ParsedImageDataUrl.base64` for the same idea.
- `ReadToolView` coordinates two getters with a primitive flag (`imageContent ? undefined : getReadToolContent(...)`) — discriminated union would express this once.

**🟢 LOW / ⚪ NIT**
- Copy regression: `"Read in"` (no line count) where it used to be `"Read"`.
- `MAX_BASE64_LENGTH` is checked against the whole trimmed string, not the base64 length — misnamed.
- `parseImageDataUrl` runs on every render of `CodeEditorPanel` and `CodePreview` — memoise.
- `ErrorBoundary` around `` doesn't catch decode failures (those use `onError`) — likely dead defence.
- All failure paths in `parseImageDataUrl` return `null` silently — at least one `logger.debug` would help diagnosis.

### Convergence (independently flagged by 2+ reviewers — highest confidence)

| Finding | Reviewers |
| --- | --- |
| SVG file preview regression | qa-team/generalist-a + xp |
| `base64` vs `data` naming drift | paul + xp |
| "Read in" copy regression | qa-team/copy + paul |
| `MAX_BASE64_LENGTH` misnamed | qa-team/performance + xp |
| `ErrorBoundary` likely redundant | qa-team/generalist-b + paul |

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Solid security primitive; functional regressions for SVG files and `hasError` stickiness need fixing; perf is fine on the common path but parse is unmemoised and length cap is over-generous. |
| 👤 paul | Nice clean addition with thorough tests; mostly questions about silent error paths and whether `ErrorBoundary` around `` is actually catching anything — approve with caveats. |
| 📐 xp | Good security primitive and good tests; main work left is removing in-file duplication in `CodeEditorPanel`, unifying the two image-payload shapes, and reconciling the two sources of truth for "is this an image?". |

---

apps/code/src/renderer/components/ui/SafeImagePreview.tsx
```diff
@@ -0,0 +1,67 @@
...
+  fallback,
+}: SafeImagePreviewProps) {
+  const [hasError, setHasError] = useState(false);
+
+  if (!isAllowedImageMimeType(mimeType) || base64.length === 0) {
+    return <>{fallback ?? <DefaultFallback />}</>;
+  }
+
+  if (hasError) {
+    return <>{fallback ?? <DefaultFallback />}</>;
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/features/sessions/components/session-update/ReadToolView.tsx
```diff
@@ -53,12 +56,27 @@ export function ReadToolView({
...
         </ToolTitle>
         {filePath && <FileMentionChip filePath={filePath} />}
         <StatusIndicators isFailed={isFailed} wasCancelled={wasCancelled} />
       </Flex>
 
+      {isExpanded && imageContent && (
+        <Box className="mt-2 ml-5">
+          <Box className="max-w-4xl overflow-hidden rounded-lg border border-gray-6 bg-(--gray-2) p-2">
+            <SafeImagePreview
+              base64={imageContent.data}
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/components/ui/SafeImagePreview.tsx
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/features/sessions/components/session-update/toolCallUtils.tsx
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/features/sessions/components/session-update/ReadToolView.tsx
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/shared/utils/imageDataUrl.ts
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/features/code-editor/components/CodeEditorPanel.tsx
pauldambra: > [!NOTE]

Addressed in 456b2992 — fixes pushed by the author in parallel.

apps/code/src/renderer/components/ui/SafeImagePreview.tsx
pauldambra: > [!NOTE]

Addressed in 456b2992 — `ErrorBoundary` removed, since `` decode failures fire `onError` rather than throwing.

apps/code/src/renderer/features/sessions/components/session-update/ReadToolView.tsx
pauldambra: > [!NOTE]

Out of scope for this PR — leaving the current two-getter shape and revisiting if a third content kind appears.

apps/code/src/shared/utils/imageDataUrl.ts
pauldambra: > [!NOTE]

Out of scope for this PR — silent fallback is acceptable for now; can add logging in a follow-up if diagnostics become a pain point.

pauldambra:
> [!NOTE]

## Updated verdict: ✅ APPROVE

All actionable findings from the [original QA Swarm review](https://github.com/PostHog/code/pull/2143#issuecomment-4445793101) have been resolved by `456b2992`:

| Finding | Severity | Status |
| --- | --- | --- |
| CodeEditorPanel duplication | 🟠 HIGH | ✅ `FilePanelImagePreview` extracted |
| SVG file preview regression | 🟡 MEDIUM | ✅ `svg` dropped from `IMAGE_MIME_TYPES` |
| `hasError` state stickiness | 🟡 MEDIUM | ✅ Reset via render-time setState |
| `data` vs `base64` naming drift | 🟡 MEDIUM | ✅ Renamed to `base64` |
| ReadToolView discriminated union | 🟡 MEDIUM | ↪️ Resolved out-of-scope |
| "Read in" copy regression | 🟢 LOW | ✅ Title restructured |
| `MAX_BASE64_LENGTH` misnamed | 🟢 LOW | ✅ Renamed to `MAX_DATA_URL_LENGTH` |
| `parseImageDataUrl` unmemoised | 🟢 LOW | ✅ `useMemo` added |
| Silent fall-through logging | 🟢 LOW | ↪️ Resolved out-of-scope |
| `ErrorBoundary` likely redundant | ⚪ NIT | ✅ Removed |

CI is green except `integration-test` (still running). The shepherd will apply the `stamphog` label once that finishes.

---
*Automated by PR Shepherd — not a human review*

k11kirky (approved):

PR #2149, opened 2026-05-14
feat(worktree): use human-readable names for new worktrees

### LINKED ISSUE #1844: Use human-readable names for worktree branches instead of random hashes
### Problem

New worktrees are created with random SHA-looking branch names. At a glance these look nearly identical to a detached HEAD state, causing momentary confusion when checking which branch you're on.

### Expected behavior

Worktree branches should use human-readable random names (e.g. city names, adjective-noun pairs from a safe word list) instead of hex-like strings. This is a pattern used by X and provides a nicer developer experience.

Image
Image

agallardol:
## Problem

Closes #1844.

Worktrees are created with 4-digit numeric folder names like `5142` — visually indistinguishable from a SHA prefix when scanning `git worktree list`. The issue suggests human-readable names; Claude Code uses `claude/<adj>-<noun>` and [sst/opencode][1] uses `opencode/<adj>-<noun>` for the same UX reason.

> _Screenshot placeholder — before/after of a freshly-created worktree._

## Changes

Three changes, bundled because #2 is exposed by #1 (splitting would chain two dependent PRs):

**1. Naming (`packages/git`).** `WorktreeManager.generateWorktreeName()` now returns `adjective-noun` from a new word list in `worktree-names.ts` (102 × 100 = **10,200 combinations**, above the previous 9,000 namespace). Adapted from sst/opencode (MIT). The 100-retry collision loop in `generateUniqueWorktreeName` is preserved; `createWorktreeForExistingBranch` and `createDetachedWorktreeAtCommit` still single-shot with a `${name}${Date.now()}` fallback (unchanged behavior, just slightly uglier output: `brave-falcon1700000000000`).

**2. Path resolution (`apps/code`).** Latent bug: `isLegacyWorktreeName(name) = !/^\d+$/.test(name)` classified every non-numeric name as "legacy" and routed it to a different folder layout. With the new slugs, this silently broke every fresh worktree lookup. Replaced the regex with an empirical probe of both supported layouts (`{base}/{name}/{repo}` modern, `{base}/{repo}/{name}` legacy), defaulting to modern when neither exists.

**3. Consolidation.** `ArchiveService` and `SuspensionService` each had a private async copy of the same probe-both pattern using `fs.access`. Both deleted; both services now use the shared helper. The shared helper is sync (`existsSync`); the swap is intentional — these call sites aren't in hot loops and the sync signature is simpler to use. Net **–52 lines** in the app layer.

## Trade-offs

- **Static word list vs. LLM-generated names.** Claude Code generates branch names via the LLM. Considered, rejected: a static list matches the issue's ask, costs nothing at runtime, and keeps this PR reviewable.
- **`existsSync` in `workspace.service.ts` loops.** Probe adds 2 stats per call; `handleGitStateChanged` multiplies by N task associations. Dwarfed by the `getBranchFromPath` git spawn already in the same iteration, so practical impact is negligible.
- **`worktreeLocation` setting change.** Probe defaults to modern (non-existent path), `fs.existsSync` removes the association. Pre-existing blind spot from before this PR.

## Test plan

23 tests across `worktree-names.test.ts`, `worktree.test.ts`, and `worktree-helpers.test.ts`. All passing under `pnpm --filter @posthog/git test` and `pnpm --filter code test`. Typecheck and Biome lint clean on touched files.

## Publish to changelog?

Yes — suggested entry: _"Worktrees now use human-readable adjective-noun names (e.g. `brave-falcon`) instead of 4-digit numbers."_

As an external contributor I can't post to #changelog directly — happy to provide more details for whoever handles it.

[1]: https://github.com/sst/opencode/blob/dev/packages/core/src/util/slug.ts


agallardol:
I'll attach an screenshot from the app UI point of view once I can get an invitation code so I can test end to end properly.

charlesvien:
Done in https://github.com/PostHog/code/pull/2340

PR #2151, opened 2026-05-16
fix(inbox): fix some inbox papercuts

joshsny:
## Summary

A bunch of small things were annoying me in the inbox UI, here are fixes for them, don't think anything here is 🌶️

### Signals inbox list
- Drop the yellow tint on suggested-reviewer rows — the amber row background read as "highlighted/selected"; the amber eye badge alone is enough to mark a card as suggested-for-you.
- Clamp list-row summary to 3 lines (was 4).

### Signal cards (report detail)
- Drop the "Weight: x.x" badge from signal card headers — noisy internal detail with no actionable meaning.
- Only render the "Verified" badge when the finding is actually verified — drop the gray "Unverified" rendering.
- Error tracking signal cards now render a "View issue" link (built from `signal.source_id` via a new `errorTrackingIssueUrl()` helper that picks up the user's cloud region and project id), mirroring the existing "View on GitHub" / Zendesk "Open" links.

### Report detail header & task bars
- Move the **Create PR / View PR** action from the bottom implementation task bar into the top-right header next to Dismiss. Header order is now: copy-link, Dismiss, (Create PR ⇄ View PR pill), close. The Create PR button and the View PR pill are mutually exclusive — same slot, same role.
- Drops the corresponding `onCreateImplementationTask` / `isAwaitingInput` plumbing from `ReportTaskLogs` and the `role="button"` workaround that only existed so the nested run button could coexist with the expand toggle.
- Shorten task-bar labels: "Research task" → "Research", "Implementation task" → "Implementation".

### Interaction
- Plain-clicking the already-selected report no longer calls `clearSelection()` (and no longer drops the user back to the "Select a report" empty state). Cmd-click still toggles; shift-click still range-selects.

Screenshot 2026-05-16 at 18 46 46

### DIFF
--- a/apps/code/src/renderer/features/inbox/components/InboxSignalsTab.tsx
+++ b/apps/code/src/renderer/features/inbox/components/InboxSignalsTab.tsx
@@ -283,18 +283,12 @@ export function InboxSignalsTab() {
         );
       } else if (event.metaKey) {
         toggleReportSelection(reportId);
-      } else if (
-        selectedReportIdsRef.current.length === 1 &&
-        selectedReportIdsRef.current[0] === reportId
-      ) {
-        // Plain click on the only selected report — deselect it
-        clearSelection();
       } else {
-        // Plain click — select only this report
+        // Plain click — select only this report (no-op if already the sole selection)
         setSelectedReportIds([reportId]);
       }
     },
-    [selectRange, toggleReportSelection, setSelectedReportIds, clearSelection],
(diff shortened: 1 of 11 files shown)


Twixes (approved):
I think this all makes sense, with one exception: Let's keep the yellow tint. We'll get rid of it when we move to a list where you _only_ see your reports by default, but we should do this in one go. Dropping the highlight while still showing everything will result in confusion. Approving with the assumption we restore the tint.

PR #2174, opened 2026-05-18
feat(signals): Add Slack notification config

Twixes:
## Problem

Signals inbox items previously surfaced only in-app. Users wanting a heads-up when a new item lands and they're a suggested reviewer had no way to opt in.

Even after the initial Slack notifications work landed, picking a channel still required a detour through PostHog Web — users had to leave PostHog Code to install the Slack integration, then come back. That's too much friction for what should be a one-time setup.

## Changes

Adding Slack notifications under **Inbox** source settings.

What it includes:

- Slack workspace picker (shown only when more than one Slack `Integration` is connected; otherwise the single workspace is implied).
- Notification channel picker, populated from `/api/environments/{teamId}/integrations/{id}/channels/` - same endpoint as Insight Alerts. Includes an explicit "Off" option.
- Minimum priority filter (also includes an "All priorities" option).

Also, in-app Slack connect flow so users no longer have to leave PostHog Code for PostHog Web.

- `SlackIntegrationService` in the main process. Mirrors the existing `GitHubIntegrationService` but registers a separate `slack-integration` deep-link key so each provider's handler stays isolated.
- `slackIntegration` tRPC router.
- The empty state in `SignalSlackNotificationsSettings` shows a "Connect Slack workspace" button that runs connection without redirecting to the Integrations page of Cloud settings.

Pairs with the backend changes in PostHog/posthog#58774.

## How did you test this?

Ran this locally.

## Publish to changelog?

no

---

_Created with_ [_PostHog Code_](https://posthog.com/code?ref=pr)


andrewm4894 (approved):
lgtm - i guess at some stage down the road we might want ability to control a bit more what projects or repos go to what slack channels etc.

or maybe even have something like a routing agent in some way if we want some sort of feature or theme related routing 

eg llma related stuff im tagged on pings my in that channel etc so rest of team can see the thread etc.

but makes sense to start simple

PR #2185, opened 2026-05-18
Fix local Codex queued messages getting stuck

fredluz:
## Problem

Local Codex queued follow-up messages can remain stuck in the conversation after the active turn completes.

Closes #2182

## Changes

`_posthog/turn_complete` can arrive through the streamed ACP subscription before the matching JSON-RPC `session/prompt` response for local Codex sessions. The renderer was clearing `currentPromptId` for all sessions when that notification arrived. When the later prompt response arrived, it was treated as stale and returned before the local queue-drain path could run.

This changes `turn_complete` prompt-state clearing to apply only to cloud sessions. Local sessions keep using the matching prompt response as the queue-drain boundary.

Added a regression test for the local Codex ordering case: queued message present, `turn_complete` arrives first, then the matching prompt response drains the queue.

## How did you test this?

Manual:
- Reproduced before the fix: start a local Codex turn, send a follow-up while pending, observe the follow-up remains stuck as queued after the first turn completes.
- Retested after the fix: same flow, queued follow-up sends automatically after the first turn completes.

Automated:

```bash
pnpm --filter code exec biome check src/renderer/features/sessions/service/service.ts src/renderer/features/sessions/service/service.test.ts
pnpm --filter code typecheck
pnpm --filter code exec vitest run src/renderer/features/sessions/service/service.test.ts
```

Also run by pre-commit hook:

```bash
biome check --write --unsafe --files-ignore-unknown=true --no-errors-on-unmatched
pnpm typecheck
```

## Publish to changelog?

no

### DIFF
--- a/apps/code/src/renderer/features/sessions/service/service.test.ts
+++ b/apps/code/src/renderer/features/sessions/service/service.test.ts
@@ -73,7 +73,7 @@ const mockSessionStoreSetters = vi.hoisted(() => ({
   enqueueMessage: vi.fn(),
   removeQueuedMessage: vi.fn(),
   clearMessageQueue: vi.fn(),
-  dequeueMessagesAsText: vi.fn(() => null),
+  dequeueMessagesAsText: vi.fn((): string | null => null),
   dequeueMessages: vi.fn(
     () =>
       [] as Array<{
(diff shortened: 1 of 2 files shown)


fredluz:
Thanks — updated.

Cloud sessions still use `_posthog/turn_complete` as the turn boundary for prompt-state clearing and queue flushing. The change only prevents local sessions from clearing `currentPromptId` on that notification, because local Codex can receive `turn_complete` before the matching prompt response.

I also added `currentPromptId: null` to `cancelPrompt()` so cancelled local prompt responses are treated as stale and cannot drain queued messages.

I manually tested the local Codex flow before and after the fix. I can’t manually run cloud tasks from my current local setup, but the existing session-service cloud queue/hydration coverage still passes.

k11kirky:
Thanks @fredluz - I actually had a fix on a branch for this already here: https://github.com/PostHog/code/pull/2197

Looks very similar so will merge and close

k11kirky:
duplicate PR https://github.com/PostHog/code/pull/2197

PR #2188, opened 2026-05-18
feat(cloud): auto-retry errored cloud streams on window focus

joshsny:
## Summary

- The main process gives up on the cloud SSE stream after 5 reconnect attempts (~60s total with exponential backoff, capped at 30s). Anything longer than that — a Django deploy, a laptop sleep, a network blip — surfaces the manual "Retry" button.
- This change adds a `focus` listener that calls `SessionService.retryUnhealthyCloudSessions()` whenever the renderer regains focus. The method enumerates every session in `error` state where `isCloud === true` and kicks off the same `cloudTask.retry` mutation the manual button uses.
- Idempotent by construction: `retryCloudTaskWatch` flips status to `disconnected` before issuing the mutation, so a second focus event in the same tick is a no-op. Failures are caught + logged so one bad retry doesn't block others.

## Files

- `apps/code/src/renderer/features/sessions/service/service.ts` — adds `retryUnhealthyCloudSessions()` after the existing `retryCloudTaskWatch`.
- `apps/code/src/renderer/components/GlobalEventHandlers.tsx` — new `useEffect` mirroring the existing `loadFolders`-on-focus pattern.
- `apps/code/src/renderer/features/sessions/service/service.test.ts` — three new tests (retries only errored cloud sessions, no-op when none are errored, swallows per-session failures).

## Test plan

- [ ] `pnpm --filter code typecheck`
- [ ] `pnpm vitest run src/renderer/features/sessions/service/service.test.ts`
- [ ] Manual: start a cloud task → kill Django (or yank wifi for >60s) → wait for the "Cloud stream disconnected" Retry button → cmd-tab away → cmd-tab back → confirm the stream reconnects without clicking Retry.
- [ ] Manual: confirm local sessions and already-connected cloud sessions are not affected by focus events.

---

### DIFF
--- a/apps/code/src/renderer/features/sessions/service/service.ts
+++ b/apps/code/src/renderer/features/sessions/service/service.ts
@@ -3150,6 +3150,30 @@ export class SessionService {
     }
   }
 
+  /**
+   * Retries every cloud session whose stream is in the `error` state, i.e. the
+   * main process exhausted its SSE reconnect budget and surfaced the manual
+   * Retry button. Invoked on window focus so users coming back to the app
+   * after a Django deploy, laptop sleep, or network blip don't have to click
+   * Retry themselves.
+   */
+  public retryUnhealthyCloudSessions(): void {
+    const sessions = sessionStoreSetters.getSessions();
+    for (const session of Object.values(sessions)) {
+      if (!session.isCloud) continue;
+      if (session.status !== "error") continue;
+      log.info("Auto-retrying errored cloud session on focus", {
+        taskId: session.taskId,
+      });
+      this.retryCloudTaskWatch(session.taskId).catch((error) => {
+        log.warn("Auto-retry of errored cloud session failed", {
+          taskId: session.taskId,
+          error,
+        });
+      });
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


apps/code/src/renderer/features/sessions/service/service.ts
```diff
@@ -3150,6 +3150,30 @@ export class SessionService {
...
+      log.info("Auto-retrying errored cloud session on focus", {
+        taskId: session.taskId,
+      });
+      this.retryCloudTaskWatch(session.taskId).catch((error) => {
+        log.warn("Auto-retry of errored cloud session failed", {
+          taskId: session.taskId,
+          error,
+        });
+      });
+    }
```
tatoalo, line 3174: this is triggered on window focus for all such sessions, usually this if fine I imagine, but let's stay that users open a lot of tasks via command center mode, we could have some hammering going on but maybe not such a big deal

tatoalo (approved):

apps/code/src/renderer/features/sessions/service/service.ts
joshsny: i think that's okay, at worst this fires 100 requests at django, should be fine

PR #2189, opened 2026-05-18
Fix review state render loop

fredluz:
## Problem

The chat panel can hit React error #185 / maximum update depth. While investigating, I found a reproducible render loop in the code review/diff state path: `useCollapseState()` was synchronizing derived deferred-path state by calling React state setters during render when `deferredPaths` changed by object identity.

Closes #2165

## Changes

- Replace render-phase deferred-path state synchronization in `ReviewShell.useCollapseState()` with an effect guarded by a semantic deferred-path key.
- Preserve reveal state when equivalent deferred paths are recreated with new object identity.
- Add a focused regression test that recreates `changedFiles` across renders and verifies the hook no longer hits React's re-render limit.

### Follow-up note

While investigating this, I also noticed a similar render-phase prop/state synchronization smell in `InteractiveFileDiff.tsx`. I’m not changing it in this PR because I don’t currently have a focused repro showing it triggers the reported crash, but it may be worth hardening separately.

## How did you test this?

I couldn’t manually test the exact scenario, but this PR includes a focused regression test that reproduces the React #185-shaped loop on main and verifies the fix.

- `pnpm lint`
- `pnpm --filter code exec vitest run src/renderer/features/code-review`
- `pnpm --filter code exec vitest run src/main/services/archive/service.integration.test.ts --testTimeout=30000 --reporter=verbose`

I also ran `pnpm test`; the full local suite hit unrelated 5s timeout failures in `src/main/services/archive/service.integration.test.ts` under full-suite load. That same file passes when run directly with a larger timeout, as listed above.

## Publish to changelog?

do not publish to changelog

### DIFF
--- a/apps/code/src/renderer/features/code-review/components/ReviewShell.test.tsx
+++ b/apps/code/src/renderer/features/code-review/components/ReviewShell.test.tsx
@@ -1,11 +1,19 @@
-import { render } from "@testing-library/react";
+import { render, renderHook } from "@testing-library/react";
+import { useEffect, useState } from "react";
 import { describe, expect, it, vi } from "vitest";
 
 vi.mock("@renderer/features/code-review/stores/reviewNavigationStore", () => ({
   useReviewNavigationStore: vi.fn(),
 }));
 vi.mock("@features/code-editor/stores/diffViewerStore", () => ({
-  useDiffViewerStore: vi.fn(),
+  useDiffViewerStore: vi.fn((selector) =>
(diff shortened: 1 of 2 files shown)


jonathanlab:
Hey @fredluz, thanks a lot, I think this PR got opened a few minutes after I merged in #2113 , which should help a lot with performance and rerender issues. Let me know if you still run into this issue, we can reopen in that case :D

PR #2202, opened 2026-05-18
fix(code-review): refresh staged/unstaged diffs on index changes

adboio:
## Summary

- External git operations (terminal `git add`, commits, etc.) write `.git/index` but the file watcher's `onGitStateChanged` handler only invalidated branch queries, so cached `getDiffCached` and `getDiffUnstaged` results stayed stale. The Local Review viewer kept showing the previous staged version after staging through the terminal.
- Dropped the `hasStagedFiles` gate on `getDiffCached` so the staged section appears/disappears in lockstep with the query result rather than racing the changed-files query (which is itself gated by `isActive` since #2113).

## Repro before fix

1. Edit `fileX`, `git add fileX`, then edit `fileX` again so it has both staged and unstaged changes.
2. Open the Review tab — only the staged version shows; the unstaged entry is missing or stale.
3. Run `git add fileX` from the terminal — viewer keeps showing the prior split instead of collapsing into a single combined staged diff.

## Test plan

- [ ] `pnpm dev`, open a local repo task
- [ ] Make staged + unstaged changes to the same file via terminal — both sections render with that file
- [ ] Run `git add <file>` in the terminal while the Review tab is open — the "Changes" section drops the file and "Staged Changes" updates to include the combined diff
- [ ] In-app stage/unstage button (`+` in Changes panel) still works
- [ ] `pnpm --filter code typecheck` passes
- [ ] `pnpm --filter code test` passes

---

### DIFF
--- a/apps/code/src/renderer/features/code-review/hooks/useReviewDiffs.ts
+++ b/apps/code/src/renderer/features/code-review/hooks/useReviewDiffs.ts
@@ -31,7 +31,7 @@ export function useReviewDiffs(
     trpc.git.getDiffCached.queryOptions(
       { directoryPath: repoPath as string, ignoreWhitespace: hideWhitespace },
       {
-        enabled: isActive && !!repoPath && hasStagedFiles,
+        enabled: isActive && !!repoPath,
         staleTime: 30_000,
         gcTime: 0,
         refetchOnMount: "always",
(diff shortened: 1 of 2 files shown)


jonathanlab (approved):

jonathanlab (approved):

adboio:
### Merge activity

* **May 19, 2:51 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/2202).

PR #2203, opened 2026-05-18
feat(sidebar): multi-select tasks with bulk archive

fercgomes:
## Summary

- Cmd/Ctrl-click toggles a task in the selection; shift-click extends a range from the last clicked anchor.
- Right-clicking a task while 2+ are selected opens a bulk context menu with "Archive N tasks" (confirm dialog). Right-clicking outside the selection clears it and falls through to the single-task menu.
- Escape clears the selection (ignored when focus is in an input/textarea/contenteditable).
- Selection lives in a small Zustand store (`taskSelectionStore`) modeled after the inbox pattern, with a prune effect that drops ids no longer visible when filters change.
- `archiveTaskImperative` gets a sibling `archiveTasksImperative` helper used by both the new bulk path and the existing "Archive prior tasks" action.

## Test plan

- [ ] Plain click navigates and clears any selection
- [ ] Cmd/Ctrl-click toggles selection without navigating
- [ ] Shift-click selects a contiguous range across pinned + chronological tasks
- [ ] Right-click on a selected task with 2+ selected opens the bulk menu and "Archive N tasks" archives all of them
- [ ] Right-click on an unselected task with a selection clears the selection and shows the normal single-task menu
- [ ] Escape clears the selection; typing in an input does not
- [ ] Hover toolbar (pin/archive icons) is hidden while a multi-selection is active
- [ ] Changing filters (e.g. organize/sort/visibility) prunes any selected ids that are no longer visible
- [ ] `pnpm --filter code typecheck`, biome, and vitest (43 tests in sidebar/tasks) all pass

### DIFF
--- a/apps/code/src/main/services/context-menu/schemas.ts
+++ b/apps/code/src/main/services/context-menu/schemas.ts
@@ -11,7 +11,7 @@ export const taskContextMenuInput = z.object({
 });
 
 export const bulkTaskContextMenuInput = z.object({
-  taskCount: z.number().int().positive(),
+  taskCount: z.number().int().min(2),
 });
 
 export const archivedTaskContextMenuInput = z.object({
--- a/apps/code/src/main/services/context-menu/schemas.ts
+++ b/apps/code/src/main/services/context-menu/schemas.ts
@@ -10,6 +10,10 @@ export const taskContextMenuInput = z.object({
   hasEmptyCommandCenterCell: z.boolean().optional(),
 });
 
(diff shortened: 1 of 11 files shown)


adboio:
thank you, this is great!

fixed merge conflict + greptile comments + a bug (that might have been a result of my conflict "fix" lol)

approved, feel free to reverse-review if you want and merge!

adboio (approved):

PR #2205, opened 2026-05-18
fix(sessions): strip remote workspace prefix from file paths in chat

timgl:
## Summary

When viewing a cloud task, file chips in the chat rendered the full sandbox path (e.g. `/tmp/workspace/repos/posthog/posthog.com/src/c`). For cloud tasks the local workspace cwd is empty (`folderPath: ""`, `worktreePath: null`), so `useCwd()` returned `undefined` and `FileMentionChip`'s `toRelativePath` had nothing to strip.

This adds a small `useDisplayRepoPath(taskId)` hook that prefers the local cwd, and falls back to the conventional sandbox root `/tmp/workspace/repos/<owner>/<repo>` derived from `task.repository` when there isn't one. `FileMentionChip` uses it instead of `useCwd` so the chip shows just the repo-relative directory on cloud tasks.

- `Read`/`Edit` views are unchanged structurally — they already use `FileMentionChip`; the chip just gets a correct prefix to strip now.
- Local tasks keep the existing behavior (cwd takes precedence).
- The `ToolCallView` fallback (`Glob`/`Grep`/etc. via `compactHomePath(title)`) is unchanged — out of scope for this fix.

## Test plan

- [x] `pnpm vitest run src/renderer/features/sessions` — 161 tests pass, including 7 new tests for `useDisplayRepoPath` covering local cwd, derived remote path, missing/malformed repository, and missing task.
- [x] `pnpm exec tsc -p tsconfig.web.json --noEmit` — clean.
- [x] `biome check` — clean (auto-formatted on commit via lint-staged).
- [ ] Manually: open a cloud task that has performed a `Read`, verify the chip shows `src/...` rather than `/tmp/workspace/repos/<owner>/<repo>/src/...`.
- [ ] Manually: open a local task and verify nothing regresses.

---

### DIFF
--- a/apps/code/src/renderer/features/sessions/hooks/useDisplayRepoPath.ts
+++ b/apps/code/src/renderer/features/sessions/hooks/useDisplayRepoPath.ts
@@ -1,0 +1,21 @@
+import { useCwd } from "@features/sidebar/hooks/useCwd";
+import { useTasks } from "@features/tasks/hooks/useTasks";
+import { parseRepository } from "@utils/repository";
+
+const REMOTE_WORKSPACE_PREFIX = "/tmp/workspace/repos";
+
+export function useDisplayRepoPath(
+  taskId: string | undefined,
+): string | undefined {
+  const localCwd = useCwd(taskId ?? "");
+  const { data: tasks = [] } = useTasks();
+
+  if (localCwd) return localCwd;
+  if (!taskId) return undefined;
+
+  const task = tasks.find((t) => t.id === taskId);
+  const parsed = task?.repository ? parseRepository(task.repository) : null;
+  if (!parsed) return undefined;
+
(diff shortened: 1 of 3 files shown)


apps/code/src/renderer/features/sessions/hooks/useDisplayRepoPath.ts
```diff
@@ -0,0 +1,21 @@
+import { useCwd } from "@features/sidebar/hooks/useCwd";
+import { useTasks } from "@features/tasks/hooks/useTasks";
+import { parseRepository } from "@utils/repository";
+
+const REMOTE_WORKSPACE_PREFIX = "/tmp/workspace/repos";
+
+export function useDisplayRepoPath(
```
timgl, line 13: I think we could use a comment here to explain what/why (brief!)

PR #2206, opened 2026-05-18
feat(task-input): cache cloud repositories so picker renders instantly

timgl:
## Summary

Clicking **New task** next to a repo in the sidebar would show a "Loading repos..." spinner in the GitHub repo picker, even though the answer was almost always the same repo the user picked last time. The picker waited on a full integrations + per-installation repo fetch from PostHog every mount, and only then could it validate the selection against the live list.

This change persists the user repository map in `settingsStore` and uses it as a stand-in while the live data loads, so the picker renders the last-used repo immediately. The network request still fires in the background and replaces the cache when it returns; the existing cleanup effect will still clear the selection if the repo turns out to be gone.

### Changes

- `settingsStore`: new persisted field `cachedCloudRepositoryMap` (repo name → `{ userIntegrationId, installationId }`) + setter.
- `useUserRepositoryIntegration`:
  - Writes fresh repo map to cache once everything has loaded.
  - Clears the cache when the user has no integrations.
  - Substitutes the cached map for `repositoryMap` while live queries are pending and `repositoryMap` is still empty.
  - Treats `hasGithubIntegration` as true while integrations are loading if cache is non-empty, so the picker doesn't briefly collapse to local mode on cold start.

No changes to `TaskInput` are required — the existing `selectedCloudRepository` validation and the cleanup effect both already work against whatever the hook returns.

## Test plan

- [ ] Unit: settingsStore persists and rehydrates cachedCloudRepositoryMap (added tests).
- [ ] Cold start with a cached repo: clicking **New task** in the sidebar shows the previously selected repo immediately, no "Loading repos..." spinner.
- [ ] First-ever launch (empty cache): behavior is unchanged — shows the loading state until repos load.
- [ ] User removes a GitHub integration that contains the cached repo: after live data loads, the selection clears (existing cleanup effect).
- [ ] User has zero GitHub integrations: cache gets cleared, picker shows "No GitHub repos".
- [ ] Refreshing repos via the picker's refresh button still works; cache gets updated to the fresh data.

---

### DIFF
--- a/apps/code/src/renderer/hooks/useIntegrations.ts
+++ b/apps/code/src/renderer/hooks/useIntegrations.ts
@@ -5,6 +5,7 @@ import {
   useIntegrationSelectors,
   useIntegrationStore,
 } from "@features/integrations/stores/integrationStore";
+import { useSettingsStore } from "@features/settings/stores/settingsStore";
 import type { UserGitHubIntegration } from "@renderer/api/posthogClient";
 import { useQueries, useQueryClient } from "@tanstack/react-query";
 import {
@@ -503,6 +504,13 @@ export function useUserRepositoryIntegration() {
     useUserGithubIntegrations();
   const [isRefreshingRepos, setIsRefreshingRepos] = useState(false);
 
+  const cachedCloudRepositoryMap = useSettingsStore(
+    (s) => s.cachedCloudRepositoryMap,
+  );
+  const setCachedCloudRepositoryMap = useSettingsStore(
+    (s) => s.setCachedCloudRepositoryMap,
+  );
+
   const {
     repositoryMap,
     reposByInstallationId,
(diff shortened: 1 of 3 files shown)


apps/code/src/renderer/hooks/useIntegrations.ts
```diff
@@ -503,32 +504,82 @@ export function useUserRepositoryIntegration() {
...
+  useEffect(() => {
+    if (integrationsPending) return;
+    if (githubIntegrations.length === 0) {
+      if (Object.keys(cachedCloudRepositoryMap).length > 0) {
+        setCachedCloudRepositoryMap({});
+      }
+      return;
+    }
+    if (reposPending) return;
+    if (Object.keys(repositoryMap).length === 0) return;
```
tatoalo, line 534: here we can also add 
```python
if (reposPending) [..]
if (failedInstallationIds.length > 0) return;
if (Object.kyes(..)) [..]
```
so we guard against persisting a partial map

tatoalo (approved):
approved, just one nit regarding an additional guard to the `setCachedCloudRepositoryMap` path

charlesvien:
Carrying this work into https://github.com/PostHog/code/pull/2769

PR #2207, opened 2026-05-19
feat(signals): view agent-proposed fix as a diff

Twixes:
## Problem

When the signals research agent flags a report as actionable, we want to see _its_ proposed fix as a diff inside PostHog Code — without spinning the sandbox back up.

## Changes

- Fetch the `PROPOSED_FIX_PATCH` artefact for a signal report
- Unpack the JSON envelope `{base_sha, base_ref, stats, patch}` and render `patch` in the built-in diff viewer
- Stats banner (files / +lines / -lines) above the diff
- Drift indicator when `base_sha` no longer matches current `origin/main`
- Graceful empty state when the artefact isn't there yet (not actionable, or still running)

(Fill me in once it lands.)

## How did you test this?

I'm an agent — nothing tested yet.

## Publish to changelog?

no

---
*Pairs with https://github.com/PostHog/posthog PR for the producer side.*


charlesvien:
closing this PR because there is no diff

PR #2210, opened 2026-05-19
feat(notifications): notify when focused on a different task

richardsolomou:
## Problem

When you're focused on task A inside PostHog Code and task B asks for a permission or finishes its turn, nothing surfaces — the existing notification logic in `notifications.ts` only fires when the window is unfocused, so a same-window-but-different-task interrupt is silent. Reported as: "no meep when task need input while focus other task, sad".

## Changes

Added a `shouldNotifyForTask(taskId)` helper in `apps/code/src/renderer/utils/notifications.ts` that consults `useNavigationStore` for the currently viewed task. Both `notifyPromptComplete` and `notifyPermissionRequest` now fire (sound, desktop notification, dock badge, dock bounce) when:

- the window is unfocused (unchanged), **or**
- the window is focused but the user is viewing a different task (new).

| Window focused? | Viewing same task? | Notify? |
|---|---|---|
| No | — | Yes (unchanged) |
| Yes | Yes | No (unchanged) |
| Yes | No / different task | **Yes (new)** |

All existing settings (`desktopNotifications`, `dockBadgeNotifications`, `dockBounceNotifications`, `completionSound`, `completionVolume`) still gate the channels — no new preference. Falls back to "don't notify when focused" if `taskId` is unknown, so callers without a `taskId` keep their previous behaviour.

Clicking the desktop notification already routes to the right task via `TaskLinkService` → `deepLink.onOpenTask` → `useTaskDeepLink` → `navigateToTask`, so no extra wiring needed there.

## How did you test this?

- `pnpm --filter code typecheck` — clean
- `pnpm exec biome check apps/code/src/renderer/utils/notifications.ts` — clean
- Traced the click-to-navigate path end-to-end (`electron-notifier.ts:25` → `notification/service.ts:38` → `deep-link.ts:27` → `useTaskDeepLink.ts` → `navigationStore.navigateToTask`) to confirm clicking the notification still lands on the originating task.
- Did not run the app manually in this environment.

## Publish to changelog?

no

### DIFF
--- a/apps/code/src/renderer/utils/notifications.test.ts
+++ b/apps/code/src/renderer/utils/notifications.test.ts
@@ -40,11 +40,9 @@ import { notifyPermissionRequest, notifyPromptComplete } from "./notifications";
 const TASK_ID = "task-123";
 const OTHER_TASK_ID = "task-999";
 
-function setView(view: {
-  type: string;
-  data?: { id: string };
-  taskId?: string;
-}) {
+type View = { type: string; data?: { id: string }; taskId?: string };
+
+function setView(view: View) {
   useNavigationStore.setState({
     // biome-ignore lint/suspicious/noExplicitAny: test-only narrow cast
     view: view as any,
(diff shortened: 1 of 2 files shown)


apps/code/src/renderer/utils/notifications.test.ts
jonathanlab, line 1: lets parametrize if possible

jonathanlab (approved):
Nice!

PR #2213, opened 2026-05-19
feat(cloud): instruct agent to create signed commits via GraphQL API

Piccirello:
## Problem

Cloud agent commits land unsigned because the sandbox runs `git commit` with no signing key. This prevents us from enforcing signed commits in our repos.

## Changes

Switch the cloud system prompt to instruct the agent to use GitHub's `createCommitOnBranch` mutation instead, which signs commits with the API token identity automatically.

## How did you test this?

Untested tbqh

## Publish to changelog?

Yes!

### DIFF
--- a/packages/agent/src/server/agent-server.ts
+++ b/packages/agent/src/server/agent-server.ts
@@ -1589,24 +1589,80 @@ export class AgentServer {
   private buildCloudSystemPrompt(prUrl?: string | null): string {
     const taskId = this.config.taskId;
     const shouldAutoCreatePr = this.shouldAutoPublishCloudChanges();
-    const attributionInstructions = `
-## Attribution
-Do NOT use Claude Code's default attribution (no "Co-Authored-By" trailers, no "Generated with [Claude Code]" lines).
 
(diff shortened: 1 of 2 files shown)


packages/agent/src/server/agent-server.ts
```diff
@@ -1589,24 +1589,80 @@ export class AgentServer {
...
+
+1. Stage the files for this commit with \`git add <files>\` (do NOT run \`git commit\`).
+2. Determine the branch's current tip SHA — for a brand-new branch this is the
+   base branch tip; for an existing branch, the latest commit on that branch.
+3. Build a JSON payload describing the file additions and deletions in this commit:
+   \`\`\`bash
+   # Diff staged tree against the current branch tip.
+   # Outputs lines like:  M\\tpath/to/file   or   D\\tpath/to/file
+   git diff --cached --name-status "$BRANCH_TIP_SHA" |
+     python3 -c '
```
Piccirello: Confirm: is python3 available in the sandbox?

Piccirello: Switched to `node` which is available.

PR #2215, opened 2026-05-19
fix(integrations): Fix "Update in GitHub" flow

Twixes:
## What's broken

If you already have GitHub connected and hit "Update in GitHub" in Code settings, it wasn't doing the right thing. It kicked off a fresh connect flow instead of taking you to your existing app installation. And when you came back after changing repo access on GitHub, Code didn't reliably pick up the update (esp. the `setup_action=update` path where GitHub doesn't send `state`).

## What this fixes

"Update in GitHub" now opens the correct GitHub installation settings page (org installs vs personal installs are different URLs), and sets up the callback so PostHog knows to finish setup when you return. You land back in Code with refreshed integration state. Initial "Connect GitHub" is unchanged..

Needs [PostHog/posthog#58951](https://github.com/PostHog/posthog/pull/58951) on the backend for the return trip to actually work end-to-end.

### DIFF
--- a/apps/code/src/renderer/api/posthogClient.ts
+++ b/apps/code/src/renderer/api/posthogClient.ts
@@ -665,7 +665,7 @@ export class PostHogAPIClient {
         body: JSON.stringify({ next }),
       },
     });
-    if (!response.ok && response.status !== 204) {
+    if (!response.ok) {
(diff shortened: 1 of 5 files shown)


charlesvien:
I'm getting reports of this from external users as well

PR #2216, opened 2026-05-19
fix: keep inline tool call open while selecting text

thmsobrmlr:
TL;DR: I want to be able to select text from AskUserQuestion, but can't because it toggles when i release the mouse button

## Summary

- Drag-selecting text inside an expanded `ToolCallView` or `ExecuteToolView` released on the outer `<Box>`, which fired the toggle's `onClick` and collapsed the card before the user could copy.
- Move the `onClick`, `cursor-pointer`, and `group` classes onto the header `<Flex>` so the expanded content is a sibling, not a child — matches the pattern already used by `ReadToolView`/`FetchToolView`/`SearchToolView`.
- Cursor is now an I-beam over selectable output (instead of a pointer), and selection just works without any `window.getSelection()` guard.
- Only these two views were affected; the other expandable tool views already used this layout, and `Edit`/`Think`/`Subagent` use dedicated `<button>` / `IconButton` for the toggle.

## Test plan

- [x] `pnpm --filter code test` (1244 tests pass)
- [x] Manual: trigger an `ExecuteToolView` (any Bash) or `ToolCallView` (unrecognized agent tool, e.g. AskUserQuestion), expand it, drag-select text in the body — card stays open, cursor shows I-beam.
- [x] Manual: click on the header (no selection) — card still toggles as before, cursor shows pointer.

---

### DIFF
--- a/apps/code/src/renderer/features/sessions/components/session-update/ExecuteToolView.tsx
+++ b/apps/code/src/renderer/features/sessions/components/session-update/ExecuteToolView.tsx
@@ -45,18 +45,18 @@ export function ExecuteToolView({
   const isExpandable = hasOutput;
 
   const handleClick = () => {
-    if (!isExpandable) return;
-    const selection = window.getSelection();
-    if (selection && selection.toString().length > 0) return;
-    setIsExpanded(!isExpanded);
+    if (isExpandable) {
+      setIsExpanded(!isExpanded);
+    }
   };
 
   return (
-    <Box
-      className={`group py-0.5 ${isExpandable ? "cursor-pointer" : ""}`}
-      onClick={handleClick}
-    >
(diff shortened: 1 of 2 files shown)


adboio:
thanks!! this is a good improvement, but i made a small change - i like to select text by triple clicking and that doesn't work here, + (to me) the cursor staying as `pointer` didn't make it clear that text was selectable

what do you think about this? updated so the click target is only the top of the card, and the expanded stuff is treated normally

https://github.com/user-attachments/assets/0a9fcdd6-9b93-449e-bf20-387d467b96be

adboio (approved):

adboio:
@thmsobrmlr feel free to merge if you like this 🫡

Issue #2159, opened 2026-05-17
Cloud job appears idle but queued messages never send


charlesvien:
After leaving a cloud job and returning to it later, the agent is not doing any work but messages I sent appear stuck in a "Queued" state. The job seems idle yet the queue never drains. This has happened a couple of times.

Thread: https://posthog.slack.com/archives/C09G8Q32R6F/p1778588082344929

<img width="1672" height="762" alt="Image" src="https://github.com/user-attachments/assets/28ca1cce-6fc5-401a-9e1c-f3d9845475a4" />

charlesvien:
https://posthog.slack.com/archives/C09G8Q32R6F/p1778741609459069

this has happened to me a couple times today. A task just starts queuing prompts but not ever working on them. I end up just archiving and starting over.

<img width="2284" height="1342" alt="Image" src="https://github.com/user-attachments/assets/57262b78-87cc-4d79-bb47-e8462b1fda0c" />

PR #2217, opened 2026-05-19
fix(code): drain cloud queue for idle resumed runs stuck disconnected

tatoalo:
## Problem

when a cloud task is resumed from a snapshot and then goes idle, an SSE transport drop (or the watcher retry it triggers) flips the session to `disconnected` even though the run is still alive (`in_progress`) on the server. A user message sent in that state is queued because `status !== "connected"`, and then nothing ever drains it

image

so basically the message stays queued forever

## Changes

- track a run-id-scoped `agentReadyForRunId` on the session, set when `_posthog/run_started` / `_posthog/turn_complete` is observed for the current run (survives hydrate-from-logs)
- boot race stays protected so that a still-booting run has no `run_started` for its run id, so recovery does not fire
- add a guarded `tryRecoverIdleCloudQueue` recovery path

handling https://github.com/PostHog/code/issues/2159

### DIFF
--- a/apps/code/src/renderer/features/sessions/service/service.recovery.integration.test.ts
+++ b/apps/code/src/renderer/features/sessions/service/service.recovery.integration.test.ts
@@ -1,0 +1,925 @@
+/**
+ * End-to-end recovery test against the REAL Zustand session store.
+ *
+ * This test exercises the actual store so the
+ * `updateSession -> getSessionByTaskId -> drain` chain is real — the precise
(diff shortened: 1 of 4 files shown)


joshsny (approved):
i think some of this logic might be getting unweildy and we might need to come back and simplify it afterwards, but sounds like this will improve things at least for now, great job finding the issue

tatoalo:
@joshsny agree completely, it's becoming a bit of a 🍝 , we need some time to have proper ACP cleanup of all this

PR #2218, opened 2026-05-19
fix(git): cap and stream untracked-file line counting

pauldambra:
Fixes an OOM in the main process that freezes the app a few seconds after the sidebar paints, on accounts whose selected directory is a monorepo with large untracked files (lockfiles, build artifacts, source maps). `countFileLines` was reading each file's full contents into V8 memory via `fs.readFile`, run 16-way concurrent against tens of thousands of paths — a single big file in any of the 16 in-flight reads could push the heap past 4 GB and crash main. Now we stream-count `\n` bytes with `createReadStream`, so peak per-stream memory is ~64 KB regardless of file size and the 16-way batch stays bounded around 1 MB total.

## Problem

- `countFileLines` in `packages/git/src/queries.ts` was reading each untracked file fully into memory via `fs.readFile(filePath, "utf-8")`, and the caller (`getChangedFilesDetailed`) runs it 16-way concurrent against up to 50k untracked paths from `streamGitStatus`.
- On a monorepo with multi-MB build artifacts, lockfiles, or other large untracked files, peak heap was `16 × file_bytes × 2` (V8's UTF-16 cost). That blew past Electron's 4 GB main-process heap on real-world repos and OOM'd the main process. The renderer appeared frozen ~2–3 s after the sidebar painted because all its in-flight tRPC calls were waiting on a dead main. Symptom in chromium.log was a silent `OOM error in V8: Scavenger: semi-space copy Allocation failed`, with no `[ipc-rate]` warning before it (no IPC storm — just main quietly running out of heap).
- Regression introduced by #2113 (c617988f), which removed the previous 10k-file cap and switched to concurrent reads when migrating to `streamGitStatus`. Bisected on a freshly-onboarded data dir pointed at this monorepo.

## Fix

- Stream the file with `createReadStream` (64 KB highWaterMark) and count `\n` bytes byte-by-byte. Per-stream memory stays at ~64 KB regardless of file size, so peak across 16 concurrent reads is ~1 MB total — well under any heap pressure threshold.
- Preserves the original return semantics: trailing-newline-aware count, 0 for empty / non-regular files / errors.
- No artificial file-size cap — every untracked file gets an accurate line count. (An earlier version of this PR gated at 1 MB, which silently returned `linesAdded: 0` for any larger file. That's a correctness regression and the streaming alone is already enough to fix the OOM, so the cap was dropped in the follow-up commit.)

## Test plan

- [x] `pnpm --filter @posthog/git test src/queries.test.ts` — new `getChangedFilesDetailed` tests pass (small-file accuracy + correct counting of a multi-MB untracked file that previously OOM'd)
- [x] `pnpm dev` against a fresh data dir pointed at this monorepo — previously OOM'd within ~3 s of `MainSidebar` paint, now stays responsive
- [ ] CI

### DIFF
--- a/packages/git/src/queries.test.ts
+++ b/packages/git/src/queries.test.ts
@@ -244,7 +244,12 @@ describe("getBranchDiffPatchesByPath", () => {
   });
 });
 
-describe("getChangedFilesDetailed", () => {
+// Picked to land well past the default 64KB highWaterMark of
+// `createReadStream` so the regression test actually exercises the
+// across-chunk path of the streaming line counter.
+const LINE_COUNT_LARGER_THAN_READ_STREAM_CHUNK = 800_000;
+
+describe("getChangedFilesDetailed > untracked line counts", () => {
   let repoDir: string;
 
   afterEach(async () => {
@@ -254,17 +259,33 @@ describe("getChangedFilesDetailed", () => {
     }
   });
 
-  it("reports line counts for small untracked files", async () => {
+  it.each([
+    { name: "trailing newline", content: "a\nb\nc\n", expected: 3 },
+    { name: "no trailing newline", content: "a\nb\nc", expected: 3 },
+    { name: "single byte, no newline", content: "a", expected: 1 },
+    { name: "lone newline", content: "\n", expected: 1 },
+    { name: "consecutive newlines", content: "\n\n", expected: 2 },
+    // CRLF: legacy `split("\n")` counted only `\n` separators, so
+    // `"a\r\nb\r\n"` -> 2 lines. Byte-counter matches.
+    { name: "CRLF endings", content: "a\r\nb\r\n", expected: 2 },
+  ])("counts $name as $expected line(s)", async ({ content, expected }) => {
     repoDir = await setupRepo();
-    await writeFile(path.join(repoDir, "small.txt"), "a\nb\nc\n");
-    await writeFile(path.join(repoDir, "no-trailing.txt"), "a\nb\nc");
+    await writeFile(path.join(repoDir, "f.txt"), content);
 
     const files = await getChangedFilesDetailed(repoDir);
-    const small = files.find((f) => f.path === "small.txt");
-    const noTrailing = files.find((f) => f.path === "no-trailing.txt");
+    const f = files.find((file) => file.path === "f.txt");
 
-    expect(small).toMatchObject({ status: "untracked", linesAdded: 3 });
-    expect(noTrailing).toMatchObject({ status: "untracked", linesAdded: 3 });
+    expect(f).toMatchObject({ status: "untracked", linesAdded: expected });
+  });
+
+  it("reports 0 lines for empty untracked files", async () => {
+    repoDir = await setupRepo();
+    await writeFile(path.join(repoDir, "empty.txt"), "");
+
+    const files = await getChangedFilesDetailed(repoDir);
+    const empty = files.find((f) => f.path === "empty.txt");
+
+    expect(empty).toMatchObject({ status: "untracked", linesAdded: 0 });
   });
 
   // Regression guard for the OOM in #2218. Before the fix `countFileLines`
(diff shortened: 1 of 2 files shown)


packages/git/src/queries.ts
```diff
@@ -415,9 +416,28 @@ function matchesExcludePattern(filePath: string, patterns: string[]): boolean {
 
 async function countFileLines(filePath: string): Promise<number> {
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + paul + xp]** 🟡 MEDIUM

The new streaming `countFileLines` doesn't accept or honour an `AbortSignal`, and `mapWithConcurrency` doesn't check one between files either. The rest of `getChangedFilesDetailed` already threads `gitOptions?.abortSignal` through to `streamGitStatus` and `executeRead`, so this is the one place that drops the signal. With up to 50 k untracked files at concurrency 16 — the exact workload this PR is hardening against — a cancelled tRPC call still streams every file to completion before the result is discarded.

Fix: thread `signal` down into `countFileLines` and wire it into `createReadStream(filePath, { signal })` (Node supports that natively). Same change naturally covers the "file deleted mid-read" case.

packages/git/src/queries.ts
```diff
@@ -415,9 +416,28 @@ function matchesExcludePattern(filePath: string, patterns: string[]): boolean {
...
+        if (buf.length > 0) lastByte = buf[buf.length - 1];
+      });
+      stream.on("end", () => {
+        if (lastByte === -1) {
+          resolve(0);
+          return;
+        }
+        resolve(lastByte === 0x0a ? newlines : newlines + 1);
+      });
+      stream.on("error", () => resolve(0));
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + paul + xp]** 🟡 MEDIUM

`stream.on("error", () => resolve(0))` collapses every read failure (permission denied, mid-read truncation, disk error) into "empty file". That's not a regression — the outer `try { … } catch { return 0; }` already did this — but the whole reason the size cap was dropped two commits ago is that "5 MB lockfile shows 0 lines" is a correctness regression. A swallowed stream error gives the same wrong answer without telling anyone.

Given this PR exists because a real monorepo blew up and nobody knew why until someone bisected it, log `path + error` here (via the package's logger). Silent zeros in a metrics-shaped field are exactly the smell that hid the original bug.

pauldambra (commented):
> [!NOTE]
>
> Multi-perspective review by qa-team (reliability / performance / data-integrity / compatibility / two generalists), paul-reviewer, and xp-reviewer. All three reviewers converged on the same two MEDIUM findings, posted inline. Six lower-severity findings (Buffer.indexOf perf, dead string-chunk branch, symlink-follow risk, extra fs.stat, lastByte dead branch, edge-case test coverage, magic-number extraction) were noted but are being addressed silently in the next commit.

Verdict: 💬 **APPROVE WITH NITS** — no blocking findings.

packages/git/src/queries.ts
pauldambra: > [!NOTE]

Fixed in `3475526b` — added a `signal?: AbortSignal` parameter to `countFileLines`, wired it into `createReadStream(filePath, { signal })`, and made `mapWithConcurrency` check the signal between items. `getChangedFilesDetailed` now passes `gitOptions?.abortSignal` through to both.

packages/git/src/queries.ts
pauldambra: > [!NOTE]

Fixed in `3475526b` — `stream.on("error", ...)` now logs `path + error` via `console.warn` (matches the existing pattern in `utils.ts:122`). Still resolves to 0 so callers don't have to handle a new failure shape, but the breadcrumb is there next time a "shows 0 lines" mystery shows up.

packages/git/src/queries.test.ts
```diff
@@ -242,3 +243,46 @@ describe("getBranchDiffPatchesByPath", () => {
...
+    await writeFile(path.join(repoDir, "small.txt"), "a\nb\nc\n");
+    await writeFile(path.join(repoDir, "no-trailing.txt"), "a\nb\nc");
+
+    const files = await getChangedFilesDetailed(repoDir);
+    const small = files.find((f) => f.path === "small.txt");
+    const noTrailing = files.find((f) => f.path === "no-trailing.txt");
+
+    expect(small).toMatchObject({ status: "untracked", linesAdded: 3 });
+    expect(noTrailing).toMatchObject({ status: "untracked", linesAdded: 3 });
+  });
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `3475526b` — the two `it` cases are now an `it.each` table covering trailing newline, no trailing newline, single byte, lone newline, consecutive newlines, and CRLF endings, plus an explicit empty-file case alongside.

jonathanlab (approved):

PR #2225, opened 2026-05-19
feat(agent): route slack-originated tasks to slack_app gateway product

VojtechBartos:
## Problem

The agent in the sandbox calls the LLM gateway as `posthog_code` regardless of how the task started. Slack-bot mentions are now billed differently from desktop tasks (gateway tags `slack_app` traffic with `\$ai_billable=true`), so the agent needs to send the correct product per task.

## Changes

- Add `"slack_app"` to the `GatewayProduct` type union (`packages/agent/src/utils/gateway.ts`).
- Add `"slack"` to the `Task.origin_product` union (`packages/agent/src/types.ts`) to mirror the Python `Task.OriginProduct.SLACK` value already set on the backend.
- `configureEnvironment` now accepts `originProduct?: Task["origin_product"] | null`. Mapping: `"slack"` → `slack_app`, `internal=true` → `background_agents`, else → `posthog_code`. Call site reads `preTask?.origin_product` from the Task fetched at session init.
- Three new unit tests covering the slack → slack_app mapping, slack-over-internal precedence, and non-slack fallback.

Pairs with PostHog/posthog#59040 (gateway-side product registration + per-turn quota gate). Both ship behind the existing `posthog-code-slack-availability` flag — neither side has user-visible effect until the flag is enabled.

## How did you test this?

I'm an agent. No manual testing.

- `pnpm --filter @posthog/agent test src/server/agent-server.configure-environment.test.ts` → **8 passed** (5 pre-existing + 3 new).
- Typecheck error count unchanged from clean main (pre-existing module-resolution noise unrelated to this change).
- `biome format --write` clean on touched files.

## Publish to changelog?

no

### DIFF
--- a/packages/agent/src/server/agent-server.configure-environment.test.ts
+++ b/packages/agent/src/server/agent-server.configure-environment.test.ts
@@ -1,10 +1,11 @@
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import type { Task } from "../types";
 import { AgentServer } from "./agent-server";
 
 interface TestableServer {
   configureEnvironment(args?: {
     isInternal?: boolean;
-    originProduct?: string | null;
+    originProduct?: Task["origin_product"] | null;
     signalReportId?: string | null;
     taskId?: string | null;
     taskRunId?: string | null;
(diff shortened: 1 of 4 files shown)


joshsny (approved):
LGTM

PR #2229, opened 2026-05-19
fix(code): show user skills in / popup and drop the autofill

richardsolomou:
## Problem

In running tasks, the `/` popup only showed `/good /bad /feedback` until the agent's `available_commands_update` arrived. And `/` mid-prompt opened an inline autocomplete instead of the popover.

## Changes

- Running tasks fall back to the trpc-fetched skills list while session commands are empty. Loader moved to `PromptInput` so session views populate it too.
- `/` opens the popover anywhere in the prompt (matching `@` and `#`), instead of the inline ghost-text autofill. The autofill code stays in tree but is no longer wired up.
- An explicit empty `available_commands_update` from the agent now suppresses the draft-store fallback, so the agent stays authoritative.

## How did you test this?

- `pnpm --filter code exec vitest run src/renderer/features/message-editor/` (70 tests pass, 6 in the new `getSuggestions.test.ts`)

## Publish to changelog?

no


adboio:
this is a regression, i'm gonna take a peek in case there's a smaller/simpler bug here we can fix!

adboio:
@richardsolomou clarifying - mid-prompt you should still be able to see commands, but they just "autofill" instead of showing the whole popover like you'd see at the start of the task. is that also not working for you?

richardsolomou:
@adboio that does work, but is that the intended way of doing it? It feels..off. why don't we just reuse the same approach? Similar to Slack mentions, you can @ someone at the beginning or the middle of the message and you have the same behavior

k11kirky:
Sorry @richardsolomou 

Still got one remaining this

pr_fetched_at is written but never read

updatePrCache still stamps prFetchedAt: now(), but a repo-wide grep confirms nothing reads it (only schema/migration/mock/write sites). getTaskPrStatus still fires void this.revalidateTaskPrStatus(...) unconditionally on every call:

const cachedPrState = (cached?.prState ?? null) as SidebarPrState;
void this.revalidateTaskPrStatus(taskId, cloudPrUrl);   // no staleness gate
if (cachedPrState) return { prState: cachedPrState, hasDiff: false };
So the headline cold-start scenario — every visible card spawning a gh subprocess at once on restart — is made non-blocking but not suppressed. The react-query staleTime: 60_000 only dedupes within a live session; it resets on restart, which is precisely when the column was meant to help. Still a "wire it in or drop it" call. A two-line gate closes it:

const TTL = 60_000;
const fetchedAt = cached?.prFetchedAt ? Date.parse(cached.prFetchedAt) : 0;
if (Date.now() - fetchedAt > TTL) void this.revalidateTaskPrStatus(taskId, cloudPrUrl);

we don't want the TTL, instead delete the column + its migration line so you're not shipping a written-but-unread field

richardsolomou:
> @adboio that does work, but is that the intended way of doing it? It feels..off. why don't we just reuse the same approach? Similar to Slack mentions, you can @ someone at the beginning or the middle of the message and you have the same behavior

@adboio what do you think about this?

adboio:
@richardsolomou sorry for the delay, i think you're right! testing now, will merge shortly :) thank you!

adboio:
* **#2229** <a href="https://app.graphite.com/github/pr/PostHog/code/2229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2229?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

adboio (approved):

adboio:
### Merge activity

* **Jun 8, 6:49 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/2229).

PR #2241, opened 2026-05-20
feat(code): add Discuss button to inbox reports

andrewm4894:
## Problem

Inbox reports have Dismiss and Create PR actions, but no lightweight way to talk through a report with the agent before deciding what to do. Manually opening a new task, referencing the report, choosing the right mode, and submitting is too much friction for “let me think about this with you.”

## Changes

Kick off discussion and optionally add a question or prompt:

image

Then a chat will open in auto mode and use the mcp to try answer/respond to your initial prompt:

image

- Adds a **Discuss** split button to the inbox report detail header.
- Primary click starts a chat for the current report immediately.
- Chevron opens a compact popover where you can optionally add the first question before starting the discussion.
- Discuss tasks auto-submit the initial prompt and start in **Auto Mode**, so the agent can use the inbox MCP tools without stopping on the initial permission prompt.
- The initial prompt includes the report ID plus a clickable app deeplink back to the inbox item.
- Markdown rendering now explicitly supports `posthog-code://` and `posthog-code-dev://` deeplinks.
- Auto-submit waits for task config/model readiness before creating the cloud run, and clears the transient Discuss draft after successful submission so New Task opens cleanly afterward.
- The report association banner remains attached to the created task, and repository selection reuses the same report/repo fallback path as Create PR.

## How did you test this?

- Focused Biome checks on touched files.
- Pre-commit hook ran Biome and full `pnpm typecheck`.
- Manual local smoke test in the Electron dev app:
  - default Discuss flow starts a chat automatically
  - optional first-question popover submits the custom question
  - Discuss starts in Auto Mode and uses inbox MCP tools
  - inbox deeplink renders as clickable in chat
  - returning to New Task no longer preserves the previous Discuss prompt

## Publish to changelog?

no

---

Small, frontend-focused workflow improvement — candidate for **stamphog** AI-assisted approval.

### DIFF
--- a/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
+++ b/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
@@ -4,6 +4,7 @@ import { HighlightedCode } from "@components/HighlightedCode";
 import { List, ListItem } from "@components/List";
 import { parseGithubIssueUrl } from "@features/message-editor/utils/githubIssueUrl";
 import { Blockquote, Checkbox, Code, Kbd, Text } from "@radix-ui/themes";
+import { isPostHogCodeDeeplink } from "@shared/deeplink";
 import { memo, useMemo } from "react";
 import type { Components } from "react-markdown";
 import ReactMarkdown, { defaultUrlTransform } from "react-markdown";
@@ -22,16 +23,6 @@ interface MarkdownRendererProps {
 // Ensures `---`, `***`, `___` are preceded by a blank line
 function preprocessMarkdown(content: string): string {
   return content.replace(/\n([^\n].*)\n(---+|___+|\*\*\*+)\n/g, "\n$1\n\n$2\n");
-}
-
-function isPostHogCodeDeeplink(href: string | undefined): href is string {
-  if (!href) return false;
-  try {
-    const protocol = new URL(href).protocol;
-    return protocol === "posthog-code:" || protocol === "posthog-code-dev:";
-  } catch {
-    return false;
-  }
 }
 
 function markdownUrlTransform(value: string): string {
--- a/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
+++ b/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
@@ -4,9 +4,10 @@ import { HighlightedCode } from "@components/HighlightedCode";
 import { List, ListItem } from "@components/List";
 import { parseGithubIssueUrl } from "@features/message-editor/utils/githubIssueUrl";
 import { Blockquote, Checkbox, Code, Kbd, Text } from "@radix-ui/themes";
+import { trpcClient } from "@renderer/trpc/client";
 import { memo, useMemo } from "react";
 import type { Components } from "react-markdown";
-import ReactMarkdown from "react-markdown";
+import ReactMarkdown, { defaultUrlTransform } from "react-markdown";
 import remarkGfm from "remark-gfm";
 import type { PluggableList } from "unified";
 import { GithubRefChip } from "./GithubRefChip";
@@ -22,6 +23,21 @@ interface MarkdownRendererProps {
 // Ensures `---`, `***`, `___` are preceded by a blank line
 function preprocessMarkdown(content: string): string {
   return content.replace(/\n([^\n].*)\n(---+|___+|\*\*\*+)\n/g, "\n$1\n\n$2\n");
+}
+
+function isPostHogCodeDeeplink(href: string | undefined): href is string {
+  if (!href) return false;
+  try {
+    const protocol = new URL(href).protocol;
+    return protocol === "posthog-code:" || protocol === "posthog-code-dev:";
+  } catch {
+    return false;
+  }
+}
+
+function markdownUrlTransform(value: string): string {
+  if (isPostHogCodeDeeplink(value)) return value;
+  return defaultUrlTransform(value);
 }
 
 const HeadingText = ({ children }: { children: React.ReactNode }) => (
@@ -92,9 +108,15 @@ export const baseComponents: Components = {
         </GithubRefChip>
       );
     }
+    const isDeeplink = isPostHogCodeDeeplink(href);
     return (
       <a
         href={href}
+        onClick={(event) => {
+          if (!isDeeplink || !href) return;
+          event.preventDefault();
+          void trpcClient.os.openExternal.mutate({ url: href });
+        }}
         target="_blank"
         rel="noopener noreferrer"
         className="markdown-link inline-flex items-center gap-[2px]"
@@ -189,6 +211,7 @@ export const MarkdownRenderer = memo(function MarkdownRenderer({
       remarkPlugins={plugins}
       rehypePlugins={rehypePlugins}
       components={components}
+      urlTransform={markdownUrlTransform}
     >
       {processedContent}
     </ReactMarkdown>
--- a/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
+++ b/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
@@ -4,7 +4,6 @@ import { HighlightedCode } from "@components/HighlightedCode";
 import { List, ListItem } from "@components/List";
 import { parseGithubIssueUrl } from "@features/message-editor/utils/githubIssueUrl";
 import { Blockquote, Checkbox, Code, Kbd, Text } from "@radix-ui/themes";
-import { trpcClient } from "@renderer/trpc/client";
 import { memo, useMemo } from "react";
 import type { Components } from "react-markdown";
 import ReactMarkdown, { defaultUrlTransform } from "react-markdown";
@@ -115,7 +114,11 @@ export const baseComponents: Components = {
         onClick={(event) => {
           if (!isDeeplink || !href) return;
           event.preventDefault();
-          void trpcClient.os.openExternal.mutate({ url: href });
+          // Lazy-load the tRPC client so tests that render markdown don't need
+          // an electronTRPC global at module init time.
+          void import("@renderer/trpc/client").then(({ trpcClient }) =>
+            trpcClient.os.openExternal.mutate({ url: href }),
+          );
         }}
         target="_blank"
         rel="noopener noreferrer"
--- a/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
+++ b/apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
@@ -4,6 +4,7 @@ import { HighlightedCode } from "@components/HighlightedCode";
 import { List, ListItem } from "@components/List";
 import { parseGithubIssueUrl } from "@features/message-editor/utils/githubIssueUrl";
 import { Blockquote, Checkbox, Code, Kbd, Text } from "@radix-ui/themes";
+import { trpcClient } from "@renderer/trpc/client";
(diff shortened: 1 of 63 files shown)


apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
```diff
@@ -480,6 +484,43 @@ export function TaskInput({
...
+  useEffect(() => {
+    if (!isAutoSubmitting || hasAutoSubmittedRef.current) return;
+    if (!canSubmit || !initialPrompt) return;
+    hasAutoSubmittedRef.current = true;
+    void handleSubmit({
+      segments: [{ type: "text", text: initialPrompt }],
+    }).then((ok) => {
+      if (!ok) setIsAutoSubmitting(false);
+    });
+  }, [isAutoSubmitting, canSubmit, handleSubmit, initialPrompt]);
```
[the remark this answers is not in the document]
andrewm4894: Fixed in 8abd14c1. I exported `canSubmitBase` from `useTaskCreation`, use it for the auto-submit effect, and allow `handleSubmit` to run with a `contentOverride` even while the editor is not mounted.

apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
```diff
@@ -480,6 +484,43 @@ export function TaskInput({
...
+    setIsAutoSubmitting(!!autoSubmit);
+  }, [prefillRequestKey, autoSubmit]);
+
+  // Fire the first message automatically once preconditions are satisfied.
+  // Uses `contentOverride` so we don't race the editor hydrating from the
+  // draft store — `handleSubmit` only requires `canSubmitBase` in that path.
+  useEffect(() => {
+    if (!isAutoSubmitting || hasAutoSubmittedRef.current) return;
+    if (!canSubmit || !initialPrompt) return;
+    hasAutoSubmittedRef.current = true;
```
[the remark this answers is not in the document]
andrewm4894: Fixed in 8abd14c1. Auto-submit now gates on `canSubmitBase` instead of the editor-ref-gated `canSubmit`, and `handleSubmit(contentOverride)` no longer requires `editorRef.current`, so the loading placeholder path can start the discussion automatically.

apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
```diff
@@ -24,6 +24,21 @@ function preprocessMarkdown(content: string): string {
...
+function isPostHogCodeDeeplink(href: string | undefined): href is string {
+  if (!href) return false;
+  try {
+    const protocol = new URL(href).protocol;
+    return protocol === "posthog-code:" || protocol === "posthog-code-dev:";
+  } catch {
+    return false;
+  }
+}
+
```
jonathanlab: This should not live here, maybe we already have something for this

apps/code/src/renderer/stores/navigationStore.ts
```diff
@@ -258,7 +266,6 @@ export const useNavigationStore = create<NavigationStore>()(
 
         navigateToInbox: () => {
           navigate({ type: "inbox" });
-          track(ANALYTICS_EVENTS.INBOX_VIEWED);
```
jonathanlab, line 261: Why are we removing this event?

apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
```diff
@@ -449,36 +459,102 @@ export function TaskInput({
...
+  // the user can submit manually once they pick a repo.
+  useEffect(() => {
+    if (!isAutoSubmitting) return;
+    const timer = setTimeout(() => {
+      if (!hasAutoSubmittedRef.current) {
+        setIsAutoSubmitting(false);
+      }
+    }, 2500);
+    return () => clearTimeout(timer);
+  }, [isAutoSubmitting]);
```
jonathanlab: This can easily take longer than 2.5s. I'm sure there's a better solution than relying on a timeout :/

apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
```diff
@@ -449,36 +459,102 @@ export function TaskInput({
...
```
jonathanlab: This whole thing is doing way too much IMO:
- three effects
- canSubmitBase as a workaround..?
- 2.5s timeout
- lots of new state

Maybe it's better if you just directly call `taskService.createTask` from the Discuss component instead

apps/code/src/shared/types/analytics.ts
```diff
@@ -501,6 +502,11 @@ export interface InboxReportActionProperties {
   signal_section?: "relevant_code" | "data_queried";
   why_field?: "priority" | "actionability";
   task_section?: "research" | "implementation";
+  // True when the user submitted Discuss with a first question via the popover.
+  has_question?: boolean;
+  // The first question text the user typed before hitting Discuss. Truncated to
+  // 500 chars to keep event payloads bounded.
+  question_text?: string;
```
jonathanlab, line 509: I don't think this should go inside the event right? And if we're truncating it, why even include it

apps/code/src/renderer/features/inbox/components/detail/ReportDetailPane.tsx
```diff
@@ -366,6 +372,42 @@ export function ReportDetailPane({
     fireDetailAction,
   ]);
 
+  const handleDiscussReport = useCallback(
+    (question?: string) => {
+      const trimmedQuestion = question?.trim();
+      const reportLink = `${getDeeplinkProtocol(import.meta.env.DEV)}://inbox/${report.id}`;
+      const prompt = trimmedQuestion
+        ? `Discuss PostHog inbox report ${report.id} ([inbox item](${reportLink})). Use the inbox MCP tools to fetch the report, then answer this first: ${trimmedQuestion}`
+        : `Discuss PostHog inbox report ${report.id} ([inbox item](${reportLink})). Use the inbox MCP tools to fetch the report, then give me a brief readout and ask what I want to dig into.`;
```
jonathanlab: Let's extract this to a builder

apps/code/src/shared/types/analytics.ts
andrewm4894: would like to keep this one - the motivation is mostly analytics ergonomics on a brand-new feature. Even truncated, having the first question text on the event makes it really easy to get a quick feel for the kinds of follow-ons people want without having to join against task/chat data each time. The full prompt is already going to the backend via the created task, so the truncated snippet on the event doesn't really add new PII exposure on top of that. Happy to revisit and drop it once Discuss is more mature and we have a clearer view of what's worth keeping long-term. 

(and when we have some sort of opt out layer too for things like this for users)

apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
andrewm4894: Good catch — there was already a `getDeeplinkProtocol` helper in `apps/code/src/shared/deeplink.ts`, so I moved `isPostHogCodeDeeplink` there alongside it and updated the protocol literals to use the existing constants. `MarkdownRenderer` now just imports it. Fixed in 75662a87.

apps/code/src/renderer/features/inbox/components/detail/ReportDetailPane.tsx
andrewm4894: Done — extracted to `apps/code/src/renderer/features/inbox/utils/buildDiscussReportPrompt.ts` with its own unit tests in 32fe704f. `ReportDetailPane` now just calls `buildDiscussReportPrompt({ reportId, question, isDevBuild })` so all the prompt copy lives in one place.

apps/code/src/renderer/stores/navigationStore.ts
andrewm4894: Not removed, just relocated — when I merged main into this branch, the work from #2228 (`feat(inbox): instrument analytics for engagement events`) had already redefined `INBOX_VIEWED` to require a rich `InboxViewedProperties` payload (report counts, filters, etc.). The bare `track(ANALYTICS_EVENTS.INBOX_VIEWED)` call here no longer typechecks, and the new structured fire happens inside `InboxSignalsTab.tsx` once the inbox actually renders with data — see [`InboxSignalsTab.tsx:478-506`](https://github.com/PostHog/code/blob/inbox-discuss-button/apps/code/src/renderer/features/inbox/components/InboxSignalsTab.tsx#L478-L506). So we still get the event, just with proper props and from a more accurate spot in the lifecycle.

apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
andrewm4894: Refactored as suggested in 9615f397. The Discuss button now goes straight through a small `useDiscussReport` hook that calls `taskService.createTask` directly — no TaskInput round-trip. That dissolves all of the things you flagged: the three effects, the `canSubmitBase` workaround, the 2.5s timeout, the loading placeholder, plus the matching transient state in `navigationStore` (`autoSubmit`, `initialExecutionMode`) and the editor-bypass changes in `useTaskCreation`. While the task is being created the Discuss button shows an in-place spinner and a `toast.loading` (same pattern as `connectivityToast.ts`); on success we jump to the task detail page, on failure a toast surfaces and we stay on the inbox. Net diff: ~125 lines removed.

apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
andrewm4894: Gone in 9615f397 — the new `useDiscussReport` hook calls `taskService.createTask` directly so there's nothing to time out against. While creation runs the user sees a `toast.loading` ("Starting discussion...") and the Discuss button shows an inline spinner; on success we navigate, on failure we surface a toast and stay on the inbox.

apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
andrewm4894: hopefully a bit simpler now - just uses a toast and flow all still feels same as a user - tested it locally

apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
```diff
@@ -92,9 +98,19 @@ export const baseComponents: Components = {
...
         href={href}
+        onClick={(event) => {
+          if (!isDeeplink || !href) return;
+          event.preventDefault();
+          // Lazy-load the tRPC client so tests that render markdown don't need
+          // an electronTRPC global at module init time.
+          void import("@renderer/trpc/client").then(({ trpcClient }) =>
+            trpcClient.os.openExternal.mutate({ url: href }),
+          );
+        }}
```
jonathanlab, line 110: Any way we can avoid this inline import pattern?

jonathanlab (approved):

apps/code/src/renderer/features/editor/components/MarkdownRenderer.tsx
andrewm4894: Done in cfe91072. Inline import is gone — back to a top-level `import { trpcClient }` like every other call site.

The reason it was lazy in the first place was that `UserMessage.test.tsx` was failing at module init because `trpc/client.ts` calls `ipcLink()` which needs `window.electronTRPC`, and `MarkdownRenderer` was the first thing in `UserMessage`'s import graph to reach it. Fixed properly by adding `vi.mock("@renderer/trpc/client", ...)` to the test — same pattern 14 other test files already use.

PR #2249, opened 2026-05-20
[codex] fix cloud queue idle recovery

tatoalo:
## Problem

repeated `in_progress` updates could re-enter idle recovery and hang the renderer 

## Changes

- coalesce deferred cloud queue flushes and skip idle recovery while a flush is already scheduled or dispatching


tatoalo:
Closing in favor of a clean branch rebased on latest main with only the cloud queue idle recovery commit.

PR #2277, opened 2026-05-21
fix(task-input): preserve picked folder when adding to recents

richardsolomou:
## Problem

When creating a new task and clicking **Open folder...** in local/worktree mode, the selected folder is added to the **Recent** list but isn't actually selected as the task's folder.

## Changes

The `useEffect` in `TaskInput.tsx` that syncs `selectedDirectory` from `view.folderId` had `folders` in its dependency array and re-ran every time `folders` changed. Picking a new folder via the file dialog triggers `addFolder`, which invalidates `getFolders`, which refetches and updates `folders` — causing the effect to fire and revert `selectedDirectory` back to the original `view.folderId` folder, overwriting the user's pick.

Track the last applied `folderId` in a ref so the effect only syncs when `view.folderId` actually changes, not on every folders refetch. The dependency on `folders` is preserved so the initial sync still works when the folder list hasn't loaded yet on mount.

## How did you test this?

- Typecheck (`pnpm --filter code typecheck`) passes.
- Build (`pnpm build`) succeeds.
- `pnpm --filter code test` — all TaskInput-related tests pass; the 2 failing tests are pre-existing archive service integration timeouts unrelated to this change.
- Did not manually verify in the running app — no display available in this environment.

## Publish to changelog?

no

---

### DIFF
--- a/apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
+++ b/apps/code/src/renderer/features/task-detail/components/TaskInput.tsx
@@ -48,6 +48,7 @@ import { useQuery } from "@tanstack/react-query";
 import { FOCUSABLE_SELECTOR } from "@utils/overlay";
 import { LayoutGroup, motion } from "framer-motion";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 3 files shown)


k11kirky (approved):
fine for now, will be refactored out of a hook at some point soon

PR #2297, opened 2026-05-22
feat: branching tasks

clr182:
## Problem

Tasks are fully isolated, there's no way to spin off a new task that continues from where another one is, with its context carried over.

## Changes

Adds a **Branch task** action to the task header. Branching summarises the current conversation (LLM call) and creates a new task seeded with that summary as its starting context. The new task starts from a clean tree carrying over code changes would also be nice but not in scope of this PR. A "Branched from #N" chip links a branched task back to its parent.

image
image

https://posthog.slack.com/archives/C09G8Q32R6F/p1779459522870039?thread_ts=1779447719.812489&cid=C09G8Q32R6F

## How did you test this?

ran local test
and added 2 new tests branchContext.test.ts`, `generateBranchSummary.test.ts`

## Publish to changelog?

Yes "Branch a task to start a new one that continues with its context."


adboio:
> Branching summarises the current conversation (LLM call) and creates a new task seeded with that summary as its starting context

is this the way other tools work? in my head, "branching" was more like just forking from the current point instead of summarizing, but maybe that's not right - and maybe those are just two things we could offer for differnet use-cases?

charlesvien:
I think we should implement forking instead

PR #2300, opened 2026-05-22
fix(cloud): signing mcp cloud only

tatoalo:
## Problem

local signing MCP tool must be only tied to cloud runs. it wasn't. now it is.


Piccirello (approved):
Very low context stamp

PR #2314, opened 2026-05-22
refactor(mobile): centralize safe-area spacing in useScreenInsets

Gilbert09:
## Why

On mobile, screens each called `useSafeAreaInsets()` and then hand-added their own magic offset for top/bottom spacing (`12 / 16 / 20 / 24 / 32 / 40 / 50`). The device insets are the correct cross-platform/cross-device primitive — and they're used — but the *policy layered on top* was duplicated and inconsistent per screen. (Originated from a question about the gap under the task-screen composer: `Math.max(insets.bottom, 50)`, correct for that screen but not shared anywhere.)

## What

- **`useScreenInsets` hook** — single source of truth for spacing policy. Wraps `useSafeAreaInsets()` and exposes a rationalized **12/24/40** scale via `bottom("compact" | "default" | "roomy")`, plus `composerBottom()`, `fabBottom()`, `contentTop()`, and raw `insets`.
- **`SheetContainer` component** — extracts the duplicated bottom-sheet shell (backdrop, pinned panel, rounded top, border, shadow, drag handle, inset-aware padding). `SelectSheet` and `AttachmentSheet` now render through it.
- **Migrated 17 call sites** — FABs, composers, page sheets, and the mcp/settings/pr-diff/automation-create screens.

## Rationalization (intentional)

Two values snap onto the scale; everything else maps exactly:
- `settings` bottom: **32 → 24**
- `pr-diff` bottom: **16 → 24**

## Left bespoke on purpose

Functional clearances that aren't safe-area policy: DismissReportSheet's `+120` (clears its sticky footer), automation/create's computed floating-button clearance, inbox/[id]'s `+100`/`+16`, and measured floating-header heights (`insets.top + 60/64`).

## Testing

- `biome check`: clean
- `tsc --noEmit`: no new errors (pre-existing test-file errors unchanged vs. `main`)
- ⚠️ Not visually verified on a simulator — the two 8px shifts (settings, pr-diff) are low-risk but worth an eyeball in light/dark.

### DIFF
--- a/apps/mobile/src/app/task/[id].tsx
+++ b/apps/mobile/src/app/task/[id].tsx
@@ -60,6 +60,10 @@ export default function TaskDetailScreen() {
   const router = useRouter();
   const queryClient = useQueryClient();
   const { insets, composerBottom } = useScreenInsets();
+  // Pre-compute outside the worklet: useAnimatedStyle runs on the UI thread and
+  // can't call the non-worklet getter. Capturing the primitive keeps the worklet
+  // closure stable (matches the pattern in task/index.tsx).
+  const composerBottomValue = composerBottom();
   const themeColors = useThemeColors();
   const [task, setTask] = useState<Task | null>(null);
   const [loading, setLoading] = useState(true);
@@ -123,9 +127,9 @@ export default function TaskDetailScreen() {
     // height, so the composer sits at the keyboard top — no extra gap needed
     // when open. Closed state keeps a comfortable bottom inset.
     return {
-      marginBottom: height.value < 0 ? 0 : composerBottom(),
+      marginBottom: height.value < 0 ? 0 : composerBottomValue,
     };
-  }, [composerBottom]);
+  }, [composerBottomValue]);
 
   useEffect(() => {
     if (!taskId) return;
--- a/apps/mobile/src/app/task/[id].tsx
+++ b/apps/mobile/src/app/task/[id].tsx
@@ -12,7 +12,6 @@ import {
 } from "react-native";
 import { useReanimatedKeyboardAnimation } from "react-native-keyboard-controller";
 import Animated, { useAnimatedStyle } from "react-native-reanimated";
-import { useSafeAreaInsets } from "react-native-safe-area-context";
 import { FloatingBackButton } from "@/components/FloatingBackButton";
 import { getTask, runTaskInCloud } from "@/features/tasks/api";
 import { FloatingTaskHeader } from "@/features/tasks/components/FloatingTaskHeader";
@@ -36,6 +35,7 @@ import { useTaskSessionStore } from "@/features/tasks/stores/taskSessionStore";
 import { useTaskStore } from "@/features/tasks/stores/taskStore";
 import type { Task } from "@/features/tasks/types";
 import { getSessionActivityPhase } from "@/features/tasks/utils/sessionActivity";
+import { useScreenInsets } from "@/hooks/useScreenInsets";
 import { logger } from "@/lib/logger";
 import { useThemeColors } from "@/lib/theme";
 
@@ -59,7 +59,7 @@ export default function TaskDetailScreen() {
   }>();
   const router = useRouter();
   const queryClient = useQueryClient();
-  const insets = useSafeAreaInsets();
+  const { insets, composerBottom } = useScreenInsets();
   const themeColors = useThemeColors();
   const [task, setTask] = useState<Task | null>(null);
   const [loading, setLoading] = useState(true);
@@ -123,9 +123,9 @@ export default function TaskDetailScreen() {
     // height, so the composer sits at the keyboard top — no extra gap needed
     // when open. Closed state keeps a comfortable bottom inset.
     return {
-      marginBottom: height.value < 0 ? 0 : Math.max(insets.bottom, 50),
+      marginBottom: height.value < 0 ? 0 : composerBottom(),
     };
-  }, [insets.bottom]);
+  }, [composerBottom]);
 
   useEffect(() => {
     if (!taskId) return;
--- a/apps/mobile/src/components/SheetContainer.tsx
+++ b/apps/mobile/src/components/SheetContainer.tsx
@@ -1,5 +1,5 @@
 import type { ReactNode } from "react";
-import { Modal, Pressable, View } from "react-native";
+import { Modal, Pressable, View, type ViewStyle } from "react-native";
 import {
   type BottomGapVariant,
   useScreenInsets,
@@ -17,10 +17,17 @@ interface SheetContainerProps {
   open: boolean;
   onClose: () => void;
   children: ReactNode;
-  /** Bottom padding gap above the safe-area inset. Defaults to "compact". */
+  /**
+   * Bottom padding gap above the safe-area inset. Defaults to "compact"
+   * because that is the gap both legacy sheets (SelectSheet, AttachmentSheet)
+   * inherited — deliberately not BOTTOM_GAP's own "default" (24). Pass
+   * "default"/"roomy" when a sheet wants more breathing room.
+   */
   bottomGap?: BottomGapVariant;
   /** Extra classes for the sheet panel. */
   className?: string;
+  /** Extra inline styles for the sheet panel (e.g. a measured maxHeight). */
+  style?: ViewStyle;
 }
 
 /**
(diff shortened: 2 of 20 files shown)


apps/mobile/src/app/task/[id].tsx
```diff
@@ -123,9 +123,9 @@ export default function TaskDetailScreen() {
     // height, so the composer sits at the keyboard top — no extra gap needed
     // when open. Closed state keeps a comfortable bottom inset.
     return {
-      marginBottom: height.value < 0 ? 0 : Math.max(insets.bottom, 50),
+      marginBottom: height.value < 0 ? 0 : composerBottom(),
     };
-  }, [insets.bottom]);
+  }, [composerBottom]);
```
Comment (dmarticus) [inline apps/mobile/src/app/task/[id].tsx]: Subtle: `useScreenInsets` returns a fresh object each render, and the getter functions are re-created on every `insets` change. `[composerBottom]` happens to be correct today, but only because nothing in the hook guarantees referential stability of these methods — a future change to `useScreenInsets` could silently break this. Two options: revert to `[insets.bottom]` (which is the actual underlying dependency), or wrap the getters in `useCallback` inside `useScreenInsets` so this pattern is safe everywhere.

apps/mobile/src/hooks/useScreenInsets.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  return useMemo(() => {
+    return {
+      /** Raw device insets, for cases the helpers below don't cover. */
+      insets,
+      /** Bottom padding = device inset + the standard gap for this surface. */
+      bottom: (variant: BottomGapVariant = "default") =>
+        insets.bottom + BOTTOM_GAP[variant],
+      /** Top padding for page-sheet content (inset + standard top gap). */
+      contentTop: () => insets.top + TOP_GAP,
```
dmarticus: `contentTop()` is asymmetric with `bottom(variant)` — it hides `TOP_GAP = 8` as a private constant with a single zero-arg getter, while `bottom` exposes a scale. Several screens still hand-roll `insets.top + 60` (mcp add-custom, installation/[id], template/[id]) and `insets.top + 56` (task/index). If those are intentional one-offs that's fine, but consider either renaming this to `sheetContentTop()` to signal "this is only the +8 sheet-content case," or accepting a variant like `bottom()` does. As written, the name reads broader than the implementation.

apps/mobile/src/hooks/useScreenInsets.ts
```diff
@@ -0,0 +1,61 @@
...
+      /** Bottom offset for a floating action button. */
+      fabBottom: () => insets.bottom + FAB_GAP,
+      /** Composer bottom margin floor (never smaller than the min). */
+      composerBottom: () => Math.max(insets.bottom, COMPOSER_MIN_BOTTOM),
```
dmarticus, line 71: Worth a one-line JSDoc on `fabBottom()` and `composerBottom()` clarifying these are domain-specific, frozen choices — not variants of `BOTTOM_GAP`. Sitting next to `bottom(variant)` they read like siblings, and a future caller will reasonably try `fabBottom("roomy")`.

apps/mobile/src/components/SheetContainer.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * bottom padding. Tapping the backdrop closes; taps inside the panel don't.
+ *
+ * Use this instead of re-implementing the `mt-auto rounded-t-2xl …` markup so
+ * every sheet shares one shape and one inset policy.
+ */
+export function SheetContainer({
+  open,
+  onClose,
+  children,
+  bottomGap = "compact",
```
dmarticus, line 45: Default `bottomGap = "compact"` matches the legacy sheet shell (both `SelectSheet` and `AttachmentSheet` previously used `+12`), but the name footgun is real: a future caller will reasonably expect the default to be `BOTTOM_GAP.default` (24). A brief comment here noting "compact is the inherited sheet default" would prevent that confusion.

apps/mobile/src/components/SheetContainer.tsx
```diff
@@ -0,0 +1,66 @@
...
+  className = "",
```
dmarticus, line 46: CLAUDE.md mobile rules say custom wrappers should accept both `className` and `style` so call sites can override without forcing inline style elsewhere. Cheap to add now; otherwise a future need for `maxHeight` or a measured value forces a fork.

apps/mobile/src/hooks/useScreenInsets.ts
```diff
@@ -0,0 +1,61 @@
...
+    };
+  }, [insets]);
+}
```
dmarticus, line 74: No test for the hook. Now that this is the single source of truth for spacing, a 6-line vitest pinning the scale (`bottom("compact")` === `insets.bottom + 12`, etc.) would catch any future "bump compact to 14" accident cheaply.

dmarticus (approved):
Nice cleanup. The hook and `SheetContainer` both pull their weight — 12/24/40 is the right level of abstraction for the scatter it's replacing, and the sheet shell is genuinely shared rather than over-fit. Approving; comments below are non-blocking nits to look at before/after merge.

One thing worth eyeballing on a simulator given you flagged it: the two declared 8px shifts (settings 32→24, pr-diff 16→24) and the `roomy` variant collapsing the previous +40 sites — 60 seconds on iPhone SE + Pro Max in both modes should be enough.

apps/mobile/src/app/task/[id].tsx
```diff
@@ -59,7 +59,7 @@ export default function TaskDetailScreen() {
   }>();
   const router = useRouter();
   const queryClient = useQueryClient();
-  const insets = useSafeAreaInsets();
+  const { insets, composerBottom } = useScreenInsets();
   const themeColors = useThemeColors();
```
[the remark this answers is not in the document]
Comment (Gilbert09) [reply apps/mobile/src/app/task/[id].tsx:67]: Good catch — fixed in 43803597. `composerBottom()` is now pre-computed into `composerBottomValue` outside the `useAnimatedStyle` callback, so the UI thread no longer calls the non-worklet getter (mirrors the `restingBottom` pattern in task/index.tsx).

apps/mobile/src/app/task/[id].tsx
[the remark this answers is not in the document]
Comment (Gilbert09) [reply apps/mobile/src/app/task/[id].tsx]: Done in 43803597 — the worklet now reads the scalar `composerBottomValue` and the dependency array is `[composerBottomValue]`.

Comment (Gilbert09) [reply apps/mobile/src/app/task/[id].tsx]: Agreed. Fixed in 43803597 by capturing the primitive `composerBottomValue` outside the worklet, so the dep array is now a stable number rather than relying on the getter's referential stability.

apps/mobile/src/hooks/useScreenInsets.ts
Gilbert09: Renamed to `sheetContentTop()` in 43803597 and updated all four callers (all of which are sheets/filter menus). Added a JSDoc note that full screens needing a measured header height should compute `insets.top + <headerHeight>` directly.

apps/mobile/src/hooks/useScreenInsets.ts
Gilbert09: Added JSDoc to both in 43803597 spelling out that the `+FAB_GAP` / `COMPOSER_MIN_BOTTOM` values are frozen domain-specific choices, not `BOTTOM_GAP` variants.

apps/mobile/src/components/SheetContainer.tsx
Gilbert09: Added a comment on `bottomGap` in 43803597 noting compact is the inherited sheet default, deliberately not BOTTOM_GAP's own "default" (24).

apps/mobile/src/components/SheetContainer.tsx
Gilbert09: Added a `style?: ViewStyle` prop in 43803597 that merges into the panel's inline style (after the shadow), so call sites can pass a measured maxHeight without forking.

apps/mobile/src/hooks/useScreenInsets.ts
Gilbert09: Added `useScreenInsets.test.ts` in 43803597 pinning the 12/24/40 scale, the sheet-top/fab offsets, and the composer floor (incl. the zero-inset case).

PR #2340, opened 2026-05-24
feat(worktree): use human-readable names for worktree branches

### LINKED ISSUE #1844: Use human-readable names for worktree branches instead of random hashes

MukundaKatta:
## Problem

Closes #1844

New worktrees were labeled with random 4-digit numbers (`crypto.randomInt(1000, 10000)`), which read like short SHA snippets and made it momentarily unclear which worktree you were on.

## Changes

- Added `packages/git/src/worktree-name.ts` with a small curated adjective-noun word list and a `generateHumanReadableName()` helper that returns names like `swift-otter-42`.
- `WorktreeManager.generateWorktreeName()` now delegates to this helper. All existing collision-handling paths and the `${name}${Date.now()}` defensive fallback are unchanged.
- Names stay lowercase, filesystem-safe (a-z, 0-9, `-`), and short (under 32 chars).
- The `~116k` combinations from 36 adjectives x 36 nouns x 90 suffix values keep collisions rare in practice.

## How did you test this?

- Added `packages/git/src/worktree-name.test.ts` covering the name shape, character safety, length bound, and rough uniqueness over 50 draws.
- `pnpm --filter @posthog/git test` -> 159 passed (8 files), including the 4 new ones.
- `pnpm --filter @posthog/git typecheck` -> clean.
- `biome check` on the touched files -> clean.

## Publish to changelog?

no

### DIFF
--- a/packages/git/src/worktree-name.test.ts
+++ b/packages/git/src/worktree-name.test.ts
@@ -7,7 +7,7 @@ describe("generateHumanReadableName", () => {
     expect(name).toMatch(/^[a-z]+-[a-z]+-\d{2}$/);
   });
 
-  it("produces varied names over multiple calls", () => {
+  it("produces varied names over many calls", () => {
     const names = new Set<string>();
     for (let i = 0; i < 50; i++) {
       names.add(generateHumanReadableName());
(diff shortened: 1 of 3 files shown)


MukundaKatta:
Addressed greptile: switched the loop-based filesystem-safe and compact-length tests to `it.each` over a pre-generated sample set, so each draw shows as a named case in the runner.

charlesvien (approved):

PR #2342, opened 2026-05-25
fix(code): pre-configure worktree with GitHub noreply email

### LINKED ISSUE #76244: Agent inconsistently auto-detects GitHub noreply email when push is blocked
When GitHub blocks a push because the commit author email is set to private, the agent's behavior is inconsistent. Sometimes it successfully discovers the user's `noreply` email (by inspecting commit history or querying the `gh` CLI) and amends the commit automatically. Other times it gives up and shows the manual unblock prompt asking the user to either toggle the GitHub email privacy setting or paste their `noreply` address.

Example of the manual prompt that appears when auto-detection fails:
Image

lost-particles:
Closes PostHog/posthog#76244

## Summary

- When the user has "Keep my email address private" enabled on GitHub, the first `git push` is rejected with GH007 ("Your push would publish a private email address"). The agent's recovery from this was inconsistent — sometimes it discovered the noreply email via `gh`/git history and amended the commit, sometimes it gave up and asked the user to either toggle the GitHub setting or paste the noreply.
- This change makes the case deterministic by pre-configuring the new worktree with the user's GitHub noreply identity at creation time, so the rejection never happens. Lookup is via `gh api user`; when GitHub reports a public email, the worktree is left untouched.
- The config is scoped to the worktree via `extensions.worktreeConfig` + `git config --worktree`, so the user's main repo identity is never modified.

## Behavior

| User state | Outcome |
|---|---|
| Email privacy on + `gh` authenticated | Worktree gets `<id>+<login>@users.noreply.github.com`; push succeeds without agent recovery |
| Public email on GitHub | `gh api user` returns a non-null email → no-op, worktree unchanged |
| `gh` missing / not authenticated | Debug-log + no-op; existing agent-side recovery still applies as a fallback |

The pre-configuration is best-effort and wrapped in try/catch — any failure logs at debug level and does not break workspace creation.

## Test plan

- [x] `pnpm --filter code test src/main/services/workspace/githubIdentity.test.ts` — 9 new tests cover the pure helper, mocked `gh` failure modes (not installed, malformed JSON, missing fields), and a real-git integration test verifying the noreply lands only on the worktree (main repo identity unchanged).
- [x] `pnpm --filter code test src/main` — all 497 existing main-process tests still pass, including the archive integration test that exercises real worktree creation.
- [x] `pnpm --filter code exec biome check` on changed files — clean.
- [ ] Manual: with a GitHub account that has email privacy enabled and `gh auth login` done, create a new task workspace and run `git config user.email` in the resulting worktree — should be `<id>+<login>@users.noreply.github.com`. The main repo's `git config --local user.email` should be unchanged.
- [ ] Manual: with a GitHub account that exposes a public email, repeat — worktree `user.email` should fall through to the user's global config, not the noreply.

## Notes

- New file: [apps/code/src/main/services/workspace/githubIdentity.ts](apps/code/src/main/services/workspace/githubIdentity.ts) — `fetchGitHubUserInfo` (5s timeout on `gh`), `computeNoreplyIdentity` (pure), `applyWorktreeIdentity`.
- Wired in [apps/code/src/main/services/workspace/service.ts](apps/code/src/main/services/workspace/service.ts) after each `worktreeManager.create*` call via a new private `preconfigureNoreplyIdentity` helper.
- `extensions.worktreeConfig` is a benign opt-in (git 2.20+) that only enables per-worktree config storage; it doesn't alter any existing setting.

### DIFF
--- a/apps/code/src/main/services/workspace/githubIdentity.test.ts
+++ b/apps/code/src/main/services/workspace/githubIdentity.test.ts
@@ -1,0 +1,171 @@
+import { execFile } from "node:child_process";
+import * as fs from "node:fs/promises";
+import * as os from "node:os";
+import * as path from "node:path";
+import { promisify } from "node:util";
+import { makeLoggerMock } from "@test/loggerMock";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../../utils/logger.js", () => makeLoggerMock());
+
+import {
+  applyWorktreeIdentity,
+  computeNoreplyIdentity,
+  fetchGitHubUserInfo,
+} from "./githubIdentity";
+
+const execFileAsync = promisify(execFile);
+
+describe("computeNoreplyIdentity", () => {
+  it("returns noreply identity when GitHub user has email privacy enabled", () => {
+    expect(
+      computeNoreplyIdentity({
+        id: 12345,
+        login: "octocat",
+        name: "Octo Cat",
+        email: null,
+      }),
+    ).toEqual({
+      name: "Octo Cat",
+      email: "12345+octocat@users.noreply.github.com",
+    });
+  });
+
+  it("falls back to login when name is missing", () => {
+    expect(
+      computeNoreplyIdentity({ id: 12345, login: "octocat", email: null }),
+    ).toEqual({
+      name: "octocat",
+      email: "12345+octocat@users.noreply.github.com",
+    });
+  });
+
+  it("returns null when GitHub user has a public email", () => {
+    expect(
+      computeNoreplyIdentity({
+        id: 12345,
+        login: "octocat",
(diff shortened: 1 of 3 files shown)


lost-particles:
Hi maintainers, this PR addresses PostHog/posthog#76244. Whenever convenient, could someone from the team take a look and let me know if the approach (pre-configuring the worktree with the GitHub noreply identity at creation time) lines up with how you would like to solve this? Happy to iterate on scope, naming, or placement based on your feedback. Thanks!

jonathanlab:
Thanks for this PR! With recent changes, agents aren't allowed to directly run `git commit` nor `git push` anymore, instead we run everything through a `git_signed_commit` hook which will use GitHub's API to create PRs instead, so a GH007 can not happen any more :) Closing

PR #2343, opened 2026-05-25
feat(code): cache per-task PR url + state for instant task switches

richardsolomou:
## Problem

Switching between tasks in the sidebar/task detail used to block on a `gh pr list --head <branch>` call to figure out which PR each task corresponds to. That made the PR badge in the sidebar and the "Open PR" button in the task header pop in noticeably late on every task switch — and on cold start, every visible card paid the cost simultaneously.

## Changes

Per-task PR info is now cached on the `workspaces` SQLite row (`pr_url`, `pr_state`, `pr_fetched_at`) and served stale-while-revalidate.

- `GitService.getTaskPrStatus` returns the cached `prState` synchronously and schedules a deduplicated background revalidation. `hasDiff` for worktree tasks still computes inline because it's local-only and cheap.
- The background revalidation runs the same `gh` lookups as before, writes through to the workspaces row, and — when the value changed — emits `WorkspaceService.taskPrInfoChanged`.
- A new `workspace.onTaskPrInfoChanged` tRPC subscription is wired in `App.tsx`; the handler updates every matching `getTaskPrStatus` query via `setQueriesData` so the badge updates in place without a refetch.
- `workspace.getCachedPrUrl(taskId)` lets `useTaskPrUrl` fall back to the cached URL so the task header's PR button opens the right PR before the live `gh` lookups return.
- The `taskPrInfoChanged` emit uses a string literal (not the `WorkspaceServiceEvent` const) to avoid a circular import — `workspace/service` eagerly loads the DI container, which re-enters `git/service`.

Migration: `0007_stiff_reptil` adds the three nullable columns to `workspaces`.

## How did you test this?

- `pnpm --filter @posthog/code typecheck` ✅
- `pnpm exec biome check apps/code/src` ✅
- `pnpm --filter @posthog/code test` ✅ (1436 pass; the 23 archive integration failures pre-exist on `main` because `git commit` is blocked in this signed-commit environment — verified by re-running them on a clean tree)

No manual UI verification was possible from this environment.

## Publish to changelog?

no

---


k11kirky:
getCachedPrUrl extends the exact router, repository pattern CLAUDE.md says not to extend

context: R10 / forbidden-patterns: "tRPC routers bypassing their service to call a repository. workspace.ts does this today; do not extend the pattern." This file is already pervasively guilty (togglePin, markViewed, getTaskTimestamps…), so it's consistent with its neighbors — but it's still net-new surface against an explicitly-named rule. Since GitService already owns the PR-cache writes, the natural home is a thin GitService.getCachedPrUrl(taskId) (or WorkspaceService) method, keeping the router a one-liner. Low effort, and it doesn't grow the debt the guide is trying to contain.

richardsolomou:
moved to a thin `GitService.getCachedPrUrl(taskId)`, router back to a one-liner. cc726b286

k11kirky (approved):

PR #2345, opened 2026-05-25
fix(agent): surface error for unsupported slash commands instead of hanging

### LINKED ISSUE #2158: Unsupported slash commands hang the chat instead of erroring
## Problem

Typing an unsupported Claude Code slash command (e.g. `/plugin install slack` from the Claude Code plugin install guide) hangs the chat indefinitely. The status bar shows "Discombobulating..." with a running timer (observed at 103m 34s, 0K/1.0M tokens) and never resolves or errors.

Thread: https://posthog.slack.com/archives/C09G8Q32R6F/p1778495163792959

## Expected behavior

When the user submits a slash command that PostHog Code does not implement, the chat should immediately surface a clear error (e.g. "Unknown command: `/plugin`") and stay responsive, instead of silently entering an unrecoverable spinning state.

## Steps to reproduce

1. Open a new task in PostHog Code
2. Type `/plugin install slack` and submit
3. Observe the chat enters "Discombobulating..." and never returns

## Notes

- Related to #670 (we do not support all slash commands), but this issue is specifically about the failure mode rather than the missing functionality. Even after #670 ships, unknown commands should still error gracefully.
- Reporter noted this was discovered while following the Claude Code plugin install guide inside PostHog Code.

lost-particles:
Closes #2158

## Summary

- When the user submits a Claude Code slash command that PostHog Code does not implement (e.g. `/plugin install slack`), the SDK consumes the input but emits no echo or `local_command_output`. The prompt loop in `ClaudeAcpAgent.prompt` then sits on the first `session_state_changed → idle` notification with `promptReplayed` still false, breaks out of the switch, and goes back to `query.next()` forever. The chat shows "Discombobulating…" indefinitely with no way to recover.
- This change detects the case specifically: when an idle arrives before any prompt replay AND the original user input was a slash command we did not handle locally, emit a clear agent message and return `end_turn` so the renderer can clear its spinner.
- Regular prompts keep the existing skip behavior, so background idles (the original purpose of the skip) do not get misclassified as failures.

## Behavior

| Input | Old behavior | New behavior |
|---|---|---|
| `/context`, `/heapdump`, `/extra-usage` (local-only commands) | Works (`isLocalOnlyCommand`, `promptReplayed=true`) | Unchanged |
| `/plugin install slack` and other SDK commands that silently consume | Loop hangs forever, "Discombobulating…" | Emits `agent_message_chunk` with `"Unsupported slash command: \`/plugin\`. PostHog Code does not implement this command."` and returns `end_turn` |
| Regular prompt where an unrelated background `idle` arrives before replay | Skipped, waits for prompt replay | Unchanged |

## Test plan

- [x] `pnpm --filter @posthog/agent test src/adapters/claude/claude-agent.slash-command.test.ts` — 2 new tests:
  - Slash command + early `idle` → emits the unsupported-command chunk and returns `end_turn`.
  - Regular prompt + unrelated early `idle` → does NOT emit the chunk; falls through to the existing `Session did not end in result` path.
- [x] `pnpm --filter @posthog/agent test` — all 446 tests pass, including the existing `claude-agent.refresh.test.ts`.
- [x] `pnpm --filter @posthog/agent typecheck` and `biome check` clean on changed files.
- [ ] Manual: open a new task in PostHog Code, type `/plugin install slack`, submit. The chat should show the "Unsupported slash command" message and stay responsive instead of spinning.

## Notes

- The detection reuses the existing `commandMatch` regex (`/^(\/\S+)/`) that the function already computes at line 350, so the diff is small and the matching rules stay consistent with the `LOCAL_ONLY_COMMANDS` check.
- The error is emitted as a regular `agent_message_chunk` (the same channel local-only commands use to forward their results), so it renders inline in the conversation rather than as a toast or error banner — matching the reporter's "surface a clear error in the chat" expectation.
- This does not enumerate which slash commands PostHog Code supports vs not (that depends on the SDK's `supportedCommands()` and on what each command needs at runtime). It just catches the "SDK swallowed it and produced nothing" failure mode, which is the actual user-visible bug.

### DIFF
--- a/packages/agent/src/adapters/claude/claude-agent.slash-command.test.ts
+++ b/packages/agent/src/adapters/claude/claude-agent.slash-command.test.ts
@@ -1,0 +1,163 @@
+import type { AgentSideConnection } from "@agentclientprotocol/sdk";
+import type { SDKMessage } from "@anthropic-ai/claude-agent-sdk";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { createMockQuery, type MockQuery } from "../../test/mocks/claude-sdk";
+import { Pushable } from "../../utils/streams";
+
+vi.mock("@anthropic-ai/claude-agent-sdk", () => ({
+  query: vi.fn(),
+}));
+
+vi.mock("./mcp/tool-metadata", () => ({
+  fetchMcpToolMetadata: vi.fn().mockResolvedValue(undefined),
+  getConnectedMcpServerNames: vi.fn().mockReturnValue([]),
+  setMcpToolApprovalStates: vi.fn(),
+  isMcpToolReadOnly: vi.fn().mockReturnValue(false),
+  getMcpToolMetadata: vi.fn().mockReturnValue(undefined),
+  getMcpToolApprovalState: vi.fn().mockReturnValue(undefined),
+}));
+
+const { ClaudeAcpAgent } = await import("./claude-agent");
+type Agent = InstanceType<typeof ClaudeAcpAgent>;
+
+interface ClientMocks {
+  sessionUpdate: ReturnType<typeof vi.fn>;
+  extNotification: ReturnType<typeof vi.fn>;
+}
+
+function makeAgent(): { agent: Agent; client: ClientMocks } {
+  const client: ClientMocks = {
+    sessionUpdate: vi.fn().mockResolvedValue(undefined),
+    extNotification: vi.fn().mockResolvedValue(undefined),
+  };
+  const agent = new ClaudeAcpAgent(client as unknown as AgentSideConnection);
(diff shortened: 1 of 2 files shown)


lost-particles:
Hi maintainers, this PR targets #2158. Whenever convenient, could a maintainer take a look and let me know if the failure-mode detection (idle-before-replay + original input is a slash command we did not handle locally) is the right place to draw the line? I considered also validating slash commands in the renderer against the `availableCommands` list before sending, but `/plugin` is in the SDK reported command set so that approach would not have caught the reported case. Happy to iterate on wording, scope, or placement based on your feedback. Thanks!

charlesvien (approved):

PR #2440, opened 2026-05-29
Show skill tool calls as "Reading `<name>` skill"

rafaeelaudibert:
## Summary
- The current display for Skill tool calls dumps the raw input as JSON, e.g. `Skill {"skill":"implementing-mcp-tools"}`.
- This change renders it as `Reading <skill-name> skill` with the skill name in the accent color, matching how we already style input previews.

## Test plan
- [ ] Trigger a session that invokes a skill (e.g. ask the agent to use the `implementing-mcp-tools` skill) and confirm the tool row reads `Reading implementing-mcp-tools skill` with the skill name highlighted.
- [ ] Confirm non-Skill tool calls (Read, Bash, Grep, MCP tools, etc.) are unchanged.
- [ ] Confirm the row is still expandable and the expanded view still shows the full raw input.

### DIFF
--- a/apps/code/src/renderer/features/sessions/components/session-update/ToolCallView.tsx
+++ b/apps/code/src/renderer/features/sessions/components/session-update/ToolCallView.tsx
@@ -74,13 +74,19 @@ export function ToolCallView({
     Wrench;
 
   const filePath = kind === "read" && locations?.[0]?.path;
-  const displayText = filePath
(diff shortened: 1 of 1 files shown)


k11kirky (approved):
Image

Looks good, would be nice for tool search too

rafaeelaudibert:
@k11kirky added tool search too!

PR #2447, opened 2026-06-01
fix: handle poor and intermittent connectivity gracefully

### LINKED ISSUE #76241: Handle poor and intermittent internet connectivity gracefully
## Problem

The app does not handle poor or intermittent internet connectivity well. Multiple users have reported issues, including:

- Sending a message while offline silently drops the message instead of queueing it or warning the user. The user has to retype and resend.
- Sessions get stuck (e.g. tool calls or streams not progressing) when the connection is flaky or constantly reconnecting, instead of recovering automatically once the connection is back.

## Scope

Audit the app's behavior under degraded network conditions (offline, packet loss, frequent reconnects) across both local and cloud sessions. Surface clear feedback to the user when the network is down, preserve unsent input, and ensure in-flight sessions recover or fail loudly rather than hanging silently.

Use this issue as a tracking issue. Attach additional user reports as they come in.

lost-particles:
## Problem

PostHog Code handled poor/intermittent connectivity poorly: sending while offline silently dropped the message, in-flight sessions could hang indefinitely, and a network-interrupted turn was shown either as a clean "finished" response or as "Interrupted by user" — never as a network failure. This addresses the reports collected in the tracking issue (offline sends lost, stuck sessions, long unexplained waits).

Closes PostHog/posthog#76241

## Changes

**Sending / input**
- A send that fails on a flaky network no longer drops the message — it's restored to the composer.
- Offline sends wait briefly for the connection (bounded grace window) instead of failing instantly; while waiting they show the **"Connection lost — waiting to reconnect…"** indicator with a live timer, then give up after ~40s.
- Locally queued follow-up messages are re-enqueued if their drain dispatch fails, instead of being lost.

**Stuck / interrupted turns**
- Main-process **prompt watchdog**: a stream that goes silent fails loudly instead of leaving the turn pending forever.
- A turn affected by a network drop is no longer presented as a clean finish (no false "finished" notification). Instead it renders an inline footer **"Response may be incomplete. Failed due to network issue"** with a **Retry** button (replacing the misleading "Interrupted by user"). Retry re-sends that turn's prompt; the label persists across retries.
- Offline give-up (~40s) cancels the hung local turn with a `connection_lost` reason rather than a user-interrupt.

**Recovery / transport**
- Cloud sessions auto-recover on the offline→online transition (previously only on window focus).
- Auth-proxy aborts the connection on a **mid-stream** upstream failure (`res.destroy`) instead of cleanly ending it, so a truncated reply surfaces as an error rather than a clean end-of-stream.
- "Slow connection? Still trying…" hint after ~8s of no stream activity.

### Screenshots

**1. Network drops mid-turn — "Connection lost — waiting to reconnect…" with a live timer**
Screenshot 2026-05-31 at 11 39 13 PM

**2. Still offline after ~40s — the turn fails with "Response may be incomplete. Failed due to network issue" + a Retry button**
Screenshot 2026-05-31 at 11 40 12 PM

**3. Clicking Retry re-runs the prompt \(if connection issue persists, it shows the first msg with "Connection lost — waiting to reconnect…" and if connection is back, it runs the flows and gives response\)**
Screenshot 2026-05-31 at 11 41 12 PM

## How did you test this?

**Automated (all run locally):**
- Added unit/integration tests and ran them: `service.test.ts` (offline grace-wait, give-up, re-enqueue, turn-complete flagging), `buildConversationItems.test.ts` (network-failed footer + prompt-text for retry), `interrupt-reason.test.ts`, `connectivityRecovery.test.ts`, `GeneratingIndicator.test.tsx`, `auth-proxy/service.test.ts`, plus updates to `agent/service.test.ts` and `useDraftSync.test.tsx`.
- `pnpm lint` (biome) — clean, 1574 files.
- `pnpm typecheck` (turbo, all packages) — 13/13 pass.
- `pnpm test` (turbo, all packages) — 1627 pass. 3 timeouts in the unrelated `archive` git-integration tests under full-suite load; confirmed passing 23/23 when run in isolation.

**Manual (live app, toggling the OS network off/on while running `pnpm dev`):**
- Verified against the running app's main-process logs: the auth-proxy returning `502` when fully offline and aborting on a mid-stream `terminated`, the offline give-up firing and cancelling with `reason: connection_lost`, and the network-interrupted turn being flagged instead of notifying "finished".
- Reproduced the flow in the screenshots above: offline retry shows the reconnecting timer; on sustained outage the turn shows "Failed due to network issue" + Retry; clicking Retry re-runs the prompt once the connection is back.

## Automatic notifications

- [x] Publish to changelog?


charlesvien:
Interesting, in practice this sounds amazing but I do wonder if we can create a better util around some of these checks. I'll review tomorrow!

lost-particles:
Good call. Took a first pass: pulled the offline-wait check into a small, reusable `waitForConnectivity(timeoutMs)` util in the connectivity feature (with its own tests), so the session service just calls into it. Happy to keep going; the per-turn give-up timers and `connectionLostPromptIds` tagging could move behind a similar helper too. Let me know what shape you'd prefer.

PR #2456, opened 2026-06-02
feat(code): TanStack Router architectural cleanup

adamleithp:
Stacks on top of #2455. Lands the high-impact cleanups from that PR's architect review; defers the navigationStore consumer migration and full store deletion to follow-up PRs that can be reviewed independently.

### PR 2 cleanups
- ✅ Check in `routeTree.gen.ts`. `.gitattributes` marks it linguist-generated so reviewers can collapse regen diffs. Fresh-clone `pnpm typecheck` works without a prior `pnpm dev`/`build`.
- ✅ TanStack Router DevTools moved to `lazy()` + `<Suspense/>` behind `import.meta.env.DEV`. The ~50KB devtools chunk no longer ships in the prod bundle (was getting included despite the conditional render).
- ✅ Drop `as string` cast in `syncToRouter` — narrower types via early-return.
- ✅ Settings `close()` now calls `router.history.back()` instead of hard `router.navigate("/code")`. Dismissing settings returns the user to their prior page (e.g. `/code/inbox`).
- ✅ Settings `open()` no longer issues `window.history.pushState` — it was colliding with `hashHistory`'s own state management.

### Break the circular import
- ✅ New `apps/code/src/renderer/navigationBridge.ts` owns all `router.navigate` calls used by Zustand stores. Stores import the bridge, not `@renderer/router`. Breaks the `store → router → routeTree → __root → store` cycle that previously worked only because of ES module live bindings.
- ✅ `SettingsCategory` moved to `features/settings/types.ts` so the type can be shared without crossing the cycle.

### Settings as a real route (PR 3)
- ✅ New `SettingsPanel.tsx` holds the visual content (left nav + sections). Identical UI to the previous `SettingsDialog` content, minus the `fixed inset-0 z-[100]` overlay positioning.
- ✅ `/settings/$category` route renders `<SettingsPanel/>` as a full-screen page.
- ✅ `__root.tsx` detects `/settings/*` matches via `useRouterState` and renders **without app chrome** for those routes (no `HeaderRow`, no `MainSidebar`, no `SpaceSwitcher`, no `TourOverlay`, no `HedgehogMode`). Settings takes the full window.
- ✅ `SettingsDialog.tsx` is now a thin overlay wrapper around `SettingsPanel`, used only in pre-router shells (`AiApprovalScreen`) where `RouterProvider` isn't mounted yet.

### Cold-boot URL restore (partial PR 5)
[...]

PR #2455, opened 2026-06-02
feat(code): introduce TanStack Router file-based routing

adamleithp:
## Summary

- Replaces the Zustand `useNavigationStore` view switch in `MainLayout` with TanStack Router (hash history for Electron `file://` compatibility).
- 10 file-based routes; task surfaces nest under `/code/*`, settings at `/settings/$category`, infra views (`command-center`, `skills`, `mcp-servers`, `folders`) at the root so future verticals have room.
- `navigationStore` + `settingsDialogStore` kept alive as transitional shims that mirror every action to `router.navigate(...)`. The ~35 nav-store and ~20 settings-dialog consumers keep working unchanged.
- TanStack Router DevTools mounted in `import.meta.env.DEV`.

## Route map

| URL | Component |
|---|---|
| `/` | redirect → `/code` |
| `/code` | `TaskInput` |
| `/code/tasks/$taskId` | `TaskDetail` |
| `/code/tasks/pending/$key` | `TaskPendingView` |
| `/code/inbox` | `InboxView` |
| `/code/archived` | `ArchivedTasksView` |
| `/settings` | redirect → `/settings/general` |
| `/settings/$category` | opens `SettingsDialog` modal (see follow-up #2) |
| `/command-center` | `CommandCenterView` |
| `/skills` | `SkillsView` |
| `/mcp-servers` | `McpServersView` |
| `/folders/$folderId` | `FolderSettingsView` |

## Known gaps — follow-up PRs

Captured here so reviewers don't read this as the chosen end-state architecture.

### PR 2 — small cleanup (~half day)
- Check in `routeTree.gen.ts` (currently gitignored — fresh-clone typecheck fails without running `pnpm dev`/`build` first).
- Fix TanStack Router DevTools tree-shaking — current conditional render keeps the devtools chunk in prod bundle (~50–80KB). Switch to `React.lazy` + conditional dynamic import.
- Drop the `as string` cast in `syncToRouter` — guard already prevents the hazard, cast just hides the type narrowing.
- Settings `close()` currently always navigates to `/code`, losing back context. Switch to `router.history.back()` (combined with the existing `window.history.back()` in the modal close, pick exactly one).

### PR 3 — Settings: commit to page or modal (~1 day)
Today `/settings/$category` renders `null` and toggles a Zustand `isOpen` flag, which mounts a Radix Dialog from `__root`. The URL changes but the chrome (sidebar, header) still shows underneath — incoherent. Either:
[...]

PR #2457, opened 2026-06-02
feat(code): convert navigationStore to router-derived facade

adamleithp:
Stacks on top of #2456. Closes out the dual-source-of-truth concern raised in #2455's architect review: the Zustand `navigationStore` no longer holds its own `view` state — it's now a thin facade that derives `view` from the router and the React Query task cache on every render.

## What changed

`useNavigationStore` keeps the same public API surface (`(selector?)`, `.getState()`, `.setState()`) so the ~91 existing call sites don't change. Internally:

- **No Zustand state.** Hook subscribes to `useRouterState` + `useTaskInputPrefillStore` to know when to re-render. Snapshot is rebuilt from those two sources on every call.
- **`view.data` populates from cache.** `getCachedTask(taskId)` in `@utils/queryClient` walks the React Query `["tasks", "list"]` queries. Consumers reading `view.data?.id` keep working; the value is just `undefined` until `useTasks()` resolves (same as the deep-link case in the old store).
- **Actions delegate.** Each `navigateTo*` action calls a `navigationBridge` function. `goBack`/`goForward` use `router.history`. No more `syncToRouter` (no state to mirror).
- **Transient TaskInput prefill moves out.** `initialPrompt`, `reportAssociation`, `initialCloudRepository`, `initialModel`, `initialMode`, `folderId`, `taskInputRequestId` now live in `useTaskInputPrefillStore` (introduced in #2456 for exactly this). `navigateToTaskInput(options)` writes prefill → navigates to `/code`. The derived view reads prefill and surfaces those fields under `view.*` exactly like before.
- **Async side effects preserved.** `navigateToTask` still does the workspace/folder reconciliation it always did — that logic is unchanged.
- **Analytics context** now fires from a `router.subscribe("onResolved")` listener at module init instead of from inside each action.

Persistence, the history stack, and `hydrateTask` are deleted. Persistence is replaced by router URL + the cold-boot localStorage restore from #2456. The history stack is replaced by `router.history`. `hydrateTask` becomes a no-op — the URL **is** the source of truth, no rehydration needed.

## New helpers in `navigationBridge`

- `getCurrentMatches()`, `getCurrentLocation()` — non-React router state accessors.
- `subscribeToRouterResolved(handler)` — hook into route resolution from outside React.
- `goForwardInHistory()` — for symmetry with `goBackInHistory`.

These keep the store cycle-free (`navigationStore` never imports `@renderer/router` directly).

## Files touched

| File | Change |
|---|---|
| `stores/navigationStore.ts` | rewritten (~400 lines → ~200) |
| `stores/navigationStore.test.ts` | rewritten (~290 → ~200, 16 → 12 tests) |
| `routes/code/tasks/$taskId.tsx` | removed obsolete URL→store sync effect |
| `navigationBridge.ts` | added 4 helpers |
| `utils/notifications.test.ts` | mocks `useNavigationStore.getState` directly |

## What's NOT in this PR

- **Settings store consumers** (PR 7): the `settingsDialogStore` is still alive. Its consumers can move to `useNavigate` / `<Link/>` directly once this lands.
- **Literal deletion of the file.** `navigationStore.ts` still exists as a hook export. Removing the file means migrating all 91 call sites to `useNavigate`/`useParams`/`useMatch`. Worth doing, but the diff would dwarf the architectural change in this PR. I'd schedule it as PR 8 once the facade has run in production for a release cycle.

## Test plan

Automated (passing):
- [x] Renderer typecheck (`pnpm exec tsc -p tsconfig.web.json --noEmit`)
- [x] Vitest suite — 131 files / 1594 tests pass (down from 1598: 4 deleted persistence/history tests no longer apply)
- [x] Biome lint clean

Manual (please verify):
- [ ] All previous flows from #2455 + #2456 still work
- [ ] Sidebar highlight tracks the active task across clicks
- [ ] `view.data` populates correctly in `HeaderRow` task section (skill buttons, diff stats badge, etc.) when viewing a task
- [ ] Cloud + local task headers still render (consumers in `GlobalEventHandlers` that pass `view.data` to handoff buttons)
- [ ] `useArchiveTask` still navigates away from an archived task correctly (it reads `view.data` to detect the current task)
- [ ] Deep link to a task that isn't in the task list yet → `view.data` is `undefined` briefly, then populates once `useTasks` resolves (same behavior as before, this case existed previously)
- [ ] Settings UX from #2456 still works
- [ ] Devtools panel shows current route correctly in dev mode


adamleithp:
## 🩹 Stabilization pass (commit `45fa39692`)

Shaking this branch out surfaced three independent defects from the migration, all presenting as *"routes stuck loading / navigation locked"*. All fixed here; verified live (new-task flow, deep links, cross-route nav, mid-load navigate-away).

### Fixes

1. **Circular import → broken code-split chunks.** `router → routeTree.gen → __root → hooks → navigationBridge → router` produced a TDZ (`Cannot access 'rootRouteImport' before initialization`) under `autoCodeSplitting`, leaving route chunks stuck. Added `routerRef` (leaf module holding the singleton); `navigationBridge` now reads `getRouter()` instead of importing the router, severing the only route-tree edge back to `router.ts`.

2. **Infinite render storm.** `useAppView` returned a **fresh object every render** (the old `navigationStore.view` was a stable ref). `SidebarMenu`'s `[view]` effect refired forever → `markViewed` mutation → cache write → re-render (~50×/2s, confirmed via `[ipc-rate]` warnings), starving the UI thread and blocking both navigation and session start ("Starting task…" hang). Memoized `useAppView` on the route's primitive values + stable prefill ref; `SidebarMenu` reads `view.taskId`.

3. **Blocking loader could hang the router.** `ensureQueryData(getTask)` never resolves for optimistic/cloud-pending tasks → route stuck pending → un-navigable. Task-detail loader is now **synchronous + cache-only**; the cold-deep-link fetch + spinner moved into the component so a hang only affects that view. `openTask` seeds the detail cache so in-app opens never fetch.

### Also added (per-route loading UX)
Router `context: { queryClient }`, `createRootRouteWithContext`, `defaultPendingMs: 0` + `defaultPendingComponent` (`RoutePending`). Navigation commits **instantly** with a per-route pending slot that can later become skeletons.

---

## 🛣️ Follow-up roadmap (separate PRs after the stack merges)

Full architect review exists locally; condensed:

- **P0 — Delete `useAppView`.** It re-derives a view-enum the router already owns (second source of truth) and *requires* hand-memoization — the footgun that caused defect #2. Replace with primitive `useRouterState` selectors + per-route `Route.useParams()`; move the prefill merge into the `/code/` route. Removes the storm class permanently.
- **P1 — Rip out `navigationBridge`.** Migrate call sites to typed `<Link>` / `useNavigate()`; they read the router from React context, removing the import cycle *structurally* so `routerRef` can also go.
- **P1 — Add `defaultErrorComponent` + `defaultNotFoundComponent`.** Currently none — a thrown loader/render = blank screen (runtime already warned about it).
- **P1 — Finish loader integration + skeletons.** Shared `queryOptions` per data route (pattern in `features/tasks/queries.ts`), loaders always non-blocking, per-route `pendingComponent` skeletons.
- **P2 — Typed search params** (`validateSearch`) as URL state to retire `settingsPageStore` / `taskInputPrefillStore`; auth as `beforeLoad` guards; tune `defaultPendingMs` once loaders are async.
- **P3 — Regression test** asserting the view hook returns a **stable reference** across re-renders (guards the storm forever) + a lint guard for hooks returning fresh objects.

> `routeTree.gen.ts` intentionally left untouched — local generator output differs from the committed file only by quote/semicolon style; that's a formatter-config cleanup, out of scope here.

adamleithp:
Superseded by #2469 — the three-PR stack is collapsed into a single PR (rebased on main, conflicts resolved, CI green: typecheck + biome + 1594 unit tests). No intermediate layer was independently green, so the stack was collapsed rather than fixed three times. Closing in favor of #2469.

PR #2459, opened 2026-06-02
feat: mcp apps for posthog's mcp

skoob13:
## Problem

We'd like to render MCP apps for PostHog's MCP. However, the CLI MCP mode is not compliant with the MCP spec, so we needed to alter the MCP apps renderer.

## Changes

Standard MCP Apps bind a tool to its UI app upfront via registration metadata. PostHog instead exposes all UI apps through one generic `exec` tool and puts the `ui://` resource URI on each tool-call *response*. This branch adds that per-call resolution path:

- **Shared types** — `POSTHOG_EXEC_TOOL_KEY` constants + `resolveResultResourceUri()` (reads the URI from a result's `_meta`, modern key with legacy fallback) + `getUiResourceByUriInput` schema.
- **Main service** — new `getUiResourceByUri()` (+ shared `fetchUiResourceByUri`/`doFetchUiResource` with caching); special-cases `posthog`/`exec` during discovery; treats `"No server config"` boot-race failures as transient.
- **Router** — new `getUiResourceByUri` procedure.
- **Renderer** — `McpAppHost`/`McpToolBlock` resolve the URI from `toolCall.rawOutput` (persisted → survives restarts), fetch by URI, dedup result delivery per `toolCallId`, and re-fetch on `onDiscoveryComplete`.
- **Auth** — adds `x-posthog-mcp-consumer: posthog-code` header.

## How did you test this?

Tests added across all three layers (`service.test.ts`, `mcp-apps.test.ts`, `auth-adapter.test.ts`).

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/apps/code/src/renderer/features/sessions/components/session-update/McpToolBlock.tsx
+++ b/apps/code/src/renderer/features/sessions/components/session-update/McpToolBlock.tsx
@@ -9,11 +9,7 @@ import {
 } from "@shared/types/mcp-apps";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { useSubscription } from "@trpc/tanstack-react-query";
-import { logger } from "@utils/logger";
-import { useEffect } from "react";
 import type { ToolViewProps } from "./toolCallUtils";
-
-const log = logger.scope("mcp-tool-block");
 
 interface McpToolBlockProps extends ToolViewProps {
   mcpToolName: string;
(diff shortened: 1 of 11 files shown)


apps/code/src/renderer/features/sessions/components/session-update/McpToolBlock.tsx
```diff
@@ -46,6 +74,28 @@ export function McpToolBlock(props: McpToolBlockProps) {
     }),
   );
 
+  useEffect(() => {
+    log.info("render state", {
```
k11kirky: should probably remove this

k11kirky (approved):
another nit:

HOST_INFO in useAppBridge.ts:65 is still { name: "Twig", ... }, even though this PR renamed the client identity to posthog-code in both service.ts:171 and createConnection.

PR #2460, opened 2026-06-02
Remove "Stop All" button from command center toolbar

k11kirky:
## Problem

Two pieces of Command Center feedback from the thread:

1. The toolbar had a **"Stop All"** button that cancels every running/waiting agent at once — an easy-to-hit, broad action whose value was questioned.
2. After trimming the grid down to a couple of tasks, opening a task **full-screen and coming back re-filled all the empty cells** with recent tasks, so the curated layout was lost every time.

## Changes

**Remove "Stop All"**
- Removed the button from `CommandCenterToolbar`, plus the now-unused `stopAll` handler, `hasActiveAgents` check, `cells` prop, and the `getSessionService` / `Stop` icon / `CommandCenterCellData` imports.
- `CommandCenterView` no longer passes `cells` to the toolbar.

**Stop autofill from clobbering a curated grid**
- Autofill re-ran on every mount (its guard was a per-mount `useRef`), so navigating away and back re-topped-up empty cells. It's now a one-time bootstrap gated on a **persisted `hasAutofilled` flag**.
- The flag is set once autofill populates the grid, when the grid is already full, or when the user manually assigns a task — so a curated layout (including deliberately empty cells) survives navigation.
- The flag stays unset when there are no eligible tasks yet, so autofill can still bootstrap the grid once recent tasks exist.
- The existing "top up empty slots on first open" behaviour is preserved.

## How did you test this?

- `vitest run` on `commandCenterStore.test.ts` and `useAutofillCommandCenter.test.ts` — 26 tests pass, including new cases: autofill no longer tops up once `hasAutofilled` is set, the flag is set on fill / full grid / manual assign, and it stays unset when there are no candidates.
- `tsc -p tsconfig.web.json` reports no errors in the command-center files; `biome check` is clean.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/apps/code/src/renderer/features/command-center/components/CommandCenterToolbar.tsx
+++ b/apps/code/src/renderer/features/command-center/components/CommandCenterToolbar.tsx
@@ -1,15 +1,10 @@
-import { getSessionService } from "@features/sessions/service/service";
 import {
   MagnifyingGlassMinus,
   MagnifyingGlassPlus,
-  Stop,
   Trash,
 } from "@phosphor-icons/react";
 import { Flex, Select, Text } from "@radix-ui/themes";
-import type {
-  CommandCenterCellData,
-  StatusSummary,
-} from "../hooks/useCommandCenterData";
+import type { StatusSummary } from "../hooks/useCommandCenterData";
 import {
   type LayoutPreset,
   useCommandCenterStore,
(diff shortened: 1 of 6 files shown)


sakce (approved):
oh yeah

PR #2467, opened 2026-06-02
Introduce as PostHog Slack app in Slack-originated runs

VojtechBartos:
Adds an identity block to the cloud system prompt so Slack-originated runs introduce themselves as "PostHog Slack app" instead of Claude. Scoped to `origin === "slack"` only.

## Showcase

Screenshot 2026-06-02 at 22 56 12

### DIFF
--- a/packages/agent/src/server/agent-server.ts
+++ b/packages/agent/src/server/agent-server.ts
@@ -1687,7 +1687,7 @@ export class AgentServer {
     const identityInstructions = isSlack
       ? `
 # Identity
-You are the PostHog Slack app, PostHog's agent for helping users with their product data and codebase from Slack. When introducing yourself or referring to yourself in messages to the user, identify as "PostHog Slack app". Do NOT refer to yourself as Claude, an Anthropic assistant, or any underlying model name.
+You are the PostHog Slack app, PostHog's agent for helping users with their product data and coding tasks from Slack. When introducing yourself or referring to yourself in messages to the user, identify as "PostHog Slack app". Do NOT refer to yourself as Claude, an Anthropic assistant, or any underlying model name.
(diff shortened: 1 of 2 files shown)


packages/agent/src/server/agent-server.ts
```diff
@@ -1683,6 +1683,13 @@ export class AgentServer {
   private buildCloudSystemPrompt(prUrl?: string | null): string {
     const taskId = this.config.taskId;
     const shouldAutoCreatePr = this.shouldAutoPublishCloudChanges();
+    const isSlack = this.getCloudInteractionOrigin() === "slack";
+    const identityInstructions = isSlack
+      ? `
+# Identity
+You are the PostHog Slack app, PostHog's agent for helping users with their product data and codebase from Slack. When introducing yourself or referring to yourself in messages to the user, identify as "PostHog Slack app". Do NOT refer to yourself as Claude, an Anthropic assistant, or any underlying model name.
```
joshsny:
```suggestion
You are the PostHog Slack app, PostHog's agent for helping users with their product data and coding tasks from Slack. When introducing yourself or referring to yourself in messages to the user, identify as "PostHog Slack app". Do NOT refer to yourself as Claude, an Anthropic assistant, or any underlying model name.
```

joshsny: minor suggestion so that it knows it can code and not just answer questions

joshsny (approved):
LGTM - left a small suggested change to the wording

packages/agent/src/server/agent-server.ts
VojtechBartos: 🚀 thanks!

PR #2490, opened 2026-06-04
fix(auth): use scoped_teams when scoped_organizations is empty

raquelmsmith:
## human description

I could not get an agent to run when using local PostHog. After an hour with claude it tells me this is the fix - and now it's working locally. 

## Summary
- `scoped_organizations` is empty when the OAuth grant is project-level — PostHog returns `scoped_teams` instead. The old code only consulted `scoped_organizations`, so project-scoped logins ended up with an empty `orgProjectsMap`, no `currentProjectId`, and "No projects available" in the switcher (downstream: `Missing auth credentials` and the agent failing to start a session).
- Add `scoped_teams` to the OAuth token Zod schema and, when `scoped_organizations` is empty, build the org/project map from `scoped_teams`: fetch each project via `/api/projects/<id>/`, group by `organization`, then fetch each org's name. Only the teams the user actually authorized are included.
- Org-level grants are unchanged — `scoped_organizations` path still wins when populated.

## Test plan
- [ ] Log out, log back in selecting **Dev** region, grant access to a single project on the PostHog consent screen.
- [ ] Verify the project switcher shows the org and the granted project, and a session starts (no `Missing auth credentials` in `~/.posthog-code/logs-dev/main.log`).
- [ ] Re-test with an org-level grant (e.g. against US/EU) to confirm the existing path still works.

### DIFF
--- a/apps/code/src/main/services/auth/service.ts
+++ b/apps/code/src/main/services/auth/service.ts
@@ -551,15 +551,23 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
     options: TokenResponseOptions,
   ): Promise<InMemorySession> {
     const scopedOrgIds = tokenResponse.scoped_organizations ?? [];
+    const scopedTeamIds = tokenResponse.scoped_teams ?? [];
     const { accountKey, currentOrgId } = await this.fetchUserContext(
       tokenResponse.access_token,
       options.cloudRegion,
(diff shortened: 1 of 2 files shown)


raquelmsmith:
convo here https://posthog.slack.com/archives/C09G8Q32R6F/p1780610822021469

PR #2491, opened 2026-06-04
feat(code): top-level app nav rail (Home / Inbox / Code)

adamleithp:
## Dev-only for now

This is **gated behind the `project-bluebird` feature flag and is not enabled for any real users.** It only shows up for people who have the flag turned on — i.e. us, internally, while the Home space is built out.

- **Flag on** (default in dev builds, plus anyone we explicitly target): the app rail appears with Home / Inbox / Code.
- **Flag off** (every prod user today): the app is **byte-for-byte the current code-only shell** — no rail, `/` and `/inbox` redirect to `/code`, and the Code chrome is untouched.

So merging this is a no-op for users until we deliberately flip the flag. It's safe to land and iterate on behind the flag.

## What

Adds a Slack-like vertical **app nav rail** that switches between three top-level spaces:

- **Home** — empty placeholder for now (this is what we're still building).
- **Inbox** — the existing inbox, full-screen.
- **Code** — the existing app, unchanged.

This is the minimal top-nav extraction from the larger canvas branch — none of the canvas / dashboards / channels machinery comes along.

## Changes

- `components/AppNav.tsx` — the rail: active-space highlighting, the inbox actionable-report badge (`useInboxSignalCount`, reuses the inbox feature's existing query — no extra polling), and 2.5rem top padding to clear the titlebar.
- `routes/index.tsx` — `/` renders an empty Home space (was a `redirect → /code`).
- `routes/inbox.tsx` — new top-level route mounting the existing `InboxView`.
- `routes/__root.tsx` — renders the rail (only when the flag is on) and branches Home/Inbox to a chrome-less `Outlet`; Code keeps its existing chrome. Redirect guard sends stranded rail-only paths to `/code` once flags resolve when the flag is off.
- `shared/constants.ts` — `PROJECT_BLUEBIRD_FLAG`.
- `hooks/useFeatureFlag.ts` — `useFeatureFlagsLoaded()` so the redirect waits for a trustworthy flag value rather than the pre-load `false` default.

## Testing

- `pnpm --filter code typecheck` — clean.
- Biome — clean.
- Did not run the Electron app in this environment.

---

### DIFF
--- a/apps/code/src/renderer/components/AppNav.tsx
+++ b/apps/code/src/renderer/components/AppNav.tsx
@@ -3,12 +3,6 @@ import { CodeIcon, HouseIcon, TrayIcon } from "@phosphor-icons/react";
 import { Badge, Button } from "@posthog/quill";
 import { Box, Flex } from "@radix-ui/themes";
 import { useNavigate, useRouterState } from "@tanstack/react-router";
-import { isMac } from "@utils/platform";
-
-// macOS draws the traffic lights over the top-left of the window
-// (titleBarStyle: hiddenInset). Reserve space so the first rail button clears
-// them.
-const MAC_TRAFFIC_LIGHT_INSET = 28;
 
 type AppNavItem = {
   id: "home" | "inbox" | "code";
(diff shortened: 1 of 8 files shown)


raquelmsmith (commented):
Hmmmm I'm not convinced we need this. Why add another sidebar instead of just using the inbox item in the existing left sidebar?

PR #2508, opened 2026-06-06
fix(agent): detect feature flags product across tool-name variants

raquelmsmith:
## human desc

The "Products used: Feature flags" chip never showed when I asked the agent to tell me how many flags were rolled out this week. that's because it didn't check for plurals, nor did it parse activity log queries for product names.

This PR does those two things.

## Problem

The session resources bar above the chat input shows which PostHog products a turn touched. During a session that investigated feature-flag rollouts, the **Feature flags** chip never appeared. Two gaps in the `exec`-call classifier (`packages/agent/src/posthog-products.ts`) caused it:

1. **Naming mismatch.** Matching was an exact domain *prefix* (`name === domain || name.startsWith(`${domain}-`)`) against the singular key `feature-flag`. The live MCP also ships:
   - plural `feature-flags-*` (e.g. `feature-flags-activity-retrieve`) — `startsWith("feature-flag-")` is false (char after `feature-flag` is `s`).
   - verb-first `create-feature-flag` / `update-feature-flag` / `delete-feature-flag` — don't start with `feature-flag-`.

   All fell through to the generic `posthog` fallback, showing a plain "PostHog" chip.

2. **Activity-log scope ignored.** The natural way to answer "what flags changed this week" is `activity-log-list` / `advanced-activity-logs-list` with `scope: "FeatureFlag"`. Those domains map to `null` (generic audit readers) and the classifier never inspected the `scope` argument, so the calls surfaced nothing.

## Changes

- **Pattern-based domain matcher.** Kept the single canonical domain keys and replaced the prefix check with a precompiled per-domain regex `(^|-)<domain>s?($|-)` — the domain matches as a complete hyphen-delimited token run anywhere in the sub-tool name, tolerating a plural `s`. One pattern covers prefix, verb-first, and plural forms for **every** product, so it can't drift out of date as tools are added.
- **Activity-log scope attribution.** Added `ACTIVITY_SCOPE_PRODUCT` (PascalCase scope → product; admin/meta scopes intentionally omitted) and `classifyPostHogActivityLog()`, and branched `classifyPostHogExecCall` for `activity-log*` / `advanced-activity-logs*`. The name-only `classifyPostHogSubTool` still returns `null` for those domains, so existing behavior is unchanged.

## How did you test this?

- `vitest run src/posthog-products.test.ts` — 58 passed. Added cases for plural/verb-first variants (incl. non-flag `create-survey` / `create-experiment` to prove generality) and for activity-log scope (single, multiple-deduped, empty/unscoped, admin scope, no command text).
- Biome clean on both files.

## Automatic notifications

- [ ] Publish to changelog?

## 🤖 Agent context

Diagnosed from a real session: feature-flag rollout data was pulled via `activity-log-list {scope: FeatureFlag}` and `feature-flags-activity-retrieve`, neither of which the classifier attributed to Feature flags. Fix targets the matcher (generalizes naming) and adds scope-aware activity-log attribution.

### DIFF
--- a/packages/agent/src/posthog-products.test.ts
+++ b/packages/agent/src/posthog-products.test.ts
@@ -197,6 +197,13 @@ describe("classifyPostHogExecCall", () => {
         'call activity-log-list {"scope":"FeatureFlag"}',
       ),
     ).toEqual(["feature_flags"]);
+    // Web analytics is only reachable from the activity log via this scope.
+    expect(
+      classifyPostHogExecCall(
+        "activity-log-list",
+        'call activity-log-list {"scope":"WebAnalyticsFilterPreset"}',
+      ),
+    ).toEqual(["web_analytics"]);
   });
 
   it("attributes advanced-activity-logs to every recognized scope, deduped", () => {
--- a/packages/agent/src/posthog-products.test.ts
+++ b/packages/agent/src/posthog-products.test.ts
@@ -46,6 +46,21 @@ describe("classifyPostHogSubTool", () => {
     },
   );
 
+  // Plural and verb-first tool names resolve to the same domain product as the
+  // canonical singular form — no per-variant entry required.
+  it.each([
+    ["feature-flags-activity-retrieve", "feature_flags"],
+    ["feature-flags-status-retrieve", "feature_flags"],
+    ["create-feature-flag", "feature_flags"],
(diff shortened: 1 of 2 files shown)


charlesvien (approved):
🚀🚀🚀🚀🚀

PR #2515, opened 2026-06-07
feat(code): add Visual Studio Code style UI zoom

daniektj:
Adds window-level zoom like VS Code: `Ctrl/⌘ +`, `-`, and `0`, the View menu (Zoom In/Out/Reset), and a control in Settings → Appearance with a live percentage.

The zoom level is owned by a main-process `ZoomService`, persists across restarts (stored alongside window state), and is re-applied after web-contents reloads (zoom resets on every load). Keyboard handling lives in a pure, unit-tested `resolveZoomIntent` helper instead of react-hotkeys-hook, which mis-parses the `=`/`-`/`0` symbol keys.

There is no OS-specific code, so the behavior is portable to macOS and Linux (the native window controls intentionally stay fixed, matching VS Code).

## Testing
- 26 unit tests: key-intent mapping, clamp/percent math, and `ZoomService` behavior (zoom in/out/reset, clamping, persistence, restore)
- `typecheck` and `biome` pass clean
- Manually verified in a packaged Windows build: keyboard shortcuts, View menu, Settings control with live %, and persistence across a restart

---

### DIFF
--- a/apps/code/src/main/di/container.ts
+++ b/apps/code/src/main/di/container.ts
@@ -70,6 +70,9 @@ import { UpdatesService } from "../services/updates/service";
 import { UsageMonitorService } from "../services/usage-monitor/service";
 import { WatcherRegistryService } from "../services/watcher-registry/service";
 import { WorkspaceService } from "../services/workspace/service";
+import type { ZoomPersistence } from "../services/zoom/schemas";
+import { ZoomService } from "../services/zoom/service";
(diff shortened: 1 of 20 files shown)


charlesvien:
hey @daniektj, I can already do Ctrl/⌘ +, -, and 0, the View menu (Zoom In/Out/Reset) from the native electron renderer? I'm confused on if this is solving a specific issue you were having?

PR #2519, opened 2026-06-08
feat(cloud-task): Durable run streaming through agent-proxy

charlesvien:
## Problem

Moves event streaming to agent-proxy

## Changes

1. Route the event-ingest POST to the standalone agent-proxy via new POSTHOG_TASK_RUN_EVENT_INGEST_URL
2. Resolve the read leg once via stream_token, then read from the proxy with a run-scoped Bearer token (falls back to Django)
3. Add a stream-end sentinel as the authoritative end-of-stream signal
4. Make the reconnect loop status-unaware: stop only on stream-end or budget exhaustion, never on polled status
5. Speed up reconnect backoff (flat 500ms for 3 attempts, then exponential to 30s; ~1.5s recovery vs ~16s)
6. Cover ingest routing, read-leg routing and stream-end in tests

## How did you test this?

Manually

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/core/src/cloud-task/cloud-task.ts
+++ b/packages/core/src/cloud-task/cloud-task.ts
@@ -624,7 +624,7 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
 
     if (watcher.needsStopAfterBootstrap) {
       watcher.needsStopAfterBootstrap = false;
-      this.stopWatcher(key);
+      await this.finalizeWatcherStop(key);
       return;
     }
 
@@ -1304,8 +1304,7 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
     // the run went terminal but before stream-end arrived. Status is updated for display only, from
     // the task_run_state events the stream itself carries.
     if (watcher.streamEnded) {
-      this.emitStatusUpdate(watcher);
-      this.stopWatcher(key);
+      await this.finalizeWatcherStop(key);
       return;
     }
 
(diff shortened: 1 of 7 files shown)


charlesvien:
* **#2519** <a href="https://app.graphite.com/github/pr/PostHog/code/2519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2519?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/core/src/cloud-task/cloud-task.ts
```diff
@@ -1184,79 +1334,61 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
   private async handleStreamCompletion(
```
tatoalo, line 1407: what happens if `stream-end` sentinel is not there? we seem to recover based on that sentinel but with no fallback

packages/core/src/cloud-task/cloud-task.ts
```diff
@@ -1413,6 +1550,54 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
...
   }
 
+  private async resolveStreamTarget(watcher: WatcherState): Promise<void> {
+    const url = `${watcher.apiHost}/api/projects/${watcher.teamId}/tasks/${watcher.taskId}/runs/${watcher.runId}/stream_token/`;
+    try {
+      const response = await this.auth.authenticatedFetch(url, {
+        method: "GET",
+      });
+      if (!response.ok) {
+        // Refused, or an old server without the endpoint: read from Django with status polling.
```
tatoalo: maybe here we can treat transient failures?

packages/core/src/cloud-task/cloud-task.ts
```diff
@@ -1184,79 +1334,61 @@ export class CloudTaskService extends TypedEventEmitter<CloudTaskEvents> {
...
-          runId: watcher.runId,
-          kind: "status",
-          status: watcher.lastStatus ?? undefined,
-          stage: watcher.lastStage,
-          output: watcher.lastOutput,
-          errorMessage: watcher.lastErrorMessage,
-          branch: watcher.lastBranch,
-        });
+      if (run) {
+        this.applyTaskRunState(watcher, run);
```
tatoalo, line 1448: we are not resetting reconnection attempts anymore here and `cumulativeReconnectAttempts` when polled progress showed shows the run progressed, do you think we can drop that?

tatoalo (approved):
few questions

PR #2534, opened 2026-06-08
Add "why" context and Slack thread link to agent-created PRs

timgl:
## Problem

When the PostHog Slack app opens a PR, reviewers only see the diff — not *why* the change was requested, with no path back to the conversation. PR descriptions also tended to over-explain every change.

**Why:** Tim asked for agent-created PRs to carry the context behind the request plus a link to the originating Slack thread, and to keep the description brief.
Slack thread: https://posthog.slack.com/archives/C09G8Q32R6F/p1780952041801889

## Changes

Updated the cloud PR-creation prompt in `packages/agent/src/server/agent-server.ts` so the agent now:

- Adds a brief **Why** to the PR body (the user's motivation, not a restatement of the diff).
- Links the originating Slack thread when the task started from one (read from task run state `slack_thread_url`).
- Keeps the PR description brief — summarizes the most important changes instead of enumerating every one.

Applied to both PR-creation paths (repository auto-publish and no-repository clone-and-PR).

## How did you test this?

- `pnpm --filter agent typecheck` and `biome check` — clean.
- `pnpm exec vitest run src/server/agent-server.test.ts` — all 57 pass (env matching CI). Added 6 tests covering the Why, embedded/fallback thread link, brevity, and the no-repository path.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/agent/src/server/agent-server.test.ts
+++ b/packages/agent/src/server/agent-server.test.ts
@@ -198,9 +198,15 @@ interface TestableServer {
   ): Promise<void>;
   detectAndAttachPrUrl(payload: unknown, update: unknown): void;
   detectedPrUrl: string | null;
-  buildCloudSystemPrompt(prUrl?: string | null): string;
+  buildCloudSystemPrompt(
+    prUrl?: string | null,
+    slackThreadUrl?: string | null,
+  ): string;
   buildDetectedPrContext(prUrl: string): string;
-  buildSessionSystemPrompt(prUrl?: string | null): string | { append: string };
+  buildSessionSystemPrompt(
+    prUrl?: string | null,
+    slackThreadUrl?: string | null,
+  ): string | { append: string };
   buildCodexInstructions(systemPrompt: string | { append: string }): string;
   getRuntimeAdapter(): "claude" | "codex";
 }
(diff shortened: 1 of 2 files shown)


packages/agent/src/server/agent-server.test.ts
```diff
@@ -1456,6 +1462,56 @@ describe("AgentServer HTTP Mode", () => {
...
+        process.env.POSTHOG_CODE_INTERACTION_ORIGIN = "slack";
+        const s = createServer();
+        const prompt = (
+          s as unknown as TestableServer
+        ).buildCloudSystemPrompt();
+        expect(prompt).toContain(
+          "If the task started from a Slack thread, link to that thread.",
+        );
+        delete process.env.POSTHOG_CODE_INTERACTION_ORIGIN;
+      });
```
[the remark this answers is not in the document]
timgl: Done in eb79e80 — merged the duplicate Slack/no-arg prompt builds (and the no-repository pair) into single tests that assert why-context, brevity, and the thread-link fallback against one constructed prompt. Also wrapped the env mutation in try/finally to prevent leakage.

packages/agent/src/server/agent-server.ts
```diff
@@ -1718,6 +1730,13 @@ we want:
   Generated-By: PostHog Code
   Task-Id: ${taskId}`;
 
+    const threadLinkInstruction = slackThreadUrl
+      ? ` Since this task started from a Slack thread, also link it: ${slackThreadUrl}.`
+      : isSlack
+        ? " If the task started from a Slack thread, link to that thread."
```
adboio: i think we should remove this fallback, assuming the agent won't actually have a way to find the slack URL when it's not provided ?

adboio (changes requested):

packages/agent/src/server/agent-server.ts
timgl: Good call — removed the fallback in 3448fd0. Now the thread link is only added when a concrete `slack_thread_url` is injected (read from the task run state); when it is not available we simply omit it rather than telling the agent to link a thread it has no way to locate.

adboio (approved):

PR #2542, opened 2026-06-09
fix(message-editor): only navigate prompt history when input is empty

k11kirky:
## Problem

In the message editor (e.g. the new task page), `Shift+Up`/`Shift+Down` cycled through prompt history instead of selecting text, and plain `Up`/`Down` also jumped into history whenever the cursor sat at the start or end of the input. This broke normal multi-line selection and arrow-key editing.

**Why:** Reported in Slack — selecting multiple lines of the current prompt with Shift+Up/Down unexpectedly cycled through previous prompts. https://posthog.slack.com/archives/C09G8Q32R6F/p1780997423011589?thread_ts=1780997423.011589&cid=C09G8Q32R6F

## Changes

- Prompt history navigation now triggers only when the input is **empty**. Removed the `Shift` override and the cursor-at-edge (`isAtStart`/`isAtEnd`) triggers, so `Shift+Arrow` selection and arrow-key movement work normally while editing.
- Updated the keyboard-shortcuts help entries to reflect `Up`/`Down` (when input is empty).

## How did you test this?

- `pnpm --filter code typecheck` passes.

## Automatic notifications

- [ ] Publish to changelog?

---


sakce (approved):
seems reasonable

PR #2547, opened 2026-06-09
feat(code): Inbox 2.0

Twixes:
## Summary

Replaces the legacy Inbox list with the v2 Self-driving: Pull requests, Reports, Agents, plus scoped For you filtering.
Polished list/detail UI for PRs, reports, and agent runs, including repo/source metadata, reviewer affordances, and task-log-backed run details.
Promotes agent configuration into a first-class view.

Screenshot 2026-06-09 at 13 29 29@2x


sortafreel (approved):

joshsny (approved):
LGTM, your change to the remote debugging port looks unintentional

Twixes:
Tis intentional, but better belongs on #2527 @joshsny - moving the change there

apps/code/src/renderer/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
+import { Button as UiButton } from "@components/ui/Button";
+import { ConventionalCommitScopeTag } from "@features/inbox/components/ConventionalCommitScopeTag";
+import { InboxCardSourceMeta } from "@features/inbox/components/InboxCardSourceMeta";
+import { InboxCardTitle } from "@features/inbox/components/InboxCardTitle";
+import { SuggestedReviewerAvatarStack } from "@features/inbox/components/SuggestedReviewerAvatarStack";
```
oliverb123: I dislike the the variance in corner radius here, particularly this close together and not nested, and generally I like the idea of using icons here but making them look a bit like buttons is ugly imo

Image

packages/ui/src/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+  parseConventionalCommitTitle,
+} from "@features/inbox/utils/parseConventionalCommitTitle";
+import { extractRepoSelectionRepository } from "@features/inbox/utils/repoSelectionArtefact";
+import { LightningIcon, ThumbsDownIcon } from "@phosphor-icons/react";
+import { Button } from "@posthog/quill";
+import { Flex, Text } from "@radix-ui/themes";
+import type { SignalReport } from "@shared/types";
+import { Link, useNavigate } from "@tanstack/react-router";
+import { PriorityMonogram } from "./PriorityMonogram";
+
```
oliverb123, line 26: I really want a link to the relevant commit here, or at least it's short hash so I can go look at it.

Image

packages/ui/src/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+interface ReportCardProps {
+  report: SignalReport;
+  onDismiss: () => void;
+  dismissDisabledReason?: string | null;
+  isDismissPending?: boolean;
+}
+
+export function ReportCard({
+  report,
```
oliverb123, line 37: I see the open in GitHub button but no way to view the associated tasks? Seems like either the pulling of associated tasks is broken, or this is missing?

Image

packages/ui/src/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+}: ReportCardProps) {
+  const detailRoute = {
+    to: "/code/inbox/reports/$reportId" as const,
+    params: { reportId: report.id },
+  };
+  const { prefetch, pointerHandlers } = useInboxReportDetailPrefetch(
+    report,
+    detailRoute,
+  );
+  const navigate = useNavigate();
```
oliverb123, line 52: Feels wider than it needs to be, and puts the evidence and artifacts at the bottom. Particularly once we/I move towards "artifacts as log", we should make this two-column (dynamic to single column when the screen is too narrow) imo, otherwise your burying all the rich data (diffs, tasks, artefacts) below this wall of text

Image

packages/ui/src/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const { data: artefactsResp } = useInboxReportArtefacts(report.id, {
+    staleTime: 5 * 60 * 1000,
+    refetchOnWindowFocus: false,
+  });
```
oliverb123, line 56: My impression is "runs" is only meant to show ongoing tasks? Seems like a bit of an arbitrary limitation imo, but given it, I'd call it "running" or "live" instead

Image

apps/code/src/renderer/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+    detailRoute,
+  );
+  const navigate = useNavigate();
+  const { data: artefactsResp } = useInboxReportArtefacts(report.id, {
+    staleTime: 5 * 60 * 1000,
+    refetchOnWindowFocus: false,
+  });
+  const repoSlug = extractRepoSelectionRepository(artefactsResp?.results);
+  const firstSource = report.source_products?.[0];
+  const sourceMeta = firstSource ? SOURCE_PRODUCT_META[firstSource] : null;
```
oliverb123: UI seems to be lying about there being only 100 items (what items? reports? runs?) "for me"

Image

Also this feels like a weird place to put this, I thought it would be a tab of curated reports or something, not a filtering UI

packages/ui/src/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const repoSlug = extractRepoSelectionRepository(artefactsResp?.results);
+  const firstSource = report.source_products?.[0];
+  const sourceMeta = firstSource ? SOURCE_PRODUCT_META[firstSource] : null;
+  const updatedAtLabel = new Date(report.updated_at).toLocaleDateString(
+    undefined,
+    { month: "short", day: "numeric" },
+  );
+  const isReady = report.status === "ready";
```
oliverb123, line 68: Spacing between the "findings" header and the first finding is bad, count is in a pill but the rest of the ui uses round rects, count pill has a BG color that matches the rest of the UI but icon doesn't

Also feels weird to call them "findings" here but "evidence" in the PR version, unless I'm confused about what this is displaying (it's just a single, right?)


Image

packages/ui/src/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const dismissTooltip = dismissReportTooltip(report);
+  const conventionalTitle = parseConventionalCommitTitle(report.title);
+  const cardTitle = displayConventionalCommitTitle(
+    report.title,
+    "Untitled report",
```
oliverb123, line 72: This header in general could be removed imo, but failing that, "**this is your inbox**" is really odd and should be removed

Image

packages/ui/src/features/inbox/components/ReportCard.tsx
```diff
@@ -0,0 +1,207 @@
...
+  );
+
+  const openDetail = () => {
+    prefetch();
```
oliverb123, line 76: No way to view suggested reviewers inside a report, even though we display them in the list view, and no way to modify suggested reviewers on a report OR a PR detail page (serious regression I'd actually block merging on)

Image

oliverb123 (commented):
Did the usual and just dump3e my ui feedback somewhere vaguely sensible, and let you route it properly to the relevant code.

Seems like the major things are the fetching tasks associated with a particular report is broken, and you deleted functionality around changing the suggested reviewers list. Also I think we could go 2 column in the detail view

apps/code/src/renderer/features/inbox/components/ReportCard.tsx
Twixes: Late addition to the PR with those icons, got messed up indeed!

packages/ui/src/features/inbox/components/ReportCard.tsx
Twixes: _Primarily_ for ongoing runs, but it makes sense to me to show the ongoing ones first, and finished ones lower down

packages/ui/src/features/inbox/components/ReportCard.tsx
Twixes: Agreed, should be no more than 100 chars wide really

So, I like "artifacts" as long. It's kind of a matter of responsive design. If we have the space, via a container query, we can have the description on the left, and the other details to the right. Slightly like a Facebook page UI 🤔

packages/ui/src/features/inbox/components/ReportCard.tsx
Twixes: I think the header feels like pretty solid grounding of what you're looked at, so leaving here, but cleaned up descriptions

packages/ui/src/features/inbox/components/ReportCard.tsx
Twixes: Missed the reviewer editing as it's newer, restored and unified with reports!

apps/code/src/renderer/features/inbox/components/ReportCard.tsx
Twixes: Did feel too tab-like. Made it a segmented button on the right:

Screenshot 2026-06-10 at 09 38 21@2x

Twixes:
* **#2567** <a href="https://app.graphite.com/github/pr/PostHog/code/2567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2547** <a href="https://app.graphite.com/github/pr/PostHog/code/2547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2547?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/code/src/renderer/features/inbox/components/ReportCard.tsx
Twixes: As for the UI lying about the count, you are correct. I'll leave it be for now though, as the fix requires a backend change to be able to query "PRs" and "Reports" separately (i.e. separate list request for reports with code vs. without)

packages/ui/src/features/inbox/components/ReportCard.tsx
Twixes: BTW the "This is your inbox" got lifted from the mocks I did https://posthog-self-driving.pages.dev/  lol

Twixes:
### Merge activity

* **Jun 10, 11:58 AM UTC**: @Twixes merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/2547).

PR #2564, opened 2026-06-10
feat(mobile): fetch available model list from the LLM gateway

oliverb123:
## Problem

On mobile, the available model list was hardcoded in `composer/options.ts`, so newer models (e.g. 4.8 / 5) never showed up without shipping a new app build. The desktop app already downloads the list from the LLM gateway, so the two clients drifted.

## Changes

- Moved the LLM gateway URL construction and the model fetch/format logic into `@posthog/shared` (zero-dependency, already a mobile dependency) as the single source of truth; `packages/agent` now re-exports from it so the desktop path is literally the same code.
- Mobile fetches the model list from the gateway's `/v1/models` endpoint via a new `useModels` hook backed by a persisted cache, mapping models exactly like the desktop app (Anthropic models, gateway-formatted names). A small built-in list remains only as a cold-start / offline fallback so the picker is never empty.

## How did you test this?

- `pnpm --filter @posthog/shared test` (incl. new gateway URL + model tests) — pass
- Agent gateway tests (`gateway-models`, `utils/gateway`) — pass
- `pnpm --filter agent typecheck` and agent build — pass
- Mobile `tsc` on touched files (clean), biome on changed files (clean), and `vitest run src/features/tasks` (84 tests) — pass

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1781052224119229?thread_ts=1781052224.119229&cid=C09G8Q32R6F)*


charlesvien:
cc: @Gilbert09

rafaeelaudibert:
Oh yeah, please, I really want this!

PR #2566, opened 2026-06-10
feat(agent): ship cloud agent logs to PostHog Logs via OTEL

pauldambra:
## Problem

The cloud `AgentServer` (the process that runs each task in a sandbox) shipped **no** telemetry to PostHog — its logs only went to stdout and S3. When something goes wrong with a cloud run (e.g. the recent Slack reports), there's no way to query what happened by `run_id` / `task_id`. The `OtelLogWriter` and `otelTransport` config existed but were never wired to anything.

This makes the desktop's existing pattern (logs → PostHog Logs via OTEL, indexed by service/run) work for the cloud agent too.

## Changes

- `OtelLogWriter.emitLog(level, scope, message, data)` — maps the agent log level to an OTEL severity and stores the scope as a `log.scope` attribute, mirroring the desktop electron-log OTEL transport. Adds `service.version` to the resource attributes and defaults the endpoint to `/i/v1/logs` (matches desktop + the existing `OtelTransportConfig`; the old `/i/v1/agent-logs` default was never used).
- `AgentServer` routes every server log line through one `onLog` choke point that emits to OTEL and still mirrors to the SSE console stream; flushes on fatal error and shuts down on stop.
- New optional `POSTHOG_OTEL_LOGS_HOST` / `POSTHOG_OTEL_LOGS_API_KEY` / `POSTHOG_OTEL_LOGS_PATH` env vars (plumbed via `bin.ts` → `AgentServerConfig.otelLogs`).

Export is **disabled (no-op) unless the sandbox injects the host + key** — same behavior as the desktop transport. This is deliberate: the sandbox only has the *customer's* API URL + personal key, so exporting there would dump customer task logs into customer projects (and a personal key may not even authorize log ingest). A follow-up on the infra side needs to inject a dedicated PostHog-internal logs host + ingest key to turn this on.

Scope note: this is logs-only. Errors land in PostHog **Logs** as `ERROR`-severity records; wiring exceptions into the dedicated Error Tracking product (posthog-node) is a deliberate follow-up.

## How did you test this?

- `pnpm --filter @posthog/agent typecheck` — passes.
- `pnpm --filter @posthog/agent test otel-log-writer` — 5/5 pass (added tests for `emitLog` severity/scope/body mapping and the `/i/v1/logs` default endpoint).
- `biome check` on all changed files — clean.
- Note: `agent-server.test.ts` has 55 pre-existing failures in this environment (msw `beforeAll` cascade) — identical on a clean `main` checkout, unrelated to this change.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1781077156745339?thread_ts=1781077156.745339&cid=C09G8Q32R6F)*


charlesvien:
cc: @tatoalo @jonmcwest

packages/agent/src/server/agent-server.ts
```diff
@@ -353,9 +355,39 @@ export class AgentServer {
...
+
   constructor(config: AgentServerConfig) {
     this.config = config;
-    this.logger = new Logger({ debug: true, prefix: "[AgentServer]" });
+    this.logger = new Logger({
+      debug: true,
+      prefix: "[AgentServer]",
+      // Pre-session logs go to OTEL only; the SSE console stream needs an
+      // active session, which the post-init logger wires in via handleLog.
+      onLog: this.emitOtelLog,
```
pauldambra, line 388: > [!NOTE]

**[convergent: qa-team/observability + paul + xp + codex-bot]** 🟠 HIGH

Wiring `onLog: this.emitOtelLog` onto the constructor logger silently drops every pre-session log line whenever OTEL is not configured. `Logger.emitLog` returns immediately after calling `onLog` and never falls through to `console.*` (`utils/logger.ts:38-49`), and per the PR body OTEL is disabled in every current sandbox until infra injects the env vars. So on ship day the whole pre-session phase (port-bind errors, repo-ready waits, JWT/auth failures, session-init crashes) stops going to stdout and goes to OTEL nowhere — the opposite of the intent for an observability PR.

Two knock-on effects from the same root cause:
- The `createOtelLogWriter()` catch at line 437 logs `"Failed to initialize OTEL log writer"` through this same logger, so an OTEL init failure is itself unobservable (console bypassed, `otelLogWriter` still null).
- The `EventIngest` child logger (line 406) captures `onLog` at creation and is never rewired at session init, so its logs stay OTEL-only for the whole process lifetime.

Recommendation: tee rather than replace — have the pre-session sink write to console **and** OTEL (or give `Logger` an additive/tee option), so we never end up with a strictly-worse sink than we started with. Don't switch off the old path before the new one is proven live.

packages/agent/src/otel-log-writer.ts
```diff
@@ -67,6 +94,31 @@ export class OtelLogWriter {
...
+   */
+  emitLog(
+    level: LogLevel,
+    scope: string,
+    message: string,
+    data?: unknown,
+  ): void {
+    const severity = LOG_LEVEL_TO_SEVERITY[level] ?? LOG_LEVEL_TO_SEVERITY.info;
+    const body =
+      data !== undefined ? `${message} ${safeStringify(data)}` : message;
```
pauldambra, line 112: > [!NOTE]

**[convergent: security-audit/sensitive-data-exposure + xp]** 🟡 MEDIUM

Routing all logger output through `onLog` also bypasses the `shouldLog` debug gate in `Logger.emitLog` (`utils/logger.ts:44`), and `emitLog` here stringifies caller `data` into the body with no redaction. Net effect: debug-level lines the console path would have suppressed are now exported to PostHog Logs unconditionally — including the end-user prompt preview at `agent-server.ts:911` (`promptPreview.substring(0,100)`), repo/branch/path metadata, and (LOW, gated by `emitRawSDKMessages`) extension-notification payloads at `agent-server.ts:3031`.

Per the PR body the destination is a PostHog-internal, opt-in logs project, which caps this at internal-operator exposure of customer content rather than cross-tenant leakage — but the missing severity threshold and redaction on the export path is worth closing before this is enabled. Recommendation: gate export by severity (drop debug, or make it configurable) and/or redact free-form prompt/content previews before export.

packages/agent/src/otel-log-writer.ts
```diff
@@ -42,7 +66,7 @@ export class OtelLogWriter {
     sessionContext: SessionContext,
     _debugLogger?: Logger,
   ) {
-    const logsPath = config.logsPath ?? "/i/v1/agent-logs";
+    const logsPath = config.logsPath ?? "/i/v1/logs";
```
pauldambra, line 69: > [!NOTE]

**[qa-team/observability]** 🟡 MEDIUM

Once OTEL is enabled its own failures are invisible. `OTLPLogExporter`/`BatchLogRecordProcessor` report errors through OpenTelemetry's global error handler and `diag`, which are no-ops unless `diag.setLogger(...)` is called, and the batch processor silently drops records past `maxQueueSize`. A wrong `POSTHOG_OTEL_LOGS_API_KEY` (or malformed host/path) means every cloud run exports nothing with zero signal anywhere. The `_debugLogger` constructor param is right there but unused — wire it (or stderr) into a diag logger / export-error hook so the first failed export is visible in sandbox stdout, otherwise the infra follow-up that turns this on can't verify it works.

Related: the URL is built by naive concat (`${config.posthogHost}${logsPath}`, line 71). A trailing slash on `POSTHOG_OTEL_LOGS_HOST` (accepted by `z.url()`) yields `//i/v1/logs`, and a `POSTHOG_OTEL_LOGS_PATH` without a leading slash yields `hosti/v1/logs` — either silently ships nothing. Normalize at construction or tighten the env schema.

pauldambra (commented):
QA Swarm multi-perspective review complete. See inline comments and the summary below.

pauldambra:
> [!NOTE]
>

The design is sound — a null-object OTEL writer gated on injected env vars, a single log choke point, flush-on-fatal racing a deadline, clean shutdown, and nicely parameterized tests. But all four reviewers independently landed on the same behavioural regression, and the security lens found the same `onLog` routing has a data-handling edge. Worth resolving the observability-of-its-own-failure-modes before this ships.

### Key findings

**HIGH**
- **Pre-session logs silently dropped when OTEL is off** (`agent-server.ts:388`). `Logger.emitLog` treats `onLog` as a *replacement* for console, not a tee (`utils/logger.ts:38-49`), and OTEL is a no-op in every current sandbox. So pre-session startup errors go to stdout no more and to OTEL nowhere. Same root cause makes the OTEL-init-failure catch (line 437) unobservable and routes the `EventIngest` child logger (line 406) OTEL-only for the whole process lifetime.

**MEDIUM**
- **No severity threshold / redaction on the export path** (`otel-log-writer.ts:112`). `onLog` routing bypasses the debug gate, so debug-level prompt previews (`agent-server.ts:911`) and repo/path metadata ship to PostHog Logs. Internal opt-in destination caps severity, but redaction/threshold is worth adding before enable.
- **OTLP export failures are invisible** (`otel-log-writer.ts:69-77`). No `diag` logger, unused `_debugLogger`, silent drops past `maxQueueSize`; naive URL concat means a trailing/leading-slash misconfig ships nothing silently.
- **Possibly-orphaned `emit()` event path** (`otel-log-writer.ts:125`). README moves `SessionLogWriter` to S3 + local cache and pulls OTEL out of it; the notification-emit path now looks called only by its tests. Confirm it's still wired or drop it.
- **Missing seam-level test** for the safety-critical no-op-when-unconfigured path — assert `createOtelLogWriter()` returns null with no config, logging still works, and shutdown/flush don't throw on the null writer.

**LOW / NIT**
- Fatal-error flush shares bin.ts's single 5s deadline with two prior network ops; order the flush first or give it its own sub-timeout (`agent-server.ts:841`).
- `stop()` awaits `otelLogWriter.shutdown()` with no deadline — a blackholed OTLP endpoint can block shutdown up to 30s until SIGKILL (`agent-server.ts:792`).
- Dead fallback branch `LOG_LEVEL_TO_SEVERITY[level] ?? …info` — unreachable given the `Record<LogLevel,…>` type (`otel-log-writer.ts:110`).
- Body-formatting logic duplicated 3x with inconsistent stringify safety; `emitConsoleLog`/`emit()` use raw `JSON.stringify` and will throw on a circular payload — reuse `safeStringify`.
- `emitOtelLog` vs `handleLog` may collapse to one choke point (`emitConsoleLog` already no-ops pre-session).
- Brittle `mockExport.mock.calls[0][0][0]` repeated across tests — extract a helper.

### Convergence

- **Pre-session logging regression** — flagged independently by qa-team, paul, xp, **and** the existing `chatgpt-codex-connector` bot thread. Four-plus-way convergence; highest confidence.
- **Swallowed OTEL-init-failure error** — qa-team + paul + xp.
- **Raw-stringify / body-format duplication** — paul + xp.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Sound design; a family of fail-silent regressions around the pre-session logger and invisible OTLP export failures need tightening. |
| 👤 paul | Right intent; rollout-order regression drops stdout for a sink that's a no-op until infra lands — keep stdout alive alongside OTEL. |
| 📐 xp | Good parameterized tests; the behavioural drop of stdout (not the style nits) is the thing to fix before merge. |
| 🛡 security-audit | No key-exfil/SSRF risk (host+key both injected by trusted orchestrator); real issue is unredacted debug content shipped to Logs via the bypassed debug gate. |

---

packages/agent/src/otel-log-writer.ts
pauldambra: > [!NOTE]

Deferring — this needs a human call, not an auto-fix.

Confirmed the finding is accurate: routing everything through `onLog` means every log call (including `debug`-level ones, since these loggers are all constructed with `debug: true`) now gets shipped off-box via OTEL, including free-form content like the prompt preview at `agent-server.ts:911`. Previously that content only ever reached local stdout.

There are at least two reasonable ways to close this (drop debug-level export entirely vs. make the threshold configurable) and a separate, non-trivial question of what/how to redact free-form content — those are product/security trade-offs for a human to make, not something to silently pick on a bot's behalf. Leaving this open for a decision on severity gating and/or redaction before OTEL export is enabled for real traffic.

PR #2580, opened 2026-06-10
test(di): guard against duplicate Symbol.for DI token strings

jonathanlab:
Adds a regression test that guards against the `Symbol.for` collision class such as in #2579 

---


tatoalo (approved):
yes

PR #2616, opened 2026-06-11
fix(inbox): refetch + virtualize + server-search the people pickers

Twixes:
## Problem

The inbox **assignee / "Entire project"** scope selector broke: it didn't re-fetch, so the people list went stale (reported by Olly/Rafa in Slack). While in there, the PostHog/code people listings in both **Reviewers** and **Entire project** also needed: a virtualized list, server-side search, and a helpful empty state.

## Changes

- New shared `PeoplePickerList` — a searchable, **virtualized** people list (only visible rows mount, so it stays responsive for large orgs) with the empty-state hint.
- New shared `useReviewerPickerOptions` hook driving **server-side search** through the `available_reviewers` `?query=` param (debounced), and pinning "me" only on the unfiltered base list.
- **Reviewers** picker (`SuggestedReviewersSection`) and **inbox scope** picker (`InboxScopeSelect`) both rewired onto the shared list.
- **Re-fetch bug fix:** the scope dropdown's data now lives inside a Popover whose content only mounts while open, so the query re-mounts and re-fetches on every open instead of staying mounted and stale.
- **Empty state:** when no people match, the dropdown shows _"Someone not showing up here? Ask them to connect their GitHub profile with PostHog"_, linking to the personal settings page (`/settings/user`).

> Note on "fetch missing entries as you scroll": the `available_reviewers` endpoint returns the full set of people in one response (a UUID→person dict, no `count`/pagination), so there are no missing pages to fetch — virtualization covers the rendering cost. True incremental fetch-on-scroll would need a paginated backend endpoint (in the `posthog` Django repo) first; the shared list is structured so it can be extended to infinite scroll if/when that lands.

## How did you test this?

- `pnpm build` (all packages) — green.
- `pnpm --filter @posthog/ui typecheck` — green.
- `biome check` on the changed/added files — clean.
- No existing tests reference these components; mobile uses its own separate inbox flow and is unaffected.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09SK2PAGKF/p1781192703338229?thread_ts=1781192703.338229&cid=C09SK2PAGKF)*

### DIFF
--- a/packages/ui/src/features/inbox/components/InboxScopeSelect.tsx
+++ b/packages/ui/src/features/inbox/components/InboxScopeSelect.tsx
@@ -1,56 +1,63 @@
-import { AsteriskSimpleIcon, CaretDownIcon } from "@phosphor-icons/react";
+import {
+  AsteriskSimpleIcon,
+  CaretDownIcon,
+  CheckIcon,
+} from "@phosphor-icons/react";
 import {
   INBOX_SCOPE_ENTIRE_PROJECT,
   INBOX_SCOPE_FOR_YOU,
-  type InboxScope,
-  isTeammateInboxScope,
   parseTeammateInboxScope,
   teammateInboxScope,
 } from "@posthog/core/inbox/reportMembership";
-import {
-  Combobox,
-  ComboboxContent,
-  ComboboxEmpty,
-  ComboboxInput,
-  ComboboxItem,
-  ComboboxList,
-} from "@posthog/quill";
+import { PeoplePickerList } from "@posthog/ui/features/inbox/components/PeoplePickerList";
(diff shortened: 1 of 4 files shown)


Twixes:
Addressed the greptile note: removed the unused `hasResults` field from `ReviewerPickerOptions` and its computation in `useReviewerPickerOptions`. Verified via repo-wide grep that neither caller (`ScopePickerContent` / `InboxScopeSelect`, `SuggestedReviewersBody` / `SuggestedReviewersSection`) nor any test/re-export consumed it. Package typecheck, tests (697 passing), and biome all green locally.

PR #2617, opened 2026-06-11
fix(inbox): label report tasks with the report title

Twixes:
## Problem

Tasks created from an inbox report (the **Create PR** / "Act on" action, and **Discuss**) showed dumb generic titles like *"Act on PostHog inbox report"* and ignored the report's own title.

The title is produced by LLM title *generators* that summarize the create-PR/discuss prompt. That prompt (`buildCreatePrReportPrompt`) is intentionally generic — it points the agent at the inbox MCP tools rather than inlining the report — so the best title a generator can squeeze out of it is literally "Act on PostHog inbox report". Two generators are involved:

- **Initial title:** PostHog/posthog's `generate_task_title()` (`TaskSerializer.create`) when the API receives no explicit title.
- **Overwrite on open:** this repo's `TitleGeneratorService` via `useChatTitleGenerator`. The inbox flow calls `openTask()` after creating, mounting the hook, which on the agent's first prompt re-summarizes that generic prompt and `updateTask({ title })`.

## Changes

Make report-scoped tasks carry the report's own title, and stop the generators from rewriting it:

- Plumb an optional `title` through `TaskCreationInput` → `TaskCreationSaga` → `createTask`, and have the Create-PR / Discuss inbox hooks (and `SignalReportTaskService`) set it to the report title. The backend already honors an explicit title; non-report tasks send none and keep auto-generating, so they're unaffected.
- `decideTitleGeneration` now skips auto-title generation entirely for `signal_report`-origin tasks, so `useChatTitleGenerator` can no longer clobber the seeded report title by summarizing the generic preamble.

## How did you test this?

- `vitest run` for the affected `@posthog/core` and `@posthog/ui` files — added coverage that the report title is forwarded as the task title, omitted when the report has none, and that auto-title generation never runs for report-scoped tasks (unit + hook level). All pass.
- `turbo typecheck` for `@posthog/shared`, `@posthog/core`, `@posthog/ui` — pass.
- `biome check` on the changed files — clean.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09SK2PAGKF/p1781196324488669?thread_ts=1781196324.488669&cid=C09SK2PAGKF)*

### DIFF
--- a/packages/core/src/inbox/reportTaskCreation.ts
+++ b/packages/core/src/inbox/reportTaskCreation.ts
@@ -28,6 +28,7 @@ export function selectModelFromOptions(
 export interface BuildSignalReportTaskInput {
   prompt: string;
   reportId: string;
+  reportTitle?: string | null;
   cloudRepository: string;
   githubUserIntegrationId: string;
   adapter: "claude" | "codex";
@@ -43,6 +44,7 @@ export function buildSignalReportTaskInput(
   const {
     prompt,
     reportId,
+    reportTitle,
     cloudRepository,
     githubUserIntegrationId,
     adapter,
@@ -53,6 +55,8 @@ export function buildSignalReportTaskInput(
   return {
     content: prompt,
     taskDescription: prompt,
+    // Label the task with the report title; the prompt itself is generic.
+    title: reportTitle ?? undefined,
     repository: cloudRepository,
     githubUserIntegrationId,
     workspaceMode: "cloud",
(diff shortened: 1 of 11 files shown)


Twixes:
Superseded by PostHog/posthog#63083 — the fix belongs in the backend. `TaskSerializer.create` already has the linked `SignalReport` (and its title) at creation, so it sets the task title from the report there instead of threading the report title through the desktop client. Single authoritative fix that works for every client.

PR #2626, opened 2026-06-12
fix(agent): drop rawOutput for non-MCP tool results

K3N4Y:
Standard tool results (Read, Bash, ...) render from `content`, so also copying the full result into `rawOutput` duplicated large payloads on their way to the renderer. Only MCP tools consume rawOutput (App bridge replay), so emit it solely when an MCP tool result is present.

## Problem

When PostHog Code runs standard tools like Read or Bash, the tool result payload was being duplicated: once in content (where the renderer actually reads it) and again in rawOutput. For large file reads or substantial bash output, this doubled the serialized payload sent through the renderer pipeline for no reason. Only MCP tools consume rawOutput — the App bridge replays it to display structured MCP results.

## Changes

sdk-to-acp.ts: Conditionally include rawOutput in tool_call_update only when an MCP tool result is present (ctx.mcpToolUseResult). Standard tools now emit status and metadata without the redundant rawOutput payload.
sdk-to-acp.test.ts: Added two tests covering both branches:
- Standard tools (e.g. Read) omit rawOutput entirely
- MCP tools (e.g. mcp__srv__do) forward rawOutput with content, structuredContent, and isError
This is a pure optimization — no behavioral change for consumers.

## How did you test this?

Ran pnpm --filter @posthog/agent test — all 690 tests pass, including the 2 new test cases
Ran pnpm biome lint on both changed files — 0 issues
Manually verified the MCP path: rawOutput still carries content, structuredContent, and isError for App bridge replay

## Automatic notifications

- [ ] Publish to changelog?


charlesvien:
is there anyway to prove this is a real win?

charlesvien:
This is an issue in the upstream SDK https://github.com/zed-industries/claude-agent-acp/pull/409

PR #2640, opened 2026-06-12
fix(sidebar): show spinner while archiving a task

pauldambra:
phode couldn't really cope if you archived task b while task a was still archiving

well, not any more

CleanShot 2026-06-12 at 20 17 09

## Problem

Archiving a task from the sidebar was unreliable. Clicking *Archive* fired the request and immediately removed the row (optimistic), so there was no feedback that anything was happening, failures were swallowed silently, and clicking archive/pin/right-click repeatedly during the in-flight window produced inconsistent results ("archiving doesn't always seem to take").

## Changes

- The conversation you click *Archive* on now shows a **spinner** in its row and stays put until the archive actually **succeeds**, then disappears.
- While a row is archiving, **clicking it, toggling its pin, and right-clicking are no-ops**, and repeated archive clicks are ignored.
- A failed archive now surfaces an error toast instead of failing silently.

Implemented via a small `archivingTasksStore` (UI view state) tracking in-flight ids, plus an `optimistic` flag on the `archiveTask` orchestration so the interactive single-archive path defers its cache removal to success (bulk/`archive-prior` keep the instant optimistic behavior).

## How did you test this?

- `packages/core` unit tests for the orchestration's deferred (`optimistic: false`) path — defers cache writes until archive resolves, and leaves caches untouched on failure.
- New unit tests for `archivingTasksStore`.
- `pnpm --filter @posthog/core --filter @posthog/ui typecheck` and Biome lint on the changed files, both clean.

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ACRAMJUAG/p1781269709959239?thread_ts=1781269709.959239&cid=C0ACRAMJUAG)*

### DIFF
--- a/packages/core/src/archive/archiveOrchestration.test.ts
+++ b/packages/core/src/archive/archiveOrchestration.test.ts
@@ -64,6 +64,35 @@ describe("archiveTask", () => {
     expect(harness.list.some((a) => a.taskId === TASK_ID)).toBe(true);
   });
 
+  it("with optimistic:false, defers cache writes until archive resolves", async () => {
+    let idsWhenArchiveCalled: string[] = ["sentinel"];
+    harness.deps.archive = vi.fn().mockImplementation(async () => {
+      // Snapshot the cache at the moment the request is made — the row must
+      // still be present (not yet marked archived) while it's in flight.
+      idsWhenArchiveCalled = [...harness.ids];
+    });
+
+    await archiveTask(TASK_ID, harness.deps, { optimistic: false });
+
+    expect(idsWhenArchiveCalled).not.toContain(TASK_ID);
+    // Once the archive resolves, the row is removed from the list.
+    expect(harness.ids).toContain(TASK_ID);
+    expect(harness.list.some((a) => a.taskId === TASK_ID)).toBe(true);
+  });
+
+  it("with optimistic:false, leaves caches untouched when archive fails", async () => {
+    harness.deps.getPinnedTaskIds = vi.fn().mockResolvedValue([TASK_ID]);
+    harness.deps.archive = vi.fn().mockRejectedValue(new Error("boom"));
+
+    await expect(
+      archiveTask(TASK_ID, harness.deps, { optimistic: false }),
+    ).rejects.toThrow("boom");
+
+    expect(harness.ids).not.toContain(TASK_ID);
+    expect(harness.list).toEqual([]);
+    expect(harness.deps.togglePin).toHaveBeenCalledWith(TASK_ID);
+  });
+
   it("rolls back caches and re-pins when archive fails", async () => {
     harness.deps.getPinnedTaskIds = vi.fn().mockResolvedValue([TASK_ID]);
     harness.deps.archive = vi.fn().mockRejectedValue(new Error("boom"));
--- a/packages/ui/src/features/sidebar/archivingTasksStore.ts
+++ b/packages/ui/src/features/sidebar/archivingTasksStore.ts
@@ -1,0 +1,43 @@
+import { create } from "zustand";
+
+interface ArchivingTasksState {
+  /** Task IDs with an archive request currently in flight. */
+  archivingTaskIds: Set<string>;
+}
+
+interface ArchivingTasksActions {
+  isArchiving: (taskId: string) => boolean;
+  startArchiving: (taskId: string) => void;
+  stopArchiving: (taskId: string) => void;
+}
+
+type ArchivingTasksStore = ArchivingTasksState & ArchivingTasksActions;
+
+/**
+ * Tracks which tasks are mid-archive so the sidebar can show a per-row spinner
+ * and treat clicks, pin toggles, and right-clicks on those rows as no-ops until
+ * the archive resolves.
+ */
+export const useArchivingTasksStore = create<ArchivingTasksStore>(
+  (set, get) => ({
+    archivingTaskIds: new Set(),
+
+    isArchiving: (taskId) => get().archivingTaskIds.has(taskId),
+
+    startArchiving: (taskId) =>
+      set((state) => {
+        if (state.archivingTaskIds.has(taskId)) return state;
+        const next = new Set(state.archivingTaskIds);
+        next.add(taskId);
+        return { archivingTaskIds: next };
+      }),
+
+    stopArchiving: (taskId) =>
+      set((state) => {
+        if (!state.archivingTaskIds.has(taskId)) return state;
+        const next = new Set(state.archivingTaskIds);
+        next.delete(taskId);
+        return { archivingTaskIds: next };
+      }),
+  }),
+);
--- a/packages/ui/src/features/sidebar/components/SidebarMenu.tsx
+++ b/packages/ui/src/features/sidebar/components/SidebarMenu.tsx
@@ -12,6 +12,7 @@ import {
   useArchiveCacheKeys,
   useArchiveTask,
 } from "@posthog/ui/features/archive/useArchiveTask";
+import { useArchivingTasksStore } from "@posthog/ui/features/sidebar/archivingTasksStore";
 import { useCommandCenterStore } from "@posthog/ui/features/command-center/commandCenterStore";
 import { useFeatureFlag } from "@posthog/ui/features/feature-flags/useFeatureFlag";
 import { useInboxReports } from "@posthog/ui/features/inbox/hooks/useInboxReports";
@@ -232,6 +233,11 @@ function SidebarMenuComponent() {
   }, [activeTaskId, selectedTaskIds]);
 
   const handleTaskClick = (taskId: string, e: React.MouseEvent) => {
+    // Ignore clicks on a row that's mid-archive.
+    if (useArchivingTasksStore.getState().isArchiving(taskId)) {
+      e.preventDefault();
+      return;
+    }
     if (e.shiftKey) {
       e.preventDefault();
       selectRange(taskId, orderedVisibleTaskIds, activeTaskId);
@@ -292,6 +298,12 @@ function SidebarMenuComponent() {
     e: React.MouseEvent,
     isPinned: boolean,
   ) => {
+    // Right-clicking a row that's mid-archive is a no-op.
+    if (useArchivingTasksStore.getState().isArchiving(taskId)) {
+      e.preventDefault();
+      return;
+    }
+
     // Bulk menu when 2+ tasks are in the effective selection (active + cmd/shift-clicked)
     // and the right-clicked task is one of them. Otherwise clear and fall through.
     if (effectiveBulkIds.length > 1) {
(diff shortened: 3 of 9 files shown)


pauldambra (commented):
> [!NOTE]

QA Swarm inline review — see comments below. Reviewers run: paul-reviewer, xp-reviewer, security-audit. (qa-team skipped — not present on this host.)

packages/ui/src/features/sidebar/components/SidebarMenu.tsx
```diff
@@ -337,24 +349,54 @@ function SidebarMenuComponent() {
...
+  // Guards against repeated clicks: a second call while archiving is a no-op.
+  const runArchive = useCallback(
+    (taskId: string) => {
+      const store = useArchivingTasksStore.getState();
+      if (store.isArchiving(taskId)) return;
+      store.startArchiving(taskId);
+      void archiveTask({ taskId })
+        .catch((error) => {
+          log.error("Failed to archive task", error);
+          toast.error("Failed to archive task");
```
pauldambra, line 363: > [!NOTE]

**[paul]** 🟡 MEDIUM

the whole motivation here is users reporting archives silently not landing, and we've fixed the UX (spinner + error toast) but we still can't *see* failures in the wild — we'll only know when someone complains again. could we capture an event on the failure path (and ideally success + a duration) so we can graph the archive failure rate? captain abundance of caution would want the number on a dashboard. not blocking, but i'd love this before we call "doesn't always take" solved.

packages/ui/src/features/sidebar/components/items/TaskItem.tsx
```diff
@@ -160,7 +166,7 @@ export function TaskItem({
     ) : null;
 
   const toolbar =
-    !hideHoverActions && (onArchive || onTogglePin) ? (
+    !isArchiving && !hideHoverActions && (onArchive || onTogglePin) ? (
```
pauldambra, line 169: > [!NOTE]

**[paul]** 🟢 LOW

no data-attr on the archive/pin hover buttons — sticking one on lets us find these in actions later and build a funnel for archive usage/failures. very cheap, very not blocking.

packages/ui/src/features/sidebar/components/SidebarMenu.tsx
```diff
@@ -232,6 +233,11 @@ function SidebarMenuComponent() {
   }, [activeTaskId, selectedTaskIds]);
 
   const handleTaskClick = (taskId: string, e: React.MouseEvent) => {
+    // Ignore clicks on a row that's mid-archive.
+    if (useArchivingTasksStore.getState().isArchiving(taskId)) {
```
pauldambra, line 237: > [!NOTE]

the `isArchiving(taskId)` early-return now lives in handleTaskClick, handleTaskContextMenu, handleTaskTogglePin and runArchive (plus handleTaskArchive) — ~five copies. faintly smells of shotgun surgery: add a sixth per-row interaction later and it's easy to forget the guard. that said, each call site has genuinely different early-return semantics (some preventDefault, some plain return), and the runArchive guard is the one that matters for correctness (dedupe) — the rest are UX. ThreeStrikesAndYouRefactor territory: fine to ship, revisit if a sixth shows up.

packages/core/src/archive/archiveOrchestration.test.ts
```diff
@@ -64,6 +64,35 @@ describe("archiveTask", () => {
...
+    harness.deps.archive = vi.fn().mockImplementation(async () => {
+      // Snapshot the cache at the moment the request is made — the row must
+      // still be present (not yet marked archived) while it's in flight.
+      idsWhenArchiveCalled = [...harness.ids];
+    });
+
+    await archiveTask(TASK_ID, harness.deps, { optimistic: false });
+
+    expect(idsWhenArchiveCalled).not.toContain(TASK_ID);
+    // Once the archive resolves, the row is removed from the list.
```
pauldambra, line 78: > [!NOTE]

**[xp]** 🟢 LOW

the comment "the row is removed from the list" reads backwards — appending `TASK_ID` to the archived-ids/archive-list caches is what *marks it archived*, and the assertions just below confirm it's present in those caches, not removed. the real deferral check is the line above (`idsWhenArchiveCalled` snapshotted mid-flight). suggest fixing the comment so the next maintainer isn't misled; the post-resolution assertions also restate what the optimistic-path test already covers.

packages/ui/src/features/sidebar/archivingTasksStore.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    startArchiving: (taskId) =>
+      set((state) => {
+        if (state.archivingTaskIds.has(taskId)) return state;
+        const next = new Set(state.archivingTaskIds);
+        next.add(taskId);
+        return { archivingTaskIds: next };
+      }),
+
+    stopArchiving: (taskId) =>
```
pauldambra, line 35: > [!NOTE]

**[paul]** 🟢 LOW

store design is the good part — idempotent start/stop with stable-ref-on-noop is exactly right and the `toBe(setAfterStart)` test is lovely. one question: stopArchiving is called from a `.finally` so we're robust to archive errors, but is there a path where the component unmounts mid-archive and the id is stranded in the set forever? probably harmless (keyed by task id, the row's gone), just checking you thought about it.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — not present on this host)

Focused, well-shaped fix. Deferring the cache write to *after* the archive resolves (via the new `optimistic` flag) is a smaller, more honest change than removing-then-rolling-back, and the new `archivingTasksStore` is clean and correctly placed as UI view state. No correctness or security issues found — the only thing worth doing before merge is the misleading test comment; everything else is non-blocking.

### Key findings
- 🟡 **MEDIUM** — Observability: the PR exists because archives "don't always seem to take," but there's still no way to *measure* the failure rate (no capture event, no `data-attr`). Spinner + error toast help the human; a metric would let us graph it.
- 🟢 **LOW** — Misleading test comment in `archiveOrchestration.test.ts` ("the row is removed from the list" reads backwards); a couple of redundant tail assertions.
- 🟢 **LOW** — `isArchiving` guard duplicated across ~five handlers (ThreeStrikes territory; fine for now).
- 🟢 **LOW** — `data-attr` on the archive/pin buttons for future funnels.
- 🟢 **LOW** — Question: can a task id be stranded in the store if a component unmounts mid-archive? (Likely harmless.)
- ⚪ **NITS** — success toast (in `useArchiveTask`) vs error toast (in `runArchive`) split across two files; `handleTaskArchive`'s guard is redundant with `runArchive`'s.

### Convergence
- **paul + xp** independently flagged the repeated `isArchiving` guards across handlers — higher confidence, but both judged it fine to ship under ThreeStrikesAndYouRefactor.

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 👤 paul | Solid, high-trust fix; would love a failure metric before calling "doesn't always take" solved. |
| 📐 xp | The `optimistic` flag is justified (not a wrong-abstraction smell); only real fix is a backwards test comment. |
| 🛡 security-audit | 0 findings — frontend view-state change, no new trust-boundary crossings, server remains the authz boundary. |

---

packages/ui/src/features/sidebar/components/SidebarMenu.tsx
pauldambra: > [!NOTE]

Good call — adding telemetry for archive success/failure/duration is a valuable follow-up. This is out of scope for this fix which targets the UX gap (spinner + error toast), but tracking it as a separate task would be worthwhile before declaring the "doesn't always take" issue fully resolved.

packages/ui/src/features/sidebar/components/items/TaskItem.tsx
pauldambra: > [!NOTE]

Noted — adding `data-attr` attributes to the archive/pin hover buttons would make autocapture actions easier to build. Agreed this is low priority and out of scope for this fix; tracking as a follow-up.

packages/ui/src/features/sidebar/components/SidebarMenu.tsx
pauldambra: > [!NOTE]

Agreed with the "ThreeStrikesAndYouRefactor" assessment — the guards have genuinely different semantics at each call site (some preventDefault, some plain return) and `runArchive` remains the authoritative correctness guard. Fine to ship as-is; will revisit if a sixth guard site appears.

packages/core/src/archive/archiveOrchestration.test.ts
pauldambra: > [!NOTE]

Valid — the comment "the row is removed from the list" is misleading since the assertions verify presence in archived-ids/archive-list caches (marking as archived), not removal. This would be a simple comment fix; deferring as a follow-up cleanup rather than holding this PR.

packages/ui/src/features/sidebar/archivingTasksStore.ts
pauldambra: > [!NOTE]

Good question — yes, an unmount mid-archive would strand the task id in the set, but as noted: the id is keyed by task id, the row is gone, and `stopArchiving` would be called from the `.finally` on the in-flight promise regardless of whether the component is still mounted (the callback is captured at call time, not tied to the component lifecycle). So the id will be cleaned up when the request settles. No leak.

packages/ui/src/features/sidebar/components/SidebarMenu.tsx
```diff
@@ -337,24 +349,54 @@ function SidebarMenuComponent() {
...
+          useArchivingTasksStore.getState().stopArchiving(taskId);
+        });
+    },
+    [archiveTask],
+  );
+
   const handleTaskArchive = useCallback(
     (taskId: string) => {
+      if (useArchivingTasksStore.getState().isArchiving(taskId)) return;
       const task = allSidebarTasks.find((t) => t.id === taskId);
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Intentional — the guard in `handleTaskArchive` provides an early-exit before the `find` call on the task list, giving slightly different semantics from the deduplication guard inside `runArchive`. Both the convergent QA Swarm finding and this comment agree the duplication is minor and not worth refactoring at this point.

pauldambra:
* **#2640** <a href="https://app.graphite.com/github/pr/PostHog/code/2640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/ui/src/features/sidebar/components/items/TaskItem.tsx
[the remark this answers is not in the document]
pauldambra: 🤖 takeover-stale-pr: addressed in 753468a60 — the row's `draggable` is now `{!isArchiving}` so a mid-archive row can no longer be dragged into the command center, and the row is dimmed via `isDimmed` while the archive is in flight.

PR #2645, opened 2026-06-12
fix(storybook): resolve workspace .tsx imports and mount app providers

adboio:
## Problem

Storybook stories that import a `.tsx` from a workspace package (e.g. `@posthog/ui/features/archive/ArchivedTasksView`) failed to load with **"Failed to fetch dynamically imported module"**. The underlying Vite error was `Failed to resolve import`.

Root cause: Storybook resolved those bare specifiers through each workspace package's `"./*": ["./src/*.ts", "./src/*.tsx"]` exports map. Vite matches the first pattern (`.ts`); for a `.tsx` target that `.ts` candidate doesn't exist, so resolution fails instead of falling through. Any `.tsx` subpath import broke — `.ts` ones happened to work. The renderer never hits this because it aliases `@posthog/*` → `packages/*/src` and lets extension resolution find the `.tsx`; the Storybook config just lacked those aliases.

## Changes

- **`.storybook/main.ts`** — alias the `@posthog/*` workspace packages to source (reusing the renderer's `workspaceAliases`), so extension resolution finds `.tsx`. Also switch docgen to `react-docgen-typescript`: the default Babel `react-docgen` chokes on the legacy Inversify `@injectable()` decorators that core services pull in.
- **`vite.shared.mts`** — export `workspaceAliases` so the Storybook config can share it.
- **`.storybook/withAppProviders.tsx`** *(new)* — a global decorator that mounts the same provider stack the renderer boots: `QueryClient`, the host tRPC context, a DI service container (binds what stories need; hands out inert stubs for unbound services so deep trees don't throw, while keeping `isBound` truthful for `useServiceOptional`), and a minimal TanStack Router. Components using `useHostTRPC` / `useService` / `useRouterState` now render instead of throwing "must be used within a `<Provider>`".
- **`.storybook/preview.tsx`** — register the decorator globally.
- **`PromptInput.stories.tsx`** — drop the now-redundant (and incomplete — missing `HostTRPCProvider`) per-story provider boilerplate.
- **`ArchivedTasksView.stories.tsx`** — import `ArchivedTask` from the canonical `@posthog/shared` root instead of the unexported `/archive-domain` subpath.

## Verification

Drove every story in a headless browser: **all 133 stories render, 0 failures** (previously every `.tsx`-importing story 404'd, and CreatePrDialog / ConversationView / PromptInput / ToolCallBlock hit provider errors). Lint passes.

---

### DIFF
--- a/apps/code/.storybook/main.ts
+++ b/apps/code/.storybook/main.ts
@@ -5,6 +5,7 @@ import type { StorybookConfig } from "@storybook/react-vite";
 import tailwindcss from "@tailwindcss/vite";
 import react from "@vitejs/plugin-react";
 import { mergeConfig } from "vite";
+import { workspaceAliases } from "../vite.shared.mts";
 
 const __filename = fileURLToPath(import.meta.url);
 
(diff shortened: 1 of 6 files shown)


charlesvien (approved):

adboio:
### Merge activity

* **Jun 12, 4:40 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/2645).

PR #2594, opened 2026-06-11
fix(analytics): Harden analytics setup

charlesvien:
## Changes

1. Pass main-minted session id via additionalArguments and init PostHog synchronously
2. Bind AnalyticsBootContribution first so init runs before render and all effects
3. Gate boot-time analytics reset on bootstrapComplete to stop wiping the persisted distinct_id
4. Fire buffered flag listeners at init so cached flag values apply without waiting for /flags
5. Remove now-unused getSessionId tRPC procedure
6. Add tests for the reset gate and listener buffering

## How did you test this?

Manually

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

PR #2654, opened 2026-06-13
fix(agent): Surface real session-init errors and kill leaked CLIs

charlesvien:
## Problem

Connecting to an agent could fail with a generic **"Internal error"** that named no cause, retried every ~30s, and left orphaned `claude` subprocesses behind. The real failure — a 30s session-init timeout in `ClaudeAcpAgent.createSession` — was thrown as a plain `Error`, which the ACP layer collapses into a JSON-RPC `-32603 "Internal error"` (real text buried in `data.details`) before it reaches the logs or UI.

## Changes

1. Throw a descriptive `RequestError` (not a plain `Error`) at the three session-init timeout sites so the real message survives the ACP/tRPC boundary instead of becoming "Internal error"
2. Abort the timed-out query (`abortController.abort()` + `query.close()`) so the `claude` subprocess can't leak and pile up under the retry loop
3. Surface `data.details` in the host-side "Failed to create session" log so the exported log names the actual cause for any other masked error
4. Bound the gateway `/v1/models` fetches with `AbortSignal.timeout(10s)` so a stalled gateway can't stall session init

## How did you test this?

- `pnpm --filter @posthog/agent test` — 709 passing, including a new refresh-timeout test (asserts a `RequestError` is thrown and the timed-out query is closed) and a `fetchGatewayModels` timeout test
- `@posthog/agent` and `@posthog/workspace-server` typecheck — clean
- `biome lint` on the 5 changed files — clean

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

PR #2655, opened 2026-06-13
fix(auth): Auto-recover project-less session after re-auth

charlesvien:
## Problem

After the `OAUTH_SCOPE_VERSION` bump forced a re-authentication, a transient org/projects fetch failure (a network blip on a fresh re-auth) leaves the user **authenticated but with no project selected** (`currentProjectId = null`), and nothing recovers it. That single state silently cascades:

- cloud sends fail with "Authentication required for cloud commands"
- the Cloud workspace mode disappears (the `twig-cloud-mode-toggle` flag has no posthog `project` group to evaluate against)
- the billing/usage bar goes blank
- the task list empties

Reported by Gustavo, who got stuck on local/worktree-only after re-auth.

## Changes

1. Distinguish **transient** (network/5xx) org-fetch failures from permanent (4xx) ones, and flag the session `orgProjectsIncomplete` only for the transient case
2. Add a bounded, deduped `refreshOrgProjects()` that re-fetches orgs/projects and re-selects the project via the existing `pickInitialProjectId`
3. Trigger recovery after an incomplete sync and from `attemptSessionRecovery` (connectivity-online + power-resume handlers)
4. Preserve the last-selected project across a project-less sync so recovery restores the **exact** project the user had
5. Tests for recover-and-restore on reconnect, recover on resume, and the empty-`scoped_organizations` no-retry path

Once `currentProjectId` is restored, the rest cascades back automatically (cloud sends, the cloud-mode flag via the re-set posthog `project` group, the seat/usage bar, and the task list), so no UI changes are needed.

Related (adjacent angles of the same incident): #2594 (analytics feature-flag race hiding the cloud option) and #2647 (explicit OAuth scopes instead of `*`).

## How did you test this?

- `pnpm --filter @posthog/core test` — 31 auth tests pass (3 new recovery tests + existing suite, including the existing 4xx no-retry case)
- `pnpm --filter @posthog/core typecheck` — clean
- `biome lint packages/core/src/auth/*` — clean

New unit tests fake a transient `/api/organizations/{id}/` failure during re-auth and assert the session becomes authenticated-but-project-less, then recovers and restores the previous project on reconnect and on resume; and that an empty `scoped_organizations` does not trigger a retry loop. No manual app run.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/agent/src/adapters/claude/claude-agent.refresh.test.ts
+++ b/packages/agent/src/adapters/claude/claude-agent.refresh.test.ts
@@ -1,4 +1,7 @@
-import type { AgentSideConnection } from "@agentclientprotocol/sdk";
+import {
+  type AgentSideConnection,
+  RequestError,
+} from "@agentclientprotocol/sdk";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { POSTHOG_METHODS } from "../../acp-extensions";
 import { Pushable } from "../../utils/streams";
@@ -15,6 +18,7 @@ type SdkQueryHandle = {
   setMcpServers: ReturnType<typeof vi.fn>;
   supportedCommands: ReturnType<typeof vi.fn>;
   initializationResult: ReturnType<typeof vi.fn>;
+  close: ReturnType<typeof vi.fn>;
   [Symbol.asyncIterator]: () => AsyncIterator<never>;
 };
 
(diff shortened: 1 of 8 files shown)


charlesvien:
* **#2656** <a href="https://app.graphite.com/github/pr/PostHog/code/2656?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2655** <a href="https://app.graphite.com/github/pr/PostHog/code/2655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2655?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2654** <a href="https://app.graphite.com/github/pr/PostHog/code/2654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #2656, opened 2026-06-13
fix(analytics): Resolve feature flags before first render

charlesvien:
## Problem

The "Cloud" option in the workspace-mode picker intermittently vanished on launch and only returned after a hard refresh or restart. It is gated by the `twig-cloud-mode-toggle` feature flag, but the renderer deferred PostHog initialization behind an async `getSessionId` IPC round-trip. React mounted first, so `useFeatureFlag` read the flag as `false` (flags not loaded yet) and hid the option until init finished and flags arrived. Every `useFeatureFlag` consumer was exposed to the same race; the `|| import.meta.env.DEV` fallback masked it during local development.

Closes #2598

## Changes

1. Pass the main-owned session id to the renderer synchronously via `additionalArguments`
2. Expose it on `window.__posthogBootstrap` from the preload
3. Initialize PostHog before React mounts so flags resolve on the first render
4. Keep `AnalyticsBootContribution` as an idempotent async fallback when the bridge value is absent
5. Session stitching is preserved — the renderer still uses the main process session id

## How did you test this?

- `pnpm --filter code typecheck` (main + renderer tsconfigs) — pass
- `biome check` on the changed files — clean (also enforced by the pre-commit hook)
- `pnpm --filter code exec vitest run src/main/platform-adapters/posthog-analytics.test.ts` — 7/7 pass
- `pnpm --filter @posthog/ui exec vitest run src/shell/posthogAnalyticsImpl.test.ts` — 11/11 pass

Not yet run: an end-to-end check on a production build with `twig-cloud-mode-toggle` enabled — the flag path is dev-forced on, so the race only manifests in production builds. A genuinely cold flag cache still has a one-time load window until the network flags response arrives, which is expected first-load behavior.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/apps/code/src/main/posthog-session-arg.test.ts
+++ b/apps/code/src/main/posthog-session-arg.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from "vitest";
+import { parseSessionIdArg } from "./posthog-session-arg";
+
+describe("parseSessionIdArg", () => {
+  it("returns the value when the flag is present", () => {
+    expect(parseSessionIdArg(["--posthog-session-id=abc-123"])).toBe("abc-123");
+  });
+
+  it("finds the flag among other arguments", () => {
+    expect(
+      parseSessionIdArg(["--foo", "--posthog-session-id=xyz", "--bar"]),
+    ).toBe("xyz");
+  });
+
+  it("returns null when the flag is absent", () => {
+    expect(parseSessionIdArg(["--type=renderer", "--no-sandbox"])).toBeNull();
+  });
+
+  it("ignores arguments that only share the prefix", () => {
+    expect(parseSessionIdArg(["--posthog-session-id-extra=foo"])).toBeNull();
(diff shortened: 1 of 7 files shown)


charlesvien:
* **#2656** <a href="https://app.graphite.com/github/pr/PostHog/code/2656?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2656?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2655** <a href="https://app.graphite.com/github/pr/PostHog/code/2655?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2654** <a href="https://app.graphite.com/github/pr/PostHog/code/2654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #2666, opened 2026-06-14
fix(sessions): durably write-ahead the start prompt so it can't be lost

pauldambra:
## Problem

Starting a local task run can **lose the user's prompt**. The prompt only exists in renderer memory until the session has fully initialized and `session/prompt` has been delivered (a separate step *after* `agent.start`). If session init exceeds the 30s `SESSION_VALIDATION_TIMEOUT_MS` — common in large monorepos, where loading skills/commands blows the budget — or the app is reloaded/quit/crashes during the auto-retry window, the prompt is gone and the run is left stranded in `queued`.

This is the root cause behind "I lost my prompt while starting a task" reports in the monorepo: session init repeatedly times out and the only copy of the prompt is in-memory state tied to a session that never came up.

**Why:** the prompt is the one thing in the start-a-task flow a user can't cheaply reproduce, yet nothing durable owned it before delivery.

## Changes

Persist the prompt as a durable **write-ahead** record *before* any session/agent work, and clear it only once it has actually been delivered.

- New `PendingPromptStore` contract in `@posthog/core` and an `electronStorage`-backed zustand store in `@posthog/ui` (keyed by `taskId`, mirrors the existing `sessionConfigStore`).
- `createNewLocalSession` now: writes the prompt ahead before `createTaskRun`/`agent.start`; records the `taskRunId` once the run exists; **recovers** a saved prompt when a later connect arrives without one; clears the record after `sendPrompt` resolves.
- A persisted record means "this prompt is owed delivery" — the hook for a future server-side reconciler to re-drive orphaned `queued` runs.

Out of scope (follow-ups): bumping/adapting the 30s init timeout, and the server-side reconciler + startup sweep that extends the guarantee to the case where the device never returns.

## How did you test this?

- `pnpm --filter @posthog/core typecheck` and `pnpm --filter @posthog/ui typecheck` — clean.
- `biome check` on all touched files — clean.
- New unit tests for the store (`pendingPromptStore.test.ts`).
- Full suites pass: **@posthog/core 1504 tests**, **@posthog/ui 726 tests** (including both modified host suites).

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1781437731823459?thread_ts=1781437731.823459&cid=C09G8Q32R6F)*

### DIFF
--- a/packages/core/src/sessions/sessionService.ts
+++ b/packages/core/src/sessions/sessionService.ts
@@ -57,7 +57,7 @@ import {
   OFFLINE_SESSION_MESSAGE,
   routeLocalConnect,
 } from "./connectRouting";
-import type { PendingPromptStore } from "./pendingPrompt";
+import type { PendingPromptRecord, PendingPromptStore } from "./pendingPrompt";
 import {
   type PermissionSelectionPlan,
   planPermissionResponse,
@@ -1082,13 +1082,25 @@ export class SessionService {
     const effectiveModel = model ?? recovered?.model;
     const effectiveReasoningLevel = reasoningLevel ?? recovered?.reasoningLevel;
 
-    // Write-ahead the prompt BEFORE any session/agent work. This is the one
-    // step that makes losing it impossible: everything below (run creation,
-    // the 30s-bounded agent.start, prompt delivery) can throw or be killed by
-    // an app restart, and the prompt still survives on disk to be recovered on
-    // the next connect. Cleared only once it has actually been delivered.
+    if (recovered) {
+      // Surface when the safety net actually catches something, so we can tell
+      // it's working (and gauge how long prompts sit undelivered).
+      this.d.log.info("Recovered a written-ahead prompt for a task", {
+        taskId,
+        ageMs: Date.now() - recovered.createdAt,
+      });
+    }
+
+    // Write-ahead the prompt BEFORE any session/agent work. This is the step
+    // that makes losing it very unlikely: everything below (run creation, the
+    // 30s-bounded agent.start, prompt delivery) can throw or be killed by an
+    // app restart, and the prompt still survives to be recovered on the next
+    // connect. Cleared only once it has actually been delivered. (Persistence
+    // is async and best-effort, so a crash in the first moments of a cold
+    // start can still race it — hence "very unlikely", not "guaranteed".)
+    let pending: PendingPromptRecord | undefined;
     if (effectivePrompt?.length) {
-      this.d.pendingPrompts.save({
+      pending = {
         taskId,
         taskTitle,
         repoPath,
@@ -1098,7 +1110,8 @@ export class SessionService {
         model: effectiveModel,
         reasoningLevel: effectiveReasoningLevel,
         createdAt: recovered?.createdAt ?? Date.now(),
-      });
+      };
+      this.d.pendingPrompts.save(pending);
     }
 
     const { client } = auth;
@@ -1111,22 +1124,12 @@ export class SessionService {
       throw new Error("Failed to create task run. Please try again.");
     }
 
-    // Record the run id on the write-ahead entry so it points at the run the
-    // prompt is being delivered to (a server-side reconciler can use this to
-    // re-drive an orphaned run).
-    if (effectivePrompt?.length) {
-      this.d.pendingPrompts.save({
-        taskId,
-        taskTitle,
-        repoPath,
-        initialPrompt: effectivePrompt,
-        taskRunId: taskRun.id,
-        executionMode: effectiveExecutionMode,
-        adapter: effectiveAdapter,
-        model: effectiveModel,
-        reasoningLevel: effectiveReasoningLevel,
-        createdAt: recovered?.createdAt ?? Date.now(),
-      });
+    // Stamp the same write-ahead entry with the run id it's now being
+    // delivered to (a server-side reconciler can use this to re-drive an
+    // orphaned run).
+    if (pending) {
+      pending = { ...pending, taskRunId: taskRun.id };
+      this.d.pendingPrompts.save(pending);
     }
 
     const { customInstructions: startCustomInstructions } = this.d.settings;
--- a/packages/core/src/sessions/sessionService.ts
+++ b/packages/core/src/sessions/sessionService.ts
@@ -57,7 +57,7 @@ import {
   OFFLINE_SESSION_MESSAGE,
   routeLocalConnect,
 } from "./connectRouting";
-import type { PendingPromptRecord, PendingPromptStore } from "./pendingPrompt";
+import type { PendingPromptStore } from "./pendingPrompt";
 import {
   type PermissionSelectionPlan,
   planPermissionResponse,
@@ -828,6 +828,21 @@ export class SessionService {
             ),
           );
         }
+
+        // A run stranded before its prompt was ever delivered (e.g. the
+        // original session-init timed out) routes here on the next connect
+        // rather than to createNewLocalSession, since it already has a run id
+        // and log. Its prompt is still owed: if the resumed log carries no
+        // prompt yet but we have a write-ahead record, deliver it now.
+        const owed = this.d.pendingPrompts.get(taskId);
+        if (owed?.initialPrompt.length && !hasSessionPromptEvent(events)) {
+          this.localRepoPaths.set(taskId, repoPath);
+          this.d.log.info("Recovered a written-ahead prompt on resume", {
+            taskId,
+            ageMs: Date.now() - owed.createdAt,
+          });
+          await this.deliverWrittenAheadPrompt(taskId, owed.initialPrompt);
+        }
         return true;
       } else {
         this.d.log.warn("Reconnect returned null", { taskId, taskRunId });
@@ -1098,9 +1113,8 @@ export class SessionService {
     // connect. Cleared only once it has actually been delivered. (Persistence
     // is async and best-effort, so a crash in the first moments of a cold
     // start can still race it — hence "very unlikely", not "guaranteed".)
-    let pending: PendingPromptRecord | undefined;
     if (effectivePrompt?.length) {
-      pending = {
+      this.d.pendingPrompts.save({
         taskId,
         taskTitle,
         repoPath,
@@ -1110,8 +1124,7 @@ export class SessionService {
         model: effectiveModel,
         reasoningLevel: effectiveReasoningLevel,
         createdAt: recovered?.createdAt ?? Date.now(),
-      };
-      this.d.pendingPrompts.save(pending);
+      });
     }
 
     const { client } = auth;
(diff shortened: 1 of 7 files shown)


packages/core/src/sessions/sessionService.ts
```diff
@@ -1070,27 +1111,45 @@ export class SessionService {
...
+    // Record the run id on the write-ahead entry so it points at the run the
+    // prompt is being delivered to (a server-side reconciler can use this to
+    // re-drive an orphaned run).
+    if (effectivePrompt?.length) {
+      this.d.pendingPrompts.save({
+        taskId,
+        taskTitle,
+        repoPath,
+        initialPrompt: effectivePrompt,
+        taskRunId: taskRun.id,
```
pauldambra: > [!NOTE]

The prompt record is assembled twice — once before `createTaskRun`, once here just to add `taskRunId` — 9 near-identical fields each time. Real say-it-once duplication (not premature dedup; no behavioural flag). Build it once: `const pending: PendingPromptRecord = {…}`, `save(pending)`, then after the run exists `save({ ...pending, taskRunId: taskRun.id })`. Otherwise every future field has to be threaded into two spots or they silently drift.

packages/core/src/sessions/sessionService.ts
```diff
@@ -1060,6 +1066,41 @@ export class SessionService {
...
+    const effectivePrompt = initialPrompt?.length
+      ? initialPrompt
+      : recovered?.initialPrompt;
+    const effectiveExecutionMode = executionMode ?? recovered?.executionMode;
+    const effectiveAdapter = adapter ?? recovered?.adapter;
+    const effectiveModel = model ?? recovered?.model;
+    const effectiveReasoningLevel = reasoningLevel ?? recovered?.reasoningLevel;
+
+    // Write-ahead the prompt BEFORE any session/agent work. This is the one
+    // step that makes losing it impossible: everything below (run creation,
```
pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

silly question on the durability claim — "makes losing it impossible" / "on disk" oversells it a touch. zustand persist's `setItem` here is async + fire-and-forget (rendererStorage drops rejections, and drops `setItem` entirely while a key's first read is still pending). on a fast cold-start the save can update memory but never reach disk. acceptably narrow window, but could we soften "impossible" to "very unlikely to lose"? not blocking.

packages/core/src/sessions/sessionService.ts
```diff
@@ -1060,6 +1066,41 @@ export class SessionService {
...
+    // Recover a previously written-ahead prompt when this connect didn't
+    // carry one itself: a persisted record means an earlier attempt for this
+    // task saved a prompt and never delivered it (init timed out, or the app
+    // was restarted before delivery). Adopt its prompt and the settings the
+    // user originally chose, unless the caller overrode them.
+    const recovered = initialPrompt?.length
+      ? undefined
+      : this.d.pendingPrompts.get(taskId);
+    const effectivePrompt = initialPrompt?.length
+      ? initialPrompt
```
pauldambra, line 1093: > [!NOTE]

**[paul]** 🟡 MEDIUM

observability (my #2 thing) — the whole point of this store is the moment it catches a prompt that would otherwise be lost, and that moment is silent. when `recovered` is truthy could we `log.info` (and maybe `track`) a "recovered pending prompt" with taskId + age (`Date.now() - recovered.createdAt`)? otherwise we ship a safety net we can't tell is catching anything. not blocking.

packages/ui/src/features/sessions/sessionServiceHost.recovery.integration.test.ts
```diff
@@ -135,6 +135,20 @@ vi.mock(
   () => mockSessionConfigStore,
 );
 
+const mockPendingPromptStore = vi.hoisted(() => ({
+  pendingPromptStore: {
+    save: vi.fn(),
+    get: vi.fn(() => undefined),
```
pauldambra: > [!NOTE]

**[paul]** 🟡 MEDIUM

the store CRUD tests are lovely — but the behaviour this PR *exists for* isn't exercised: `get()` is mocked to always return undefined, so the recovery branch in `createNewLocalSession` never runs in any test. could we add one where `get()` returns a record and assert the session delivers that prompt then calls `remove`? the regression i'd most want pinned. high value, not blocking.

packages/ui/src/features/sessions/pendingPromptStore.ts
```diff
@@ -0,0 +1,64 @@
...
+      getPrompt: (taskId) => get().promptsByTaskId[taskId],
+
+      removePrompt: (taskId) =>
+        set((state) => {
+          if (!(taskId in state.promptsByTaskId)) return state;
+          const { [taskId]: _removed, ...rest } = state.promptsByTaskId;
+          return { promptsByTaskId: rest };
+        }),
+
+      listPrompts: () => Object.values(get().promptsByTaskId),
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW — YAGNI

`listPrompts()` and the `taskRunId` field have no production caller yet — they exist only for the "later" startup sweep / reconciler the comments describe (rule 4). Cheap to keep, but the duplicated second `save` exists *solely* to populate `taskRunId`, which nothing reads yet. If the reconciler isn't imminent, the second save could be dropped until then.

packages/ui/src/features/sessions/pendingPromptStore.ts
```diff
@@ -0,0 +1,64 @@
...
+        set((state) => ({
+          promptsByTaskId: {
+            ...state.promptsByTaskId,
+            [record.taskId]: record,
+          },
+        })),
+
+      getPrompt: (taskId) => get().promptsByTaskId[taskId],
+
+      removePrompt: (taskId) =>
```
pauldambra, line 37: > [!NOTE]

**[xp]** ⚪ NIT

Nice symmetry with `sessionConfigStore`. One divergence: `removePrompt` guards with `if (!(taskId in …)) return state` while the sibling's `removeConfigOptions` doesn't. The guard is the safer choice (avoids a needless re-render/persist) — consider backporting it, or confirm the divergence is intentional.

pauldambra (commented):
> [!NOTE]

QA Swarm multi-perspective review (paul-reviewer, xp-reviewer, security-audit). qa-team was unavailable and skipped. See inline comments + the summary comment.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team unavailable, skipped)

Solid, well-motivated change that slots cleanly into the existing `sessionConfigStore`/`adapterStore` shape. No correctness or security blockers — all findings are non-blocking improvements. Both voice reviewers independently said "ship as you see fit".

### Key findings
- 🟡 **MEDIUM — duplicated write-ahead `save` block** *(convergent: paul + xp)*: the `PendingPromptRecord` is assembled twice, differing only by `taskRunId`. Build it once and re-save with `{ ...pending, taskRunId }`.
- 🟡 **MEDIUM — recovery is silent**: no `log`/`track` when a prompt is actually recovered, so we can't tell the safety net is catching anything (or whether the 30s-timeout theory holds in the wild).
- 🟡 **MEDIUM — recovery path untested**: `get()` is mocked to `undefined` everywhere, so `createNewLocalSession`'s recovery branch never executes in a test. Add one where `get()` returns a record and assert deliver + `remove`.
- 🟢 **LOW — "impossible to lose" oversells it**: persist is async fire-and-forget and can be dropped pre-hydration; soften the wording.
- 🟢 **LOW — YAGNI**: `listPrompts()` / `taskRunId` have no consumer until the future reconciler.
- ⚪ **NIT**: `removePrompt` guard diverges from the sibling `removeConfigOptions`.

### Convergence
- The duplicated `save` block was flagged independently by **paul** and **xp** → highest-confidence finding.

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 👤 paul | approve with comments — add a recovery log line + an integration test, soften "impossible", dedupe the save block |
| 📐 xp | sound write-ahead design; main concern is the say-it-once duplication in an already-long method; minor YAGNI on `list()`/`taskRunId` |
| 🛡 security-audit | no findings — key is a server-generated task UUID (no cross-tenant adoption), persisted via the same encrypted `electronStorage` the existing config store already uses |

---

pauldambra:
> [!NOTE]
> 🤖 Automated triage by **PR Shepherd** — not written by a human

Actioned the qa-swarm findings in bc252db:

- ✅ **Dedupe the write-ahead `save`** (convergent paul + xp) — record built once, re-saved with `{ ...pending, taskRunId }`.
- ✅ **Recovery is observable** — `log.info("Recovered a written-ahead prompt…", { taskId, ageMs })` when the safety net catches a prompt.
- ✅ **Softened "impossible to lose"** → "very unlikely" in both the doc comment and the contract, noting persistence is async/best-effort.

Deferred (author's call):
- ⏭ **Integration test for the recovery branch** — worth pinning, but it means making the store mock return a record in the 950-line host harness; left for a focused follow-up rather than auto-applied.
- ⏭ **YAGNI on `listPrompts()`/`taskRunId`** and the **`removePrompt` guard divergence** — both are intentional groundwork for the server-side reconciler; keeping as-is.

Tests stay green: core 162 (sessions), ui host+store suites 122.

pauldambra:
> [!NOTE]
> 🤖 Automated triage by **PR Shepherd** — not written by a human

Triaged the Codex + Greptile findings (they converged with qa-swarm on the duplicate `save`, already fixed):

- ✅ **Codex P2 — clear only on real delivery** (ebc75e6): `sendPrompt` can resolve without delivering (`"queued"` when the session is busy, `"rate_limited"` when `sendLocalPrompt` swallows the usage-limit error). We now only `remove()` the write-ahead record when `stopReason` is neither — otherwise the prompt is kept. Good catch; this was a real (narrow) loss window I'd reintroduced.
- ✅ **Greptile P2 — duplicate save / drifting `createdAt`** (bc252db): record built once, so `Date.now()` is evaluated once too.
- ⏭ **Codex P1 — recover owed prompts on the `resume-existing` route** (deferred): correct and important — `routeLocalConnect` returns `resume-existing` whenever a run id + log URL exist, so a run stranded by a timed-out first attempt routes to `reconnectToLocalSession`, which doesn't consult `pendingPrompts`. The client-side recovery in this PR only covers the `create-new` route. Full coverage of the resume path belongs with the **server-side reconciler / startup-sweep follow-up** already called out in the PR description (it needs "resumed run that never received its prompt" detection + tests). Tracking it there rather than widening this PR.
- ⏭ **Greptile P2 / paul — integration test for the recovery branch** (deferred): same as noted above; follow-up.

Net: two real fixes applied, the deeper resume-path recovery consciously deferred to the reconciler follow-up. Re-pushed at ebc75e6.

pauldambra:
> [!NOTE]
> 🤖 Automated triage by **PR Shepherd** — not written by a human

Closed out the remaining review threads in 9ec4d33:

- ✅ **Codex P1 — recover owed prompts on the `resume-existing` route**: a run stranded before delivery routes to `reconnectToLocalSession` on the next connect, so it now delivers the owed write-ahead prompt there too — guarded on the resumed log carrying no prompt event yet. Pulled the deliver-then-clear-on-confirmed-delivery logic into a shared `deliverWrittenAheadPrompt()` so both routes share it.
- ✅ **xp YAGNI — `listPrompts()` / `taskRunId`**: dropped the speculative surface (and the second `save` that only existed to populate `taskRunId`) since the reconciler that would consume them isn't in this PR. Bonus: this removes the duplicated `save` entirely rather than just deduping it.
- ✅ **recovery now under test**: three e2e tests against the real store — delivery on create-new, delivery on resume-existing, and *no* delivery when the resumed run already has a prompt event.
- ⚪ **`removePrompt` guard divergence (NIT)**: intentional — the guard is the safer side; left as-is.

All review threads are now actioned or resolved-by-design. Local: core 162 + ui sessions 224 green; core + ui typecheck clean.

PR #2668, opened 2026-06-14
feat(otel): Add distributed tracing across app and workspace server

pauldambra:
## Problem

We lacked end-to-end visibility into requests flowing between the Electron main process and the workspace server. This made it difficult to diagnose latency and trace tRPC calls across process boundaries.

## Changes

- Added OpenTelemetry distributed tracing across the app and workspace server.
- Introduced a new `otel-trace.ts` utility in the main process exposing `initOtelTracing`, `getMainTracer`, and `shutdownOtelTracing`, backed by the workspace server's `node-tracing` module.
- Initialized OTEL tracing at app startup in `main/index.ts` and wired graceful shutdown of the trace exporter into the app lifecycle service.
- Added a tRPC tracing middleware (`tracingMonitor`) that wraps each call via `traceTrpcCall`, composed alongside the existing call-rate monitor on `publicProcedure`.
- Tracing is gated on PostHog configuration (`VITE_POSTHOG_API_KEY` / `VITE_POSTHOG_API_HOST`) and returns `null` when not configured.
- Added the required OpenTelemetry dependencies (`@opentelemetry/api`, `context-zone`, `exporter-trace-otlp-http`, `sdk-trace-base`, `sdk-trace-node`, `sdk-trace-web`).
- Added unit tests (`otel-trace.test.ts`) covering initialization gating, provider registration, and shutdown flush behavior.

## How did you test this?

- Added and ran unit tests in `apps/code/src/main/utils/otel-trace.test.ts` covering:
  - Returning `null` when the API key or host is missing/invalid.
  - Registering a provider and returning a tracer when configured.
  - Flushing and shutting down the provider on shutdown.
  - No-op shutdown when the provider was never created.

## Automatic notifications

- [ ] Publish to changelog?

---


packages/workspace-server/src/serve.ts
```diff
@@ -33,6 +36,7 @@ const shutdown = (reason: string) => {
   if (shuttingDown) return;
   shuttingDown = true;
   process.stdout.write(`[workspace-server] shutdown (${reason})\n`);
+  void shutdownOtelTracing();
```
pauldambra: > [!NOTE]

**[qa-team/reliability]** 🟠 HIGH

`shutdown()` calls `void shutdownOtelTracing()` (fire-and-forget) and then `process.exit(0)` synchronously on the next line when `server` is null. The flush (`forceFlush()` + `shutdown()`) is async and never completes before the process exits, so buffered spans — the `BatchSpanProcessor` holds them up to `scheduledDelayMillis: 2000` — are dropped on every clean shutdown. The non-null path (`server.close()` + 3s exit timer) doesn't await it either. Make `shutdown` await the flush before exiting, bounded by the grace timer — e.g. `shutdownOtelTracing().finally(() => process.exit(0))`. Compare the main process, which awaits `shutdownOtelTracing()` in app-lifecycle.

apps/code/src/renderer/utils/otel-trace.ts
```diff
@@ -0,0 +1,130 @@
...
+    observer.observe({ entryTypes: ["longtask"] });
+  } catch {
+    // longtask is not observable in every environment; ignore.
+  }
+}
+
+function flushOnUnload(): void {
+  if (typeof window === "undefined") return;
+  window.addEventListener("pagehide", () => {
+    void tracerProvider?.forceFlush();
```
pauldambra, line 132: > [!NOTE]

**[qa-team/reliability]** 🟡 MEDIUM

The renderer only flushes via a `pagehide` listener with a fire-and-forget `void tracerProvider?.forceFlush()`, and the provider is never `shutdown()`. In Electron, `pagehide`/`beforeunload` are unreliable on app quit and window close, and the forceFlush promise isn't awaited. With the 2s batch delay, the most recent renderer spans (`react.commit`, `route.navigate`, `browser.longtask`) will routinely be lost. Consider also flushing on `visibilitychange`→hidden.

packages/workspace-server/src/node-tracing.ts
```diff
@@ -0,0 +1,113 @@
...
+    sampler: new ParentBasedSampler({
+      root: new TraceIdRatioBasedSampler(OTEL_TRACE_SAMPLE_RATIO),
+    }),
+    spanProcessors: [
+      new BatchSpanProcessor(
+        new OTLPTraceExporter({
+          url,
+          headers: { Authorization: `Bearer ${apiKey}` },
+        }),
+        { scheduledDelayMillis: 2000 },
```
pauldambra: > [!NOTE]

The exporter/sampler shape here is identical to `apps/code/src/renderer/utils/otel-trace.ts`: the `${apiHost}/i/v1/traces` path, the `new URL()` guard, the `Authorization: Bearer` header, `ParentBasedSampler(TraceIdRatioBasedSampler(OTEL_TRACE_SAMPLE_RATIO))`, and `scheduledDelayMillis: 2000`. The Web-vs-Node provider split is a real divergence and should stay split — but the validated-url + sampler + the magic `/i/v1/traces` and `2000` are one idea expressed twice. The sample ratio is already shared via `@posthog/shared`; the path and batch delay are not. Consider a small host-neutral helper taking `{ apiKey, apiHost }` so those constants live once. (Counterpoint from the reviewers: duplication may be cheaper than the wrong abstraction here — a conscious call either way.)

packages/workspace-server/src/with-span.ts
```diff
@@ -0,0 +1,25 @@
+import { type Attributes, SpanStatusCode } from "@opentelemetry/api";
+import { getWorkspaceServerTracer } from "./otel-trace";
+
+export async function withSpan<T>(
+  name: string,
+  attributes: Attributes,
+  fn: () => Promise<T>,
+): Promise<T> {
+  const tracer = getWorkspaceServerTracer();
```
pauldambra: > [!NOTE]

**[xp]** 🟡 MEDIUM

`withSpan` reaches into the `getWorkspaceServerTracer()` module singleton, while its sibling `traceTrpcCall` (same package, same concern) takes the tracer as a parameter. Two conventions for "where a span gets its tracer." Passing the tracer into `withSpan` too would make it consistent and testable without standing up the module mock that `with-span.test.ts` needs.

apps/code/src/renderer/utils/otel-trace.ts
```diff
@@ -0,0 +1,130 @@
...
+
+let tracerProvider: WebTracerProvider | null = null;
+let tracer: Tracer | null = null;
+
+export function initOtelTracing(): Tracer | null {
+  const apiKey = import.meta.env.VITE_POSTHOG_API_KEY;
+  const apiHost =
+    import.meta.env.VITE_POSTHOG_API_HOST || "https://internal-c.posthog.com";
+
+  if (!apiKey) {
```
pauldambra: > [!NOTE]

**[xp]** 🟡 MEDIUM

This is the most branch-heavy of the three init sites (key-missing gating, url-invalid guard, longtask observer, navigation span start/end pairing, profiler threshold) but has no colocated test, while `node-tracing.ts` and `with-span.ts` are well covered. At minimum test the null-gating and the `traceNavigations` start/end lifecycle — that pairing is the kind of thing that breaks silently.

apps/code/src/renderer/utils/otel-trace.ts
```diff
@@ -0,0 +1,130 @@
...
+import type { ProfilerOnRenderCallback } from "react";
+
+const SLOW_COMMIT_THRESHOLD_MS = 16;
+
+let tracerProvider: WebTracerProvider | null = null;
+let tracer: Tracer | null = null;
+
+export function initOtelTracing(): Tracer | null {
+  const apiKey = import.meta.env.VITE_POSTHOG_API_KEY;
+  const apiHost =
```
pauldambra: > [!NOTE]

**[qa-team/compatibility]** 🟢 LOW

Asymmetric no-op gating: the renderer defaults `apiHost` to `https://internal-c.posthog.com` and exports whenever `VITE_POSTHOG_API_KEY` is present, while `node-tracing.ts` requires BOTH key and host. A build with only the key set would export renderer traces but silently no-op main/workspace-server traces — half-instrumented, disconnected traces. Align the two gates (either both default the host or neither does).

packages/workspace-client/src/client.ts
```diff
@@ -32,7 +41,7 @@ export function createWorkspaceClient(connection: WorkspaceConnection) {
         false: httpBatchLink({
           url,
           transformer: superjson,
-          headers: () => headers,
+          headers: () => tracePropagationHeaders(headers),
```
pauldambra, line 44: > [!NOTE]

**[paul]** 🟢 LOW

`tracePropagationHeaders` injects traceparent on every workspace request unconditionally. When no provider is registered the global propagator should be a no-op, but worth confirming we're not adding empty/garbage headers to every tRPC call. A test asserting "no traceparent header when tracing disabled" would lock that down.

packages/workspace-server/src/services/git/service.ts
```diff
@@ -685,11 +686,10 @@ export class GitService extends TypedEventEmitter<GitCloneEvents> {
...
     signal?: AbortSignal,
   ): Promise<SyncOutput> {
-    const pullResult = await this.pull(
-      directoryPath,
-      remote,
-      undefined,
-      signal,
+    const pullResult = await withSpan(
+      "git.sync.pull",
+      { "git.remote": remote },
```
pauldambra, line 694: > [!NOTE]

**[qa-team/security]** 🟢 LOW

The `git.sync.pull`/`git.sync.push` spans set `git.remote` and export it to the OTLP endpoint. Here it's the remote *name* ("origin"), which is fine — but a resolved remote *URL* can embed credentials (`https://user:token@host/...`) or private repo names. Confirm only the remote name (never a resolved URL) can ever reach a span attribute.

pauldambra (commented):
QA Swarm review complete — see inline comments. (Diff base note: an earlier pass was run against a stale local `main`; this review reflects the authoritative PR diff against origin/main. Two findings about an `agent.ts` error-masking change and a duplicate `initializePostHog` were stale-diff artifacts and have been dropped.)

packages/workspace-server/src/node-tracing.ts
```diff
@@ -0,0 +1,113 @@
...
+  options: NodeTracingOptions,
+): NodeTracing | null {
+  const apiKey = process.env.VITE_POSTHOG_API_KEY;
+  const apiHost = process.env.VITE_POSTHOG_API_HOST;
+
+  if (!apiKey || !apiHost) {
+    return null;
+  }
+
+  const url = `${apiHost}/i/v1/traces`;
```
pauldambra: > [!NOTE]

**[security-audit/ssrf + qa-team]** 🟢 LOW / ⚪ NIT

Two small hardening notes on the URL build: (1) `${apiHost}/i/v1/traces` double-slashes if the host is configured with a trailing slash — strip it first. (2) The `new URL(url)` guard validates parseability but not scheme; a misconfigured `http://` host would send the `Authorization: Bearer` header in cleartext. Not runtime-exploitable (host is a build/launch-time constant), but asserting `protocol === "https:"` (allowing localhost http in dev) is cheap hardening. Same applies to the renderer init.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team, paul-reviewer, xp-reviewer, security-audit

Well-structured, well-gated tracing plumbing that no-ops cleanly when unconfigured, with a genuine DRY win (the shared node factory) and good parameterised tests. One real correctness bug holds it back from a clean pass: buffered spans are dropped on shutdown.

### Key findings

**🟠 HIGH**
- **`serve.ts:39` — shutdown drops spans.** `void shutdownOtelTracing()` is fire-and-forget before a synchronous `process.exit(0)`; the flush never completes, so every clean workspace-server shutdown loses up to 2s of buffered spans. Await the flush within the grace window.

**🟡 MEDIUM**
- **`renderer otel-trace.ts:128` — renderer flush is unreliable** (`pagehide`-only, fire-and-forget, no `shutdown()`); recent renderer spans routinely lost.
- **`node-tracing.ts:67` ↔ renderer — exporter/sampler shape duplicated** across node + web; `/i/v1/traces` and `scheduledDelayMillis: 2000` are magic numbers not shared even though the ratio is. *(convergent: paul + xp)*
- **`with-span.ts:9` — inconsistent tracer acquisition**: `withSpan` reaches into the singleton, `traceTrpcCall` takes the tracer as a param. Pick one.
- **`renderer otel-trace.ts` — no colocated test** for the most branch-heavy init site (gating, url-guard, longtask, navigation start/end, profiler threshold).

**🟢 LOW / ⚪ NIT**
- Gating asymmetry (renderer defaults host + needs only key; node needs both) → possible half-instrumented traces.
- `git.remote` exported (name only today — confirm a resolved URL can't reach it).
- traceparent injected unconditionally on every workspace call (add a "no header when disabled" test).
- URL build: double-slash on trailing-slash host; `new URL()` doesn't assert `https:` scheme.
- No log line at init → at 2% sampling, hard to tell "tracing flowing" from "silently broken."
- Renderer→main traces are not linked to main/workspace-server (electron-trpc IPC has no traceparent hop) — already documented as a deliberate deferral in this PR.

### Convergence (highest confidence)
- **node↔web provider duplication** — flagged independently by paul and xp.
- **renderer→main propagation gap** — flagged by qa-team and security-audit (both note it's an acknowledged, non-blocking gap).

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Correct gating + sampling; the shutdown-flush handling (server + renderer) is the one real defect — request changes on that, rest is LOW/NIT. |
| 👤 paul | No correctness blockers in his read; "don't let it drift" (dup provider setup, no init log) + a "confirm no header when disabled" test. Stamp-with-comments. |
| 📐 xp | Genuine DRY win; tighten `withSpan`/`traceTrpcCall` consistency and add the missing renderer test. "Make it right" polish on working code. |
| 🛡 security-audit | No exploitable findings. Bearer key is a public `phc_` ingestion key; span attributes deliberately low-sensitivity; inbound traceparent is behind the shared-secret auth. Two LOW hardening notes only. |

> **Diff-base note:** an earlier pass ran against a stale local `main` and surfaced two findings (an `agent.ts` error-masking change, a duplicate `initializePostHog`) that are **not** in the authoritative PR diff — those were dropped from this review.

---

packages/workspace-server/src/serve.ts
pauldambra: Fixed in 7ca2bcf — `shutdown()` now awaits `shutdownOtelTracing()` (bounded by the existing grace timer) before `process.exit`, so a clean workspace-server shutdown no longer drops buffered spans.

apps/code/src/renderer/utils/otel-trace.ts
pauldambra: Fixed in 7ca2bcf — the renderer now also flushes on `visibilitychange`→hidden in addition to `pagehide` (which is unreliable in Electron).

packages/workspace-server/src/node-tracing.ts
pauldambra: Addressed in 7ca2bcf — URL building is now `buildTracesEndpoint()` in @posthog/shared (used by node + renderer), and the batch delay is `OTEL_BATCH_DELAY_MS`, shared alongside the ratio. The magic `/i/v1/traces` and `2000` now live once. The Node-vs-Web provider stays split by design.

packages/workspace-server/src/with-span.ts
pauldambra: Fixed in 7ca2bcf — `withSpan` now takes the tracer as its first parameter, matching `traceTrpcCall`; the singleton lookup moved to the call sites and the unit test no longer needs a module mock.

apps/code/src/renderer/utils/otel-trace.ts
pauldambra: Fixed in 7ca2bcf — added `apps/code/src/renderer/utils/otel-trace.test.ts` covering null-gating, the `traceNavigations` start/end lifecycle, and the profiler commit threshold.

apps/code/src/renderer/utils/otel-trace.ts
pauldambra: Fixed in 7ca2bcf — renderer init now requires both key AND host (dropped the `internal-c` default), matching `node-tracing.ts` and the existing log transport, so a key-only build can't half-instrument.

packages/workspace-client/src/client.ts
pauldambra: Addressed in 7ca2bcf — added `packages/workspace-client/src/client.test.ts` asserting no `traceparent` header is added when tracing is disabled (OTel's global propagator defaults to no-op) and that the input headers aren't mutated.

packages/workspace-server/src/services/git/service.ts
pauldambra: Confirmed: the only value placed in the attribute is the remote *name* (defaults to `origin`) — a resolved remote URL (which could embed credentials) never reaches a span attribute. Resolving as no-change-needed.

packages/workspace-server/src/node-tracing.ts
pauldambra: Fixed in 7ca2bcf — `buildTracesEndpoint()` strips trailing slashes and asserts an `https:` scheme (allowing `localhost`/127.0.0.1 http for dev), shared by node + renderer.

packages/workspace-server/src/with-span.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      withSpan("op", { a: 1 }, async () => {
+        throw new Error("boom");
+      }),
+    ).rejects.toThrow("boom");
+
+    expect(span.recordException).toHaveBeenCalled();
+    expect(span.setStatus).toHaveBeenCalled();
+    expect(span.end).toHaveBeenCalled();
+  });
+});
```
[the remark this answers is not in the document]
pauldambra: Added in 3f2dcfc — a success-path case asserting the resolved value is returned unchanged, `span.end` is called, and neither `recordException` nor an error status is set.

packages/workspace-server/src/services/agent/agent.ts
```diff
@@ -683,49 +684,64 @@ When creating pull requests, add the following footer at the end of the PR descr
...
-          onMcpServersReady: (serverNames) => {
-            this.mcpAppsService.handleDiscovery(serverNames).catch((err) => {
-              this.log.warn("MCP Apps discovery failed", {
-                error: err instanceof Error ? err.message : String(err),
-              });
-            });
-          },
+      const acpConnection = await withSpan(
+        "agent.run.start",
+        {
```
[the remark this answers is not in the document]
pauldambra: Keeping `agent.run.start`. The span wraps only the `agent.run()` call, which resolves once the ACP connection is established — the streaming happens afterwards over the returned connection, outside this span. So the span's duration is the startup phase; `agent.run` would misleadingly imply it covers the whole (multi-minute) run. The name reflects the actual scope.

packages/shared/src/constants.ts
```diff
@@ -9,3 +9,23 @@
 export const SELF_DRIVING_SETUP_TASK_FLAG =
   "posthog-code-self-driving-setup-task";
 export const BRANCH_PREFIX = "posthog-code/";
+
+export const OTEL_TRACE_SAMPLE_RATIO = 0.02;
+
+export const OTEL_BATCH_DELAY_MS = 2000;
+
+export function buildTracesEndpoint(apiHost: string): string | null {
+  const url = `${apiHost.replace(/\/+$/, "")}/i/v1/traces`;
```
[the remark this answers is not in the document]
pauldambra: Fixed in c04e8e3 — replaced the `/\/+$/` trailing-slash regex with a linear `while (host.endsWith("/"))` loop, so there's no polynomial backtracking on a long run of slashes.

packages/workspace-client/src/client.ts
pauldambra: Correction: the unit test I mentioned was reverted in 4e9526a — adding `vitest` to workspace-client (which had no test runner) tripped react-doctor's Socket supply-chain gate on `vitest@4.0.10`. The no-op-when-disabled behaviour is guaranteed by construction anyway: with no provider registered, OTel's global propagator is the default `NoopTextMapPropagator`, so `propagation.inject` adds nothing. `tracePropagationHeaders` is private again.

PR #2673, opened 2026-06-15
feat(sessions): modernize conversation thread with collapsible tool groups

adamleithp:
## What

Rebuilds the conversation thread around a single, consistent tool-call row.

> Very much tested locally, all settings types, works really well.

- **One `ToolRow` primitive** (base-ui Collapsible): trigger = icon + text, content = left-padded box. Every tool view — execute, read, edit, search, fetch, think, MCP, generic, thinking — renders through it, replacing the old per-view card markup.
- **Collapsible tool groups.** A turn's tool work folds into a group chip with a verb-led summary (`Ran 25 commands, read a file, edited a file`) that shows the live action while running. Persisted collapse mode in Settings: **All collapsed** (default) / **Collapse when finished turn** / **No collapsing**. Per-chip expand/collapse overrides held in session view state.
- **Always-visible rows.** Setup/clone progress (`progress_group`) and MCP-app tool calls stay outside groups so MCP iframes keep their mount (`keepMounted`).
- **Assistant text inside a group** is indented to the tool-title column.

image

image

image

image

## Perf

- Grouping is a single pass per streamed token producing rows + `keepMounted` + id→row map (was several walks), with cached summaries for frozen turns.
- No per-scroll re-renders.

## Notes

- Known pre-existing, unrelated: cloud threads can show a duplicate user message (a `mergeConversationItems` dedup gap when an attachment is inlined as text vs. a chip). Not touched here.

### DIFF
--- a/packages/ui/src/features/mcp-apps/components/McpToolView.tsx
+++ b/packages/ui/src/features/mcp-apps/components/McpToolView.tsx
@@ -1,17 +1,14 @@
 import { Plugs } from "@phosphor-icons/react";
-import { Box, Flex } from "@radix-ui/themes";
-import { useState } from "react";
 import {
   getPostHogExecDisplay,
   isPostHogExecTool,
 } from "../../posthog-mcp/utils/posthog-exec-display";
+import { ToolRow } from "../../sessions/components/session-update/ToolRow";
 import {
+  ContentPre,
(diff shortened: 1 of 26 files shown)


charlesvien (approved):
LGTM

PR #2685, opened 2026-06-15
feat(ui): smooth streaming of agent message tokens

### LINKED ISSUE #2517: Smooth token streaming
https://upstash.com/blog/smooth-streaming

Make sure it still scrolls to the bottom / auto-fixes to the bottom as it streams

archievi:
## Problem

Closes #2517

Streamed assistant tokens arrive from the agent in bursty chunks. Because the message component re-renders the full accumulated text on every chunk, the text jumps forward unevenly instead of reading as smooth, steady typing. The issue asks for smooth token streaming (per the [Upstash smooth-streaming approach](https://upstash.com/blog/smooth-streaming)) while keeping the conversation pinned to the bottom as it streams.

## Changes

- Add a `useSmoothText` hook (`packages/ui/src/primitives/hooks/useSmoothText.ts`) that decouples the render rate from the token-arrival rate. It reveals the accumulated text a few characters per frame via `requestAnimationFrame`, so bursts of tokens are eased out at a steady rate (~120 chars/sec).
- Use it in `AgentMessage` to render the smoothed text. Copy-to-clipboard still copies the full `content`, not the partially revealed prefix.

Design notes:
- Text already present when the component mounts is shown immediately, so re-rendered history and completed messages never replay a typewriter effect. Only growth that happens while mounted is animated.
- It snaps instantly when the source is replaced (i.e. the new value is not a prefix of what was being revealed) or when the user has `prefers-reduced-motion: reduce` set.
- A large lag (e.g. a very long buffered chunk) snaps rather than crawling, so catch-up never feels sluggish.
- Auto-scroll-to-bottom is preserved: the virtualized list re-pins to the end whenever the rendered height changes, and since the smoothed text grows the height incrementally, the view follows it down naturally (no change needed to the scroll logic).

The reveal rate easing is extracted into a pure `nextRevealLength` function so it is unit-testable without timers.

## How did you test this?

- Added `packages/ui/src/primitives/hooks/useSmoothText.test.ts` covering the pure easing function and the hook (mocked `requestAnimationFrame`/`matchMedia`): immediate render on mount, gradual reveal of an appended burst, full catch-up, snap-on-replace, and reduced-motion. Verified the gradual-reveal and snap-on-replace tests fail when the hook is reduced to a pass-through, confirming they guard the behavior.
- `pnpm --filter @posthog/ui test` -> 82 files, 737 tests pass.
- `pnpm --filter @posthog/ui typecheck` (`tsc --noEmit`) passes; the pre-commit hook's full `pnpm typecheck` also passes.
- `biome check` passes on the changed files.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/ui/src/features/sessions/components/session-update/AgentMessage.tsx
+++ b/packages/ui/src/features/sessions/components/session-update/AgentMessage.tsx
@@ -3,6 +3,7 @@ import { Box, Code, IconButton } from "@radix-ui/themes";
 import { memo, useCallback, useMemo, useState } from "react";
 import type { Components } from "react-markdown";
 import { HighlightedCode } from "../../../../primitives/HighlightedCode";
+import { useSmoothText } from "../../../../primitives/hooks/useSmoothText";
 import { Tooltip } from "../../../../primitives/Tooltip";
 import { usePendingScrollStore } from "../../../code-editor/pendingScrollStore";
 import { MarkdownRenderer } from "../../../editor/components/MarkdownRenderer";
@@ -144,6 +145,9 @@ export const AgentMessage = memo(function AgentMessage({
   content,
 }: AgentMessageProps) {
   const [copied, setCopied] = useState(false);
+  // Reveal streamed tokens at a steady rate so bursty chunks read as smooth
+  // typing. Copy always uses the full `content`, not the partial reveal.
+  const displayedContent = useSmoothText(content);
 
   const handleCopy = useCallback(() => {
     navigator.clipboard.writeText(content);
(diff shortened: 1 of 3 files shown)


charlesvien:
hey @archievi, thanks for this! excited to try it out, can you handle greptile comments and do a second review pass on it?

archievi:
Thanks @charlesvien! Handled the Greptile comment — `nextRevealLength` is now a single `it.each` table (pushed in e0b12a1), and the "caught up past target" case is its own row instead of a buried second assertion. Did a second pass over the diff myself; all 10 tests pass locally. Ready for another look whenever you have a chance. 🙏

charlesvien:
We're going to pull your work into https://github.com/PostHog/code/pull/2716, will add you as a co-author, thank you for this! Super excited to have buttery smooth token streaming 🧈

PR #2696, opened 2026-06-16
feat(settings): label meep sounds as lorg and smol

pauldambra:
## Problem

The two meep notification sounds were labelled "Meep" and "Meep (smol)", which didn't make it clear that the first is the larger variant. Per the Slack thread, they should read "meep (lorg)" and "meep (smol)".

## Changes

Renamed the larger meep option's label from "Meep" to "Meep (lorg)" in the sound picker, on both desktop (`GeneralSettings.tsx`) and mobile (`settings/index.tsx`). The "Meep (smol)" label was already correct. Underlying sound values are unchanged.

## How did you test this?

Not tested beyond the edit — purely a display-label change.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/CT2BU33N1/p1781559380714749?thread_ts=1781559380.714749&cid=CT2BU33N1)*

### DIFF
--- a/apps/mobile/src/app/settings/index.tsx
+++ b/apps/mobile/src/app/settings/index.tsx
@@ -42,7 +42,7 @@ const FONT_SIZE_OPTIONS: ReadonlyArray<{
 
 const SOUND_OPTIONS: ReadonlyArray<{ value: CompletionSound; label: string }> =
   [
-    { value: "meep", label: "Meep" },
+    { value: "meep", label: "Meep (lorg)" },
     { value: "meep-smol", label: "Meep (smol)" },
     { value: "knock", label: "Knock" },
     { value: "ring", label: "Ring" },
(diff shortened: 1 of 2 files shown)


frankh:
personally i prefer just (smol). I've never seen lorg used before and i think having both labelled makes the (smol) less cute

idk tho lol

pauldambra:
@frankh you're right to push back emdash i am coining lorg

we're both here at the start of something huge

options right now are

* Meep - no label, as it stands, just close the PR
* Meep (og) - let people know this is the coolest thing ever that started it all
* Meep (lorg) - give the peopl what they (don't know they) want - a new way of saying large that they're embarassed about

charlesvien:
should we keep Meep and for fun you could create the Meep (lorg) sound and throw it in a PR?

edit: yes

PR #2699, opened 2026-06-16
fix(inbox): make "Entire project" scope show project-wide reports

oliverb123:
## Problem

Selecting the **Entire project** reviewer scope in the inbox still only showed reports where you are the suggested reviewer — identical to **For you**.

## Root cause

`buildSignalReportListOrdering` always appended `-is_suggested_reviewer`, floating the current user's reports to the top of every status group. Because the inbox only loads the first page (`REPORTS_PAGE_SIZE = 100`, no infinite scroll wired) and projects have thousands of reports, that first page was entirely the user's own reports. Both `for-you` and `entire-project` sent no `suggested_reviewers` param, so the only thing distinguishing them was a client-side filter applied on top of an already user-dominated page.

## Fix

Remove the always-on `-is_suggested_reviewer` ordering tiebreak and apply the reviewer scope **server-side** via the existing `suggested_reviewers` query param instead:

- **For you** → `suggested_reviewers=<current user uuid>` (the float was what previously fed this scope, so it now filters explicitly).
- **Teammate** → `suggested_reviewers=<teammate uuid>` (unchanged).
- **Entire project** → no `suggested_reviewers`, ordered `status,<field>` → first page is the project-wide top reports.
- **Runs tab** (`ignoreScope`) → unchanged (still project-wide).

`ordering` and `suggested_reviewers` are part of the React Query key, so each scope is its own cached query and switching refetches cleanly.

## Files

- `packages/core/src/inbox/reportFiltering.ts` — drop the `-is_suggested_reviewer` tiebreak.
- `packages/ui/src/features/inbox/hooks/useInboxAllReports.ts` — resolve current user; filter For-you server-side by uuid.
- `packages/core/src/inbox/reportFiltering.test.ts` — updated ordering expectations + guard.

## Testing

- `reportFiltering` unit tests pass.
- Manual: **For you** still shows your suggested reports; **Entire project** now includes reports where you are not the suggested reviewer; a **teammate** scope shows that teammate's reports.

> Note: full-monorepo typecheck currently has pre-existing failures from unbuilt sibling packages (`@posthog/platform/*`, `workspace-server`, zod v3/v4 mismatch), unrelated to this change.

---

### DIFF
--- a/packages/core/src/inbox/reportFiltering.test.ts
+++ b/packages/core/src/inbox/reportFiltering.test.ts
@@ -98,22 +98,12 @@ describe("filterReportsBySearch", () => {
 });
 
 describe("buildSignalReportListOrdering", () => {
-  it("puts status then descending field", () => {
-    expect(buildSignalReportListOrdering("total_weight", "desc")).toBe(
-      "status,-total_weight",
-    );
-  });
-
-  it("puts status then ascending field", () => {
-    expect(buildSignalReportListOrdering("created_at", "asc")).toBe(
-      "status,created_at",
-    );
-  });
-
-  it("works for signal_count", () => {
-    expect(buildSignalReportListOrdering("signal_count", "desc")).toBe(
-      "status,-signal_count",
-    );
+  it.each([
+    ["total_weight", "desc", "status,-total_weight"],
+    ["created_at", "asc", "status,created_at"],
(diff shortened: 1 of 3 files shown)


Twixes (approved):
Yess

PR #2706, opened 2026-06-16
fix(sessions): keep queued messages in queue when cancelling

MattPua:
## Problem

- Cancelling (Stop) an in-flight message dumps all queued messages back into the input box as combined text — they lose queued state.

## Change

https://github.com/user-attachments/assets/ea042c39-9cbd-4f99-bfd2-38a8739fbba9

- Stop now cancels **only the in-flight turn**; queued messages stay queued (still shown as cards, individually removable).
- New `queuePaused` flag on `AgentSession`, set by `cancelPrompt`, suppresses the auto-drain on the cancel-induced turn-end.
- Flag cleared on next user send, so kept messages resume after that turn.
- `handleCancelPrompt` no longer drains the queue into the editor.

## Notes

- Resume is "on next send" — no new UI.
- Footer `(N queued)` hides after Stop (only shows while pending), but queued cards stay visible.

## Testing

- shared / core / agent typecheck clean
- `sessionServiceHost` (116) + `sessionStore` (10) tests pass
- Biome lint clean

---

### DIFF
--- a/packages/core/src/sessions/sessionService.ts
+++ b/packages/core/src/sessions/sessionService.ts
@@ -1533,7 +1533,8 @@ export class SessionService {
     const hasQueuedMessages =
       freshSession &&
       freshSession.messageQueue.length > 0 &&
-      freshSession.status === "connected";
+      freshSession.status === "connected" &&
+      !freshSession.queuePaused;
 
     if (hasQueuedMessages) {
       setTimeout(() => {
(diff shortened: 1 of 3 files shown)


charlesvien:
I'm worried about changing this because people have come to expect this behavior as the default from claude code, codex, etc. @MattPua

MattPua:
> I'm worried about changing this because people have come to expect this behavior as the default from claude code, codex, etc. @MattPua

okay, no worries then!

PR #2708, opened 2026-06-16
feat(settings): configurable branch name prefix

KarloAldrete:
## Problem

The branch name PostHog Code generates for a task is hardcoded with the `posthog-code/` prefix. Some users want a different prefix — or none — to match their team's branch conventions.

Closes PostHog/posthog#76277

## Changes

- New **"Branch name prefix"** setting under **Settings → Worktrees** (`branchPrefix`), defaulting to the current `posthog-code/` so existing behavior is unchanged.
- `deriveBranchName` / `suggestBranchName` take an optional `prefix` argument (default `BRANCH_PREFIX`); existing callers are unaffected.
- `getSuggestedBranchName` reads the setting, so the generated branch name honors it wherever the app suggests one (Create PR, "Branch here").
- The prefix is validated before it's saved (`validateBranchPrefix`): rejects names git would reject and a leading `-`; empty = no prefix. Invalid input shows an inline error and isn't persisted.

<img width="820" src="https://raw.githubusercontent.com/KarloAldrete/code/main/.github/pr-media/branch-name-prefix.png" alt="Branch name prefix setting under Settings -> Worktrees" />

## How did you test this?

- `pnpm --filter @posthog/core --filter @posthog/ui typecheck` — clean.
- `biome lint packages/core/src/git-interaction` + Biome check on all changed files — clean, no `noRestrictedImports`.
- `vitest run` on `branchName.test.ts` + `deriveBranchName.test.ts` — 38 passing (added: custom prefix, empty prefix, task-ID fallback with prefix, `validateBranchPrefix`).
- Manually (by the contributor): set a custom prefix in Settings → Worktrees; "Create branch" on a worktree task used the configured prefix; an empty prefix dropped it; invalid values showed the inline error.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/agent/src/adapters/claude/claude-agent.ts
+++ b/packages/agent/src/adapters/claude/claude-agent.ts
@@ -1689,7 +1689,10 @@ export class ClaudeAcpAgent extends BaseAcpAgent {
       ...initialInProcess,
     };
 
-    const systemPrompt = buildSystemPrompt(meta?.systemPrompt);
+    const systemPrompt = buildSystemPrompt(
+      meta?.systemPrompt,
+      meta?.branchPrefix,
+    );
 
     if (meta?.mcpToolApprovals) {
       setMcpToolApprovalStates(meta.mcpToolApprovals);
--- a/packages/agent/src/adapters/claude/session/instructions.ts
+++ b/packages/agent/src/adapters/claude/session/instructions.ts
@@ -1,10 +1,17 @@
-const BRANCH_NAMING = `
+import { BRANCH_PREFIX } from "@posthog/shared";
+
+function buildBranchNaming(branchPrefix: string = BRANCH_PREFIX): string {
+  const namingLine = branchPrefix
+    ? `When creating a new branch, prefix it with \`${branchPrefix}\` (e.g. \`${branchPrefix}fix-login-redirect\`).`
+    : "When creating a new branch, use a short, descriptive name.";
+  return `
 # Branch Naming
 
 When working in a detached HEAD state, create a descriptive branch name based on the work being done before committing. Do this automatically without asking the user.
 
-When creating a new branch, prefix it with \`posthog-code/\` (e.g. \`posthog-code/fix-login-redirect\`).
+${namingLine}
 `;
+}
 
 const PLAN_MODE = `
 # Plan Mode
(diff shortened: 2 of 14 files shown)


KarloAldrete:
Thanks for the review! Parameterised those tests with `it.each` in 2b29791 — `validateBranchPrefix`, the `deriveBranchName` custom-prefix cases, and `suggestBranchName` — matching the parameterised-test convention in AGENTS.md.

charlesvien:
Should this be in worktree settings tho? You can create PR's outside of worktrees that would respect this setting,

jonathanlab (changes requested):
Thanks for your contribution!

Two things in this PR that still need work:

1. This should not live in the worktrees settings page, as it also affects functionality outside worktrees 
2. We should wire this customizable prefix into the agent system prompts. The agent will still use `posthog-code` there even if you customize it via the settings.

KarloAldrete:
Thanks for the review @jonathanlab! Addressed both points (also rebased onto `main`):

**1. Moved the setting out of Worktrees.** It now lives in **General → Git** ("Branch name prefix"), since it affects more than worktrees. Store field and validation are unchanged.

**2. Wired the prefix into the agent's system prompt** so a customized value actually takes effect. Threaded `branchPrefix` through the local-session path end to end, alongside the existing `customInstructions`:

- **workspace-server** — carried through `startSessionInput` / `reconnectSessionInput`, `SessionConfig`, and into `buildSystemPrompt`, which now renders the branch-naming line from the setting (empty prefix → no required prefix, matching `deriveBranchName`).
- **claude adapter** — passed via `NewSessionMeta` into the appended instructions; `buildAppendedInstructions(branchPrefix)` replaces the static `APPENDED_INSTRUCTIONS`. This covers the instruction that previously hardcoded `posthog-code/` even when the user customized it.

Everything falls back to `BRANCH_PREFIX` when the value is absent.

**Two follow-ups I deliberately left out of this PR** — happy to do either here or separately, your call:
- The **cloud `agent-server` prompt** (`buildCloudSystemPrompt`) — the setting lives in the desktop renderer store, so reaching the cloud agent process means serializing it through `AgentServerConfig` at spawn time. That's a separate plumbing path from the local one, so I kept it out to keep this PR focused.
- The **`git_signed_commit` tool descriptions** — static module constants with no session context; lower impact (tool metadata rather than a direct instruction).

Tested: `typecheck` clean across all packages; `@posthog/core` (1804), `@posthog/agent` (819), `@posthog/ui` (1067) green; `biome check` + host-boundaries clean on changed files.

PR #2710, opened 2026-06-16
perf(sidebar): stop re-rendering on every streamed token

KarloAldrete:
## Problem

The sidebar re-renders on **every streamed agent token**. It consumes the whole `sessions` record via `useSessions()`, which immer replaces on each appended event (one per token). Because the sidebar is mounted at the root, this re-renders the whole tree on every token during streaming.

Part of #2162.

## Changes

`deriveTaskData` only reads four session fields (`isPromptPending`, `pendingPermissions.size`, `cloudStatus`, `cloudOutput.pr_url`) — never `events`.

- **`computeSidebarSessionSignature`** (core, pure): a primitive signature of just those fields.
- **`useSidebarSessionMap`** (ui): subscribes to that signature and rebuilds the `taskId → session` map only when a sidebar-relevant field changes — not on every token.
- **`useSidebarData`** uses it instead of `useSessions()`.

No behavior change — `deriveTaskData` receives the same data; only the subscription granularity changed.

## How did you test this?

New render-count test (`useSidebarSessionMap.test.tsx`):

| 20 streamed tokens | sidebar re-renders |
|---|---|
| before (`useSessions()`) | **20** |
| after (`useSidebarSessionMap()`) | **0** |
| after, when a relevant field changes | 1 |

- New pure-function test (`computeSidebarSessionSignature.test.ts`): the signature ignores `events`, changes on the 4 fields.
- `pnpm --filter @posthog/core --filter @posthog/ui typecheck` — clean.
- `biome lint packages/core` — clean, no `noRestrictedImports`; host-boundaries unchanged.
- Existing suites green: core/sidebar (48 tests), ui/features/sessions (230 tests).

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/sidebar/buildSidebarData.ts
+++ b/packages/core/src/sidebar/buildSidebarData.ts
@@ -1,3 +1,4 @@
+import type { WorkspaceMode } from "@posthog/shared";
 import type { Task, TaskRunStatus } from "@posthog/shared/domain-types";
 import { getRepositoryInfo } from "./groupTasks";
 import type { TaskData } from "./sidebarData.types";
@@ -93,6 +94,8 @@ export interface TaskWorkspace {
   folderPath?: string | null;
   branchName?: string | null;
   linkedBranch?: string | null;
+  mode?: WorkspaceMode | null;
+  worktreePath?: string | null;
 }
 
 export interface TaskTimestamp {
@@ -101,7 +104,13 @@ export interface TaskTimestamp {
 }
 
 export interface DeriveTaskDataContext {
-  session: TaskSession | undefined;
+  /**
+   * Optional live session state. The sidebar omits this (it builds the list
+   * from tasks alone and layers live status in per-row via
+   * useSidebarStatusForTask); other callers — e.g. the canvas hooks — still
+   * pass it to fold live status straight into the derived data.
+   */
+  session?: TaskSession;
   workspace: TaskWorkspace | undefined;
   timestamp: TaskTimestamp | undefined;
   pinnedIds: ReadonlySet<string>;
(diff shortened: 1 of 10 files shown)


jonathanlab (changes requested):
Thank you! I do think that a map is not the most maintainable/reliable solution here. Instead, the main issue here is that the sidebar is asking for way too much data, causing all these unnecessary rerenders. We should instead build the list from the task names + their order and use that :)

KarloAldrete:
Thanks @jonathanlab, that was the right call. Reworked it to build the list from the tasks and read session status per-row, so the sidebar no longer pulls the whole sessions record.

**What changed**
- `useSidebarData` no longer touches the session store — the list (order, grouping, pinning) is built from tasks alone.
- Each row subscribes to just its own live status via `useSidebarStatusForTask` (`isGenerating` / `needsPermission` / `taskRunStatus` / `cloudPrUrl`, `shallow`-compared). A streamed token leaves those equal, so nothing re-renders; only the row whose status actually changes re-renders.
- While here: stopped each row re-subscribing to the whole workspaces query (derive `workspaceMode`/`worktreePath` once in `deriveTaskData`), and moved the archive-guard check to an imperative store read.

**Measured** (render counts, real session store):

| | main | per-row (this PR) |
|---|--:|--:|
| Sidebar re-renders / 100 streamed tokens | 100 | **0** |
| Rows re-rendered on one status change | whole list | **1** |

Verified with render-count tests (jsdom) and a real-Chromium harness (spinner / permission / cloud-status / PR-link surface on the right row only). No `map`/signature anymore.

KarloAldrete:
Superseded by #3063, which lands this same architecture (tasks-derived sidebar list + per-row session signature) as part of a broader sessions perf pass — closing in its favor. Happy to see it shipped! 🙌

PR #2712, opened 2026-06-16
fix(task-input): prefill repo dropdown when starting task from sidebar "+" icon

### LINKED ISSUE #1454: When I press the + button, it does not set the repo or cloud state for the task input page based on the project I clicked + on

mp-hog:
## Problem

Closes https://github.com/PostHog/code/issues/1454

In the Tasks sidebar each repo group header has a hover "+" button to start a new task. Clicking it did not scope the new-task screen with that repo in cloud-mode, so that tasks were started under wrong repo.

It's this relation:

new-task-repo-dropdown

## Fix

Make `folderId` drive both selectors. A registered folder already carries both `path` (local) and `remoteUrl` (the resolved `owner/repo` slug), so it's a single source of truth.

New `useInitialRepoSelectionFromFolderId` hook + pure `resolveRepoSelectionForFolder` resolver:
- always prefill the local directory from the folder path
- prefill the cloud repo when the folder's remote is a real `owner/repo` that is a connected GitHub integration
- keep the current workspace mode when it can represent the repo, and only switch (Cloud → last-used local mode) when it can't (a local-only repo while in Cloud). The switch uses the non-persisting setter, so it doesn't overwrite the user's mode preference.

## Behavior

| You're in | Click "+" on… | Result |
| --- | --- | --- |
| Cloud | cloud-capable repo | dropdown follows the repo; stay in Cloud |
| Local | dual-capable repo | directory prefilled; stay in Local (cloud repo also seeded) |
| Cloud | local-only repo | switch to Local; directory prefilled |
| — | generic **New task** button | unchanged — falls to last-used defaults |

## Testing

- Unit tests for the pure resolver, `areReposReady`, and the hook cover every branch incl. the loading-gate and once-per-`folderId` guard.
- `@posthog/ui` typecheck + biome clean; task-detail suite green (22/22); full `@posthog/ui` suite green (752/752).
- Manually with local app, + icon switches to respective repo

## Known limits (out of scope)

- A cloud-only group with **no registered local folder** falls back to defaults (no `folderId`/`remoteUrl` to key off).
- In Cloud mode, "+" on a repo whose org is **not a connected GitHub integration** intentionally switches to Local (per the "switch only when the mode can't represent the repo" rule).


charlesvien:
Hey @mp-hog, how should this work if I am working on a local task, I click "+"? I would expect this to pre-fill the repo on local, but right now it does not and always tries to find the cloud repo.

mp-hog:
> Hey @mp-hog, how should this work if I am working on a local task, I click "+"? I would expect this to pre-fill the repo on local, but right now it does not and always tries to find the cloud repo.

Thanks for your feedback @charlesvien !

Before this PR the open mode is the global last-used setting (which does not distinguish between repo). The logic for this PR the same logic, but it adds a fallback to Local if the mode is not applicable under the given repo (it wouldn't switch to Cloud).

You are right though that it might still be surprising, as we don't distinguish this setting between repos, but the PR as is improves the Repo-Selection and is not making the Mode selection worse.

I have locally started working on infering the Mode from the task history of a repo, but it touches a bunch of files and I think goes beyond the scope of this PR. That I believe would solve the surprising bit of the current logic.

I would suggest: Ship as is and improve the Mode-selection in a follow-up. What do you think?

Edit: Follow-up: https://github.com/PostHog/code/pull/2939

charlesvien (approved):

PR #2717, opened 2026-06-16
fix(sessions): advance generating verb on tool completion, not a timer

adamleithp:
## Problem

In a task run, the "generating" status word (the gerund — *Pondering…*, *Scanning…*, etc.) cycled on a fixed 2-second timer. It kept changing even when the agent was stuck, so the one signal you'd reach for to answer "is it still working or stalled?" was untrustworthy.

## Changes

The word now advances **only when a tool/MCP call finishes**, not on a timer.

- The conversation builder counts tool calls that settle into a terminal status (`completed` / `failed` / `cancelled`), counting each call once, and exposes `completedToolCallCount` on `BuildResult`.
- `GeneratingIndicator` swaps its word when that count changes (and never repeats the same word twice in a row). The 2s interval is gone.
- A stalled agent now keeps the same word; the pulsing brain icon and elapsing timer still show liveness.

Scope is the desktop/web app. Mobile's separate `ThinkingIndicator` is unchanged.

## How did you test this?

- Added unit tests for `completedToolCallCount` (pending stays 0, settles to 1 on completion, no double-count on repeat updates, counts failed/cancelled, accumulates across turns) and extended the incremental/full-rebuild equivalence check to cover the new field.
- `pnpm --filter @posthog/ui exec vitest run src/features/sessions` — 233 passed.
- `pnpm --filter @posthog/ui typecheck` — clean. Biome lint — clean.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1781649826606099?thread_ts=1781649826.606099&cid=C09G8Q32R6F)*

### DIFF
--- a/packages/ui/src/features/sessions/components/ConversationView.tsx
+++ b/packages/ui/src/features/sessions/components/ConversationView.tsx
@@ -115,6 +115,7 @@ export function ConversationView({
     items: conversationItems,
     lastTurnInfo,
     isCompacting,
+    completedToolCallCount,
   } = useConversationItems(events, isPromptPending, {
     showDebugLogs,
   });
@@ -348,6 +349,7 @@ export function ConversationView({
         pausedDurationMs={pausedDurationMs}
         isCompacting={isCompacting}
         usage={contextUsage}
+        completedToolCallCount={completedToolCallCount}
       />
     </div>
   );
--- a/packages/ui/src/features/sessions/components/GeneratingIndicator.tsx
+++ b/packages/ui/src/features/sessions/components/GeneratingIndicator.tsx
@@ -159,14 +159,13 @@ export function GeneratingIndicator({
 
   // Advance the word only when a tool/MCP call finishes (activityKey changes),
   // not on an interval. The initial word stays put until the first call settles.
+  // Adjusted during render (React's blessed pattern for deriving state from a
+  // changed prop) rather than in an effect, so it never paints a stale word.
   const prevActivityKeyRef = useRef(activityKey);
-  useEffect(() => {
-    if (activityKey === undefined || activityKey === prevActivityKeyRef.current) {
-      return;
-    }
+  if (activityKey !== undefined && activityKey !== prevActivityKeyRef.current) {
     prevActivityKeyRef.current = activityKey;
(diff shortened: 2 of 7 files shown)


adamleithp:
## Review

**What this does.** The generating status word now advances when a tool/MCP call settles to a terminal status, instead of rotating on a 2s timer. A stalled agent keeps the same word, so word movement now signals real work rather than fake liveness — the elapsed timer already carries the "still alive" signal.

**Design.** Counting happens in the builder where status transitions are already tracked, which is the right layer. `completedToolCallCount` increments once on the first terminal transition (`!wasTerminal && isTerminal`), so repeated `tool_call_update`s don't double-count, and both the `tool_call` and `tool_call_update` branches plus the fresh-entry-already-terminal case are covered. The monotonic counter is plumbed through `BuildResult` → both the full and incremental builders → the hook → `SessionFooter` → the indicator, where it's used purely as a React key for "advance the word." Idle path full-rebuilds the count from scratch; streaming path accumulates in the persistent builder — both produce the same total, no drift.

**Notes / things considered:**
- Pure-text or pre-first-tool turns now hold the initial word static. This is intentional and fine — the ticking elapsed timer carries liveness, so the word doesn't need to. If we want streaming-heavy turns to tick too, the follow-up would be to fold an assistant text/thought-block counter into the same `activityKey` (not per-turn, which is too coarse and never advances on the long single-turn stall this fix targets).
- Indicator remounts each generating session and seeds `prevActivityKeyRef` from the current count, so there's no spurious advance on mount.

**Review feedback addressed (latest commit):**
- react-doctor `no-adjust-state-on-prop-change`: the word-advance moved from an effect into the render body (React's derive-state-from-changed-prop pattern), avoiding a stale paint.
- Biome format fix on `GeneratingIndicator.tsx`.
- Terminal-status test converted to `it.each` so `completed` / `failed` / `cancelled` are each individually exercised.

No correctness bugs found; stays within the `@posthog/ui` layer with no logic leak. LGTM.

PR #2733, opened 2026-06-17
fix(skills): symlink node_modules in Codex mirror instead of deep-copying

VXNCXNX:
## Summary

**On every app launch** (and again every 30 minutes, and after any skill create/edit/install), `PosthogPluginService` re-mirrors all user skills into `~/.agents/skills`. The copy is `fs.cp(..., { recursive: true, dereference: true })` with no filter, so each skill's `node_modules` is **re-duplicated from scratch every single time** the app starts, not copied once.

For a dependency-heavy skill (~1.1 GB with ~700 MB `node_modules`) that means a ~1.4 GB copy under `~/.agents/skills` (most of it duplicated deps), rewritten on every startup. The recurring disk churn and startup I/O is the real cost here, and there's currently no way to opt out of the mirror.

## Changes

- New `copySkillDirLinkingNodeModules()`: skips the skill's top-level `node_modules` during the copy (via an `fs.cp` `filter`), then symlinks it to the source. Dependencies still resolve through the symlink, so mirrored skills stay runnable, with no duplication, no transient disk peak, and no per-startup I/O churn. Used by both `mirrorUserSkillsToCodex` and `syncCodexSkills`.
- New `POSTHOG_DISABLE_CODEX_MIRROR=1` env opt-out for users who don't use Codex; gates the startup calls (with a log line) plus early-returns in both copy paths.

## Tests

`packages/workspace-server` vitest, all green:
- `node_modules` becomes a symlink and a dependency resolves through it
- a skill without `node_modules` is mirrored unchanged
- the opt-out no-ops both copy paths (`mirrorUserSkillsToCodex` + `syncCodexSkills`)

`pnpm --filter @posthog/workspace-server test` passes; `tsc --noEmit` clean; biome clean.

## Notes

Playwright browser binaries live in the shared `~/Library/Caches/ms-playwright`, outside `node_modules`, so they are unaffected.

**Possible follow-up (happy to do if wanted):** surface the opt-out as a Settings toggle backed by `IWorkspaceSettings` (a `SettingRow`/`Switch`, mirroring the existing "Keep awake while agents work" pattern). Left out here to keep this focused, since the placement and default feel like a product call for the team.

---

### DIFF
--- a/packages/workspace-server/src/services/posthog-plugin/codex-mirror.test.ts
+++ b/packages/workspace-server/src/services/posthog-plugin/codex-mirror.test.ts
@@ -1,5 +1,13 @@
-import { existsSync } from "node:fs";
-import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { existsSync, lstatSync } from "node:fs";
+import {
+  lstat,
+  mkdir,
+  mkdtemp,
+  readFile,
+  realpath,
+  rm,
+  writeFile,
+} from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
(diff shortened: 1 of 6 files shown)


VXNCXNX:
Thanks for the review. Addressed all three points in `6c3717ed`:

- **Symlink fallback (P2):** when `symlink`/junction creation is rejected (restricted filesystem or sandbox policy), it now falls back to a real `node_modules` copy, so a mirrored skill never ends up without its dependencies. This preserves the pre-change deep-copy behavior in those environments. Added a test that forces `symlink` to throw and asserts the deps are copied.
- **Redundant `rm` (P2):** removed from the happy path (the `fs.cp` filter never writes `node_modules` and callers remove `dest` first). The `rm` now only runs in the fallback branch before the recovery copy.
- **Duplicated env save/restore (P2):** extracted into a shared `withEnvVar()` helper in `test-helpers.ts`, used by both opt-out tests.

`pnpm --filter @posthog/workspace-server test` is green (12 tests), `tsc --noEmit` clean, biome clean.

VXNCXNX:
Closing this as superseded by #2760.

That PR reworks the Codex mirror to **symlink** skills per-session (`codex-home.ts` → `linkSkillsInto`) instead of deep-copying them into the shared `~/.agents/skills`, and adds `cleanupLegacyCodexMirror()` to remove the old copies. That eliminates the `node_modules` duplication this PR targeted, at the source, so this change is no longer needed.

Thanks for the reviews here.

PR #2742, opened 2026-06-17
feat(agents): applications landing polish + chat dedup fix

dmarticus:
Polish on top of #2700 — focused on the Applications landing IA, the Agent Builder dock's distinct identity, and a real bug in the chat preview's user-message dedup.

Stacked on top of `posthog-code/m3a-final` so the diff stays scoped. Merge order: #2700 → this.

### Landing polish (commit `7a2094ed9`)
- **Single entry point to the Agent Builder.** Dropped the contextual `New agent` / `Edit with AI` header CTA across every agents view; the always-on sparkles dock toggle is now the only entry point. Removes the "two near-identical gold buttons" confusion in the page header. Deleted `EditWithAIButton` + `agentBuilderActions`.
- **Tinted the dock** with a subtle amber accent (left border + faint header wash) so it reads as the meta-tool rather than another task chat. Mirrors the existing `SparkleIcon` accent.
- **Reordered the Applications landing.** Agents list moves to the top, sectioned **LIVE** vs **DRAFTS** (drafts dimmer, count in the section label, hidden when empty). Activity rollup hides when `analytics.empty`. Operational strip drops the live-count chip — the always-visible Live now panel below owns that signal — and keeps only the pending-approvals link, amber-tinted when non-zero.
- **Tightened the Applications shell copy** to mirror Scouts: "Deployed agents triggered by chat, Slack, webhooks, or cron — preview, approve, and edit with the Agent Builder." (single line at the current width).

### Chat dedup fix (commit `c5cda4811`)
The optimistic-seed → SSE-echo dedup in `createAgentChatMapper` used strict equality on `pendingOptimistic[0]`, which broke in three real failure modes — each one surfaced the same way: the user's just-sent bubble rendered twice.

| Failure mode | Symptom | Fix |
| --- | --- | --- |
| Whitespace asymmetry | Runner echoes back with trailing `\n` or padding around the context envelope; strict equality misses by one character. | Compare `text.trim()` to `pending.trim()`. |
| Out-of-order echo | Rapid back-to-back sends land out of order; `[0]` check only inspects the head. | `findIndex` over the whole queue. |
| Runner re-emit | Runner emits the same `user_message` twice; second arrival has nothing in `pendingOptimistic` to swallow it. | Track every rendered user text in a `seenUserTexts` set; drop later duplicates. |

Four new tests in `sessionEventToAcp.test.ts` covering each failure mode. 15/15 pass; `pnpm --filter @posthog/ui typecheck` clean.

## Test plan

- [ ] Apps list looks right: LIVE/DRAFTS sections, drafts dimmer, hidden when empty.
- [ ] Open the dock; it has the amber accent and no longer competes with a page-header CTA.
- [ ] Pending-approvals link still routes to `/code/agents/applications/approvals` and tints amber when non-zero.
- [ ] Live now panel renders its own empty state when the fleet is quiet.
- [ ] Send a message in the Agent Builder dock as a fresh session — user bubble appears once (not twice).
- [ ] Send a message in a per-agent Chat preview tab — same.
- [ ] Send two messages back-to-back fast — both render exactly once each.


packages/ui/src/features/agent-applications/agent-builder/EditWithAIButton.tsx
benjackwhite, line 1: Personally i think this is a regresssion :( Having the clear prompt to push to the agent builder makes a lot of sense. Otherwise it isn't at all clear that this is even an option.

The sparkle button isn't obvious enough imo and i think this change isn't worth it

benjackwhite: I'll PR against it with what i think is preferable

benjackwhite (changes requested):

PR #2755, opened 2026-06-18
Stop mirroring bundled skills into the shared ~/.agents/skills dir

k11kirky:
## Why

A user reported that PostHog Code "installed around 100 skills into the global `.agents` directory without notifying me" and that their tokens in other coding agents started burning faster. **This was accurate.**

`~/.agents/skills` is a *shared, cross-agent* skills directory — other tools (e.g. Codex) read it too. `syncCodexSkills` was copying our **entire bundled PostHog catalog (~100 skills)** into it:

- on **every app startup** (`PosthogPluginService.initialize`), and
- at the end of **every 30-minute skills update** (`UpdateSkillsSaga`),

unconditionally (`rm` + `cp` per skill), with **no collision guard, no state tracking, no cleanup, and no notification**. That silently bloated users' global skills dir and inflated *other* agents' context on every session — the direct cause of the "tokens burning faster" report.

This is distinct from the explicit marketplace/team "Install" flows (gated on a click, write to `~/.claude/skills`, toast on success) — those are fine.

## What

Remove `syncCodexSkills` entirely. Now **only the user's own skills** (`~/.claude/skills`) are mirrored out, via the existing collision-safe `mirrorUserSkillsToCodex`:
- tracks the names it wrote in `.posthog-mirror.json`,
- never overwrites a skill it didn't create,
- removes mirrors when the source skill is deleted.

The bundled catalog stays in the app-private runtime plugin dir and is loaded only by PostHog Code's own sessions — it never needed to be in the shared dir.

### Files
- `posthog-plugin.ts` — drop the `syncCodexSkills` import + startup call; remove now-unused saga inputs (`pluginPath`, `codexSkillsDir`).
- `update-skills-saga.ts` — delete the `syncCodexSkills` function, the `sync-codex-skills` saga step, and the unused input fields.
- `posthog-plugin.test.ts` — remove the obsolete `syncCodexSkills` tests and the unused `CODEX_SKILLS_DIR` constant.
- `README.md` — rewrite the "Codex Sync" section as "Codex Mirror", documenting that only user skills are mirrored, and why.

## Verification
- `codex-mirror` + `posthog-plugin` test suites: **31/31 pass**.
- Biome lint on changed files: clean.
- Typecheck: no errors in changed files.

## Follow-up (not in this PR)
Users on an affected build still have ~100 bundled skills sitting in `~/.agents/skills`. This change stops new writes but does **not** retroactively clean up already-mirrored bundled skills — they were never tracked in `.posthog-mirror.json`, so we can't safely distinguish them from the user's own. A one-time cleanup migration could be added if we want to claw those back.

### DIFF
--- a/packages/workspace-server/src/services/posthog-plugin/README.md
+++ b/packages/workspace-server/src/services/posthog-plugin/README.md
@@ -40,7 +40,7 @@ Vite watches `plugins/posthog/` (and `local-skills/` in dev) for hot-reload.
 **On startup:**
 1. Creates `{userData}/plugins/posthog/` (the runtime plugin dir)
 2. Assembles it: copies `plugin.json` from bundled, merges bundled skills + any previously-downloaded remote skills
-3. Syncs skills to `$HOME/.agents/skills/` for Codex
+3. Mirrors the user's own skills out to `$HOME/.agents/skills/` for Codex (see below)
 4. Starts a 30-minute interval timer
 5. Kicks off the first async download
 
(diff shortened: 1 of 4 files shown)


k11kirky:
Superseded by #2760, which contains this commit plus the full per-session design (zero global writes, one-time cleanup of previously-mirrored skills, private CODEX_HOME for Codex parity, and Claude-session loading of the user's Codex skills). Closing in favor of #2760.

PR #2762, opened 2026-06-18
feat(code-review): mark files as viewed in diff review

MattPua:
## Problem

When reviewing a diff there's no way to track which files you've already looked at, no signal when a file changes after you've reviewed it, and nothing to get reviewed files out of the way.

## Changes

https://github.com/user-attachments/assets/43b76744-ad8b-44ab-9c50-7ba673bb42e9

- Adds a **Read** checkbox to each file header in the diff/review viewer
- Marking a file read collapses it; unmarking expands it (like GitHub)
- Already-read files auto-collapse when you reopen the panel
- If a file changes after you marked it read, it shows a **Changed** badge (and stays expanded) so you know to look again
- Shows a **read count** (`read / total`) next to the file count
- Read state is saved per task and survives restarts, across local, branch, PR, and cloud diffs
- Read state is cleared when a task is archived or its PR is merged, with a size cap so storage stays bounded

## How did you test this?

- Unit tests for the read-state store (mark/unmark, clearing, size-cap eviction) and the change-detection signatures
- `pnpm --filter @posthog/ui typecheck` and `biome lint` pass on the changed files (the one remaining typecheck error is a pre-existing, unrelated issue in `canvas/WebsiteLayout.tsx`)
- Not yet exercised in the running app

## Automatic notifications

- [ ] Publish to changelog?

---


jonathanlab (approved):

jonathanlab (changes requested):
Thank you! Some feedback here:

- The use of signature caching, content hashing, and ref usage here feels excessive here. Are we sure there's not a cleaner way to do this? Can't we just calculate the fingerprint once on file load/change instead of recalculating it on every render/draw?

- Let's go with "Viewed" instead of "Read", I think this is convention across most reviewing platforms and what people are used to.

- I think the button is also quite hard to read, could we either make it clearer that this is a button, similar to GitHub's UI, or at least make the font size bigger

Image

MattPua:
CleanShot 2026-07-02 at 13 50 50@2x

CleanShot 2026-07-02 at 14 06 19@2x

MattPua:
@jonathanlab made a dumber version of caching without refs to tweak it

adboio (approved):
some small feedback:

- in the full page diff panel, the file tree should show some indication that the file is viewed
- (opinion) clicking a file in the file tree that is marked as viewed should still scroll to it, but not unfurl it

otherwise looks good, as far as i can tell @jonathanlab 's feedback was well-addressed, good stuff :saluting_face:

jonathanlab (approved):

PR #2765, opened 2026-06-18
fix(agent): pr detection

joshsny:
pr detection was not accurate, this fixes it by checking the actual PR state


rafaeelaudibert:
There are conflict markers commited here

andrewm4894 (approved):

PR #2783, opened 2026-06-19
feat(notifications): record a custom completion sound

andrewm4894:
## Problem

The notification sound that plays when the agent finishes a task or needs your input is limited to a fixed list of built-in sounds. Users wanted to add their own — ideally recording one on the fly right in the app.

## Changes

Adds a **Custom (record)** option to the Sound effect picker in General settings. Selecting it reveals a small recorder that captures up to 5 seconds from the microphone via `MediaRecorder`, stores it as a data URL in the persisted settings store, and plays it like any other completion sound (Test button, volume slider, and the OS-notification silencing all work with it).

- `playCompletionSound` takes an optional custom data URL and plays it when the sound is `"custom"`.
- New `CustomSoundRecorder` component: Record / Re-record / Stop / Clear, with mic-permission error handling.
- Threads `customCompletionSound` through the notification settings provider so the agent-completion/permission notifications use it.
- macOS: adds the `audio-input` entitlement and `NSMicrophoneUsageDescription` so recording works in the packaged, sandboxed app.

## How did you test this?

- `pnpm --filter @posthog/ui exec vitest run src/features/notifications/notifications.test.ts` — 14 pass, including new cases for the custom sound (plays the recorded URL + silences the OS notification; not silenced when nothing is recorded yet).
- Typechecked the changed files and ran Biome on all touched files (clean).
- Did not manually drive the packaged app / live mic recording.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1781887457202479)*

### DIFF
--- a/apps/code/build/entitlements.mac.plist
+++ b/apps/code/build/entitlements.mac.plist
@@ -13,5 +13,9 @@
 	<!-- Enable JIT compilation for V8 -->
 	<key>com.apple.security.cs.allow-jit</key>
 	<true/>
+
+	<!-- Microphone access for recording a custom notification sound -->
+	<key>com.apple.security.device.audio-input</key>
+	<true/>
 </dict>
 </plist>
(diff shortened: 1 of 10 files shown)


charlesvien:
https://github.com/PostHog/code/pull/2138

PR #2784, opened 2026-06-19
fix(canvas): open channel-filed task notifications in /website

raquelmsmith:
## Problem

When a task needs action, the user gets a desktop notification. Clicking it **always** opened the task at `/code/tasks/$taskId` — even when the task was filed to a Project Bluebird channel, where it should open in the channel-organized view at `/website/$channelId/tasks/$taskId` (keeping the channels chrome), exactly like the command palette already does.

The notification click and `posthog-code://task/<id>` deep links both funnel through `useTaskDeepLink`, which called `openTask(task)` and unconditionally navigated to `/code`. Nothing consulted channel membership.

## Change

- **`useOpenTask.ts`** — `openTask` now accepts an optional `{ channelId }`. When present it navigates via `navigateToChannelTask` (→ `/website`); otherwise `/code` as before. Every other side effect (detail-cache seed, analytics, `ensureWorkspaceForTask`) is unchanged, so channel tasks still get workspace provisioning. Existing single-arg callers are unaffected.
- **`useTaskDeepLink.ts`** — resolves channel membership (gated behind `PROJECT_BLUEBIRD_FLAG`) via `useChannels` + `useTaskChannelMap`, mirrored into a ref so the stable `handleOpenTask` callback reads the current map without tearing down the `onOpenTask` subscription on each 30s channel poll.
- **`CommandMenu.tsx`** — collapsed its duplicated `if/else` branch onto the same parameterized `openTask`, so palette-opened channel tasks also get workspace binding.

This fixes both notification clicks and deep links, since they share the path.

## Known limitation

Cold start (app launched by the click before the channel map has loaded) falls back to `/code`. Warm clicks — the dominant case for "needs action" notifications — route correctly.

## Verification

- `pnpm --filter @posthog/ui test run …/useTaskDeepLink.test.tsx` — 4/4 pass (added cases: filed task → channel route; flag-off → `/code`).
- `pnpm --filter @posthog/ui typecheck` — clean.
- `biome lint` on changed files — clean.

### DIFF
--- a/packages/ui/src/features/command/CommandMenu.tsx
+++ b/packages/ui/src/features/command/CommandMenu.tsx
@@ -31,10 +31,7 @@ import { TaskIcon } from "@posthog/ui/features/sidebar/components/items/TaskIcon
 import { useSidebarStore } from "@posthog/ui/features/sidebar/sidebarStore";
 import { useTaskPrStatus } from "@posthog/ui/features/sidebar/useTaskPrStatus";
 import { useTasks } from "@posthog/ui/features/tasks/useTasks";
-import {
-  navigateToChannel,
-  navigateToChannelTask,
-} from "@posthog/ui/router/navigationBridge";
+import { navigateToChannel } from "@posthog/ui/router/navigationBridge";
 import { useAppView } from "@posthog/ui/router/useAppView";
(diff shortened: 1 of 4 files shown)


adamleithp (approved):
LGTM

raquelmsmith:
tested locally, works

PR #2788, opened 2026-06-19
chore(analytics): adopt posthog-js defaults 2026-05-30

pauldambra:
## Problem

The browser SDK was running without the `defaults` config flag, so it stayed on legacy posthog-js behavior instead of the current recommended defaults.

## Changes

- Set `defaults: "2026-05-30"` on the browser `posthog.init` call in `posthogAnalyticsImpl.ts`, opting into the latest posthog-js default behaviors, with a comment documenting that it's a named epoch (not a date) and requires posthog-js >= 1.378.0.
- The `2026-05-30` defaults value was introduced in posthog-js 1.378.0, so bump the dependency from `^1.283.0` to `^1.378.0` (lockfile now resolves to 1.386.6).
- Set `capture_pageview: false`. The epoch flips `capture_pageview` to `"history_change"`, but this app routes via `createHashHistory()` — the route lives in the URL hash and `$pathname` is identical for every screen, so automatic pageviews would collapse all routes into one and corrupt route-level analytics. Opt out and rely on explicit instrumentation. (If route-level pageviews are wanted later, a `before_send` that rewrites `$pathname`/`$current_url` from the hash is the path.)

Capture delta from adopting the epoch (for the record): `session_recording.strictMinimumDuration`, `external_scripts_inject_target: head`, `persistence_save_debounce_ms: 250`, `split_storage`, `detect_google_search_app`, and content-masking ignorelist tweaks. No new autocapture/network instrumentation; automatic pageviews explicitly disabled.

## How did you test this?

- `pnpm --filter @posthog/ui typecheck` passes (confirms `defaults: "2026-05-30"` and `capture_pageview: false` are valid config in the bumped SDK).
- ⚠️ Not yet verified at runtime: that session recording still starts after the ~95-minor posthog-js bump (import paths unchanged, but a manual smoke test of the Electron app is worth doing before merge).

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1781898921688269?thread_ts=1781898921.688269&cid=C09G8Q32R6F)*


packages/ui/src/shell/posthogAnalyticsImpl.ts
```diff
@@ -55,6 +55,7 @@ export function initializePostHog(sessionId?: string) {
   }
 
   posthog.init(apiKey, {
+    defaults: "2026-05-30",
```
pauldambra, line 58: > [!NOTE]

`defaults: "2026-05-30"` is doing more than it looks. It's an opaque, date-shaped *magic string* that's actually a named posthog-js defaults epoch — and it's silently coupled to the `^1.378.0` floor you bumped in package.json (that value doesn't exist below 1.378). The epoch also flips a bag of options (historically autocapture, exception capture, web vitals, dead-clicks, tracing headers…), some of which overlap with what you set explicitly right below.

Two asks: (1) confirm the precedence — do the explicit options below win over the epoch, or does the epoch win? It matters for `capture_exceptions`: in dev you rely on the explicit `false` to keep exceptions off, so if the epoch took precedence you'd quietly enable dev exception capture. (2) Drop a one-line comment naming what the epoch pins and the version it requires, e.g. `// pins posthog-js default config to the 2026-05-30 epoch; requires posthog-js >= 1.378.0; explicit options below intentionally override it`. Saves the next person a trip through the changelog.

packages/ui/src/shell/posthogAnalyticsImpl.ts
```diff
@@ -55,6 +55,7 @@ export function initializePostHog(sessionId?: string) {
...
     api_host: apiHost,
     ui_host: uiHost,
     disable_session_recording: false,
```
pauldambra, line 67: > [!NOTE]

Recording is now asserted three ways: the epoch defaults, `disable_session_recording: false`, and the `loaded` callback force-calling `startSessionRecording()`. No behaviour bug (explicit options come after `defaults` and win), but a redundant `disable_session_recording: false` reads as load-bearing when it may not be. Worth checking what the 2026-05-30 epoch actually sets for `disable_session_recording` / `capture_exceptions` / recording startup, then either prune what it already does or keep an "overriding on purpose" note so the next reader can tell which lines are intentional.

packages/ui/src/shell/posthogAnalyticsImpl.ts
pauldambra, line 58: > [!NOTE]

**[paul]** 🟡 MEDIUM

Knight-Capital-shaped concern: this flips new capture behaviour on for every user at once with no flag, and the stated motivation is "the app feels slow". New defaults can *add* autocapture/network instrumentation — i.e. more overhead — which would be a funny own-goal given the goal is performance. Not blocking, but could you note the expected before/after capture diff in the PR description, and consider whether the epoch is worth gating behind a flag (or is that overkill for analytics config)? Genuinely asking.

packages/ui/package.json
```diff
@@ -91,7 +91,7 @@
     "fzf": "^0.5.2",
     "inversify": "catalog:",
     "lucide-react": "^1.7.0",
-    "posthog-js": "^1.283.0",
+    "posthog-js": "^1.378.0",
```
pauldambra, line 94: > [!NOTE]

**[paul]** 🟢 LOW

~95 minor versions in one jump (1.283 → 1.386, lock 1.386.6). Silly question — did anything across those minors touch the recorder bundle pinned so specifically at the top of `posthogAnalyticsImpl.ts` (`module.full.no-external` + `posthog-recorder`)? That import is load-bearing and version-sensitive. Worth running the app and confirming a recording still actually lands before merge.

pauldambra (commented):
> [!NOTE]

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — not available on disk)

No blockers. The whole change is one `defaults: "2026-05-30"` epoch flag plus the matching `posthog-js` floor bump — minimal and correct. The asks are about *documenting and verifying* what the new defaults epoch turns on, not about correctness.

### Key findings

**🟡 MEDIUM**
- The `defaults: "2026-05-30"` magic string is an opaque posthog-js *epoch*, silently coupled to the `^1.378.0` floor, and overlaps options set explicitly below it. Confirm precedence (esp. `capture_exceptions` in dev) and add a one-line comment naming what it pins + the version it needs.
- Behaviour-change-to-everyone with no flag, while the motivation is "feels slow" — new defaults may *add* capture overhead. Note the expected before/after capture diff; consider whether a flag is warranted.

**🟢 LOW**
- Recording is now asserted three ways (epoch + `disable_session_recording: false` + `startSessionRecording()` in `loaded`); prune or annotate the redundant ones.
- ~95-minor jump (1.283 → 1.386.6): sanity-check the specifically-pinned recorder bundle still starts recording on the running app.

### Convergence

Both paul and xp independently flagged (a) the opaque epoch string / version coupling and (b) that the epoch may make neighboring explicit options redundant — highest-confidence items.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 👤 paul | approve-with-comments: document the epoch interaction, treat as a behaviour-change rollout, verify the recorder still works after the big version jump |
| 📐 xp | clean minimal opt-in, no behaviour bug; two expressiveness nits — opaque epoch string and possibly-redundant neighboring options |
| 🛡 security-audit | 0 findings — privacy-neutral config flag + first-party SDK bump; defaults epoch does not loosen input masking |

---

packages/ui/src/shell/posthogAnalyticsImpl.ts
pauldambra: > [!NOTE]

Fixed in `82d6675` — added a comment above the line documenting that `2026-05-30` is a named posthog-js defaults epoch (not a date), that it requires posthog-js >= 1.378.0 (kept in sync with package.json's floor), and that explicit options below intentionally override it. Confirmed the precedence against the posthog-js source: it merges the epoch first and the explicit config last, so explicit options win.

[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Good catch — verified. The 2026-05-30 epoch sets `capture_pageview: "history_change"` (`capture_pageview: !e || "2025-05-24" > e || "history_change"`), and this app routes via `createHashHistory()` (router.ts:29), so `$pathname` is identical across routes. Fixed in `d139813` by setting `capture_pageview: false` to opt out of automatic pageviews. Alternative if route-level pageviews are wanted later: a `before_send` that rewrites `$pathname`/`$current_url` from the hash, per the hash-routing tutorial you linked.

packages/ui/src/shell/posthogAnalyticsImpl.ts
pauldambra: > [!NOTE]

Checked what the epoch sets: its `session_recording` value is only `{strictMinimumDuration: true}` (for epochs >= 2025-11-30) and it does **not** set `disable_session_recording`. So `disable_session_recording: false` is not redundant with the epoch — keeping it (plus the explicit `startSessionRecording()` in `loaded`) as intentional, self-documenting config. Resolving as a nit.

packages/ui/src/shell/posthogAnalyticsImpl.ts
pauldambra: > [!NOTE]

The main overhead/own-goal concern was the epoch flipping `capture_pageview` to `history_change` (a pageview per route change) — now opted out in `d139813`. Before/after capture delta from adopting the epoch, for the PR record: `capture_pageview` history_change (now disabled), `session_recording.strictMinimumDuration`, `external_scripts_inject_target: head`, `persistence_save_debounce_ms: 250`, `split_storage`, `detect_google_search_app`, plus content-masking ignorelist tweaks — no new autocapture/network instrumentation enabled. Given that, gating analytics config behind a flag is overkill here. Resolving.

packages/ui/package.json
pauldambra: > [!NOTE]

Deferring to a human: the recorder import paths are unchanged (`module.full.no-external` + `posthog-recorder`) and typecheck passes, but confirming a recording actually lands needs a runtime smoke test of the Electron app, which I can't run here. Flagging for a manual check before merge.

packages/ui/src/shell/posthogAnalyticsImpl.ts
```diff
@@ -55,8 +55,20 @@ export function initializePostHog(sessionId?: string) {
   }
 
   posthog.init(apiKey, {
+    // Pins posthog-js' baseline config to the 2026-05-30 defaults epoch (a named
+    // bundle of recommended defaults, not a date anything happens on). This value
+    // only exists in posthog-js >= 1.378.0 — keep package.json's floor in sync.
+    // posthog-js merges the epoch first and the explicit options below last, so
+    // every option set here deliberately overrides the epoch.
+    defaults: "2026-05-30",
```
pauldambra:
```suggestion
    defaults: "2026-05-30",
```

charlesvien (approved):

PR #2798, opened 2026-06-20
perf(renderer): stop refetching every query on window focus

pauldambra:
## Problem

Returning to PostHog Code (alt-tab back into the window) froze the UI: the renderer pegged the main thread at 100%+ while the main process sat idle at ~2%.

`apps/code/src/renderer/utils/queryClient.ts` set `refetchOnWindowFocus: true` globally, and a `focusManager` listener deliberately maps window `focus`/`blur` to refetch (Electron renderers stay "visible" on app-switch, so `visibilitychange` alone never fires). So **every active query refetched the instant you returned to the app** — a synchronous refetch-and-re-render burst, worst after being away >5min when every query is also stale.

## Change

- Default `refetchOnWindowFocus` to `false` in the renderer query client. This kills the storm; data still refreshes via tRPC subscriptions and the file watcher, and the inbox hooks already set `false` per-query.
- Correct the `focusManager` listener comment: it is **not** removed (it still gates `refetchInterval` background-poll pausing, which TanStack ties to focus state), but its comment no longer claims to drive focus-refetch.

`currentUser` keeps its `refetchOnWindowFocus: "always"` override (unchanged from main) so the AI-approval gate still clears on return — that is one cheap query, not the storm.

## Measurement

Captured against the running dev app over CDP (React commit hook + `longtask` observer, focus synthesized to reproduce "return to app"):

| per focus event | before | after (global flip) | idle baseline |
| --- | --- | --- | --- |
| React commits | 12 | 2-4 | 4 / 5s |
| long tasks | 1 | 0 | 0 |
| main thread blocked | 53-73ms | 0ms | 0ms |
| commit work | ~125ms | 11-28ms | - |

(With `currentUser` keeping `"always"`, expect ~1 extra query refetch + a small commit per focus — still far below the original 12-commit storm.)

## Follow-ups (separate PRs)

- #2799 — debounce/coalesce file-watcher query invalidations.
- Streaming render perf (CodePreview editor recreation, tool-view memoization) — pending a live streaming trace before touching render code.

### DIFF
--- a/apps/code/src/renderer/utils/queryClient.test.ts
+++ b/apps/code/src/renderer/utils/queryClient.test.ts
@@ -63,6 +63,14 @@ describe("getCachedTask", () => {
   });
 });
 
+describe("defaultOptions", () => {
+  it("defaults refetchOnWindowFocus to false so returning to the app does not refetch every query", () => {
+    expect(queryClient.getDefaultOptions().queries?.refetchOnWindowFocus).toBe(
+      false,
+    );
+  });
+});
+
 describe("focusManager", () => {
   it("flips focus state on window focus/blur events", () => {
     window.dispatchEvent(new Event("blur"));
--- a/apps/code/src/renderer/utils/queryClient.ts
+++ b/apps/code/src/renderer/utils/queryClient.ts
@@ -10,10 +10,12 @@ export const queryClient = new QueryClient({
   },
 });
 
-// Electron renderers stay visible when the BrowserWindow loses OS focus, so
-// `document.visibilitychange` (TanStack's default signal) never fires on
-// app-switch. Listen to window `focus`/`blur` as well so refetchOnWindowFocus
-// actually triggers when the user returns from an external browser.
+// Electron renderers stay "visible" when the BrowserWindow loses OS focus, so
+// `document.visibilitychange` (TanStack's default focus signal) never fires on
+// app-switch. Track window `focus`/`blur` so `focusManager` knows when the app
+// is backgrounded and pauses `refetchInterval` polling for the queries that
+// don't set `refetchIntervalInBackground`. Refetch-on-focus is off (see
+// `defaultOptions` above); this listener exists only for the polling gate.
 focusManager.setEventListener((handleFocus) => {
   if (typeof window === "undefined") return;
 
--- a/apps/code/src/renderer/utils/queryClient.ts
+++ b/apps/code/src/renderer/utils/queryClient.ts
@@ -5,7 +5,7 @@ export const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       staleTime: 1000 * 60 * 5,
-      refetchOnWindowFocus: true,
+      refetchOnWindowFocus: false,
     },
   },
 });
--- a/packages/ui/src/shell/App.tsx
+++ b/packages/ui/src/shell/App.tsx
@@ -59,7 +59,6 @@ function App() {
     client: authenticatedClient,
     enabled:
       isAuthenticated && hasCompletedOnboarding && hasCodeAccess === true,
-    refetchOnWindowFocus: "always",
   });
   const currentOrg = currentUser?.organization;
   const needsAiApproval =
--- a/packages/ui/src/shell/App.tsx
+++ b/packages/ui/src/shell/App.tsx
@@ -59,6 +59,7 @@ function App() {
     client: authenticatedClient,
     enabled:
       isAuthenticated && hasCompletedOnboarding && hasCodeAccess === true,
+    refetchOnWindowFocus: "always",
   });
   const currentOrg = currentUser?.organization;
   const needsAiApproval =


pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team, paul-reviewer, xp-reviewer, security-audit (all on Opus; Fable was unavailable)

The core change is sound and well-measured (per focus event: React commits 12 -> 2-4, long tasks 1 -> 0, main-thread block 53-73ms -> 0ms). No blocking issues. Two things surfaced; one is being fixed in a follow-up commit, one is a judgement call for the author.

### Convergent finding (qa-team + paul + xp) — and an important correction

All three independently flagged the `focusManager.setEventListener` block in `queryClient.ts` as "now dead code" because its comment says it exists "so refetchOnWindowFocus actually triggers," and this PR disables focus-refetch everywhere.

**On verification this is only half right — do NOT delete the listener.** It is still load-bearing: many queries use `refetchInterval` *without* `refetchIntervalInBackground` (`useTasks`, `useGitQueries`, `useChannels`, `usePrDetails`, `useHomeSnapshot`, `useInboxAllReports` which sets it explicitly `false`, ...). TanStack pauses interval polling while the window is unfocused, and `focusManager` decides "unfocused." In Electron `visibilitychange` doesn't fire on app-switch, so without this custom listener `focusManager` would treat the window as *always focused* and all that polling would keep running while the app is backgrounded — a CPU/battery regression and the opposite of this PR's intent.

**Resolution:** keep the listener and its test; correct the now-misleading comment to state its real (and now sole) purpose — pausing `refetchInterval` polling while the window is blurred. Applying this in a follow-up commit on this branch.

### Judgement call for the author (deferred) — `useCurrentUser` freshness

qa-team (MEDIUM): removing the `refetchOnWindowFocus: "always"` override leaves `currentUser` (org / `is_ai_data_processing_approved` / `isAdmin`) refreshed only on mount + the 5min `staleTime`, with no subscription/watcher invalidating it. So org/approval state can stay stale up to 5min after returning to the app.

security-audit independently cleared the safety of this: the AI-consent gate **fails closed** under staleness (stale data keeps the wall up, never grants access), and `isAdmin` drives only UI affordances behind server-side enforcement. So it's a **UX freshness** question, not a security one. Options: accept the trade-off, or re-add `refetchOnWindowFocus: true` (not `"always"`) to just that one query. Left to the author.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Sound fix; flag currentUser staleness + the focusManager block. MEDIUM. |
| 👤 paul | Right change, tiny diff, lovely numbers; clean up the focusManager block. Approve. |
| 📐 xp | Correct and minimal; the disabled feature's plumbing shouldn't be left pretending it works. |
| 🛡 security-audit | No findings. Stale org state fails closed; no access decision fails open. |

---

pauldambra:
Addressed the `useCurrentUser` regression (raised by @stamphog and qa-team) in 3862ced0d.

Restored `refetchOnWindowFocus: "always"` on the `useCurrentUser` query so the AI-approval gate still clears on return after an admin approves. The global default stays `false`, so the focus-refetch storm this PR fixes is gone — only this one query refreshes on focus, which is the deliberate freshness it always had.

Net effect: this PR is now just the global `refetchOnWindowFocus: false` flip plus a corrected `focusManager` comment. The `App.tsx` diff is gone (currentUser is unchanged from main).

pauldambra:
> [!NOTE]
>
> Clean-room re-review (reviewers given only the diff, no prior-thread context): qa-team, paul-reviewer, xp-reviewer, security-audit — all on Opus.

## Verdict: ✅ APPROVE (unanimous, low risk)

All four independent reviewers approved. One test added in response (a63a647d4): pins the `refetchOnWindowFocus: false` default so a flip back to `true` can't silently restore the focus-refetch storm.

### Rebuttal to @stamphog's "AI-approval gate won't clear" review

stamphog flagged this as an unaddressed regression. **It is addressed — this is a false positive from reading the diff alone.** `useCurrentUser` in `App.tsx:62` keeps `refetchOnWindowFocus: "always"`, which overrides the global default, so the AI-approval gate still refetches on focus and clears on return. That override is unchanged from `main`, so it doesn't appear in this PR's diff — but it's in the file and load-bearing. Two reviewers verified this independently against the real branch:
- **security-audit:** "the only permission/org source is `useCurrentUser` ... sets `refetchOnWindowFocus: \"always\"` explicitly (App.tsx:62), which overrides the global default. ... security-relevant freshness is preserved."
- **qa-team:** "the single focus-sensitive query (the AI data-processing approval gate) keeps an explicit `\"always\"` override."

Re-applying the label to re-trigger a review against `a63a647d4`.

### Non-blocking nits (deferred to author)
- **paul + xp:** the `App.tsx:62` override is now a lonely line a future cleanup might "simplify" away, re-breaking the gate. Protect it with a call-site note or an `AiApprovalScreen` test. (Note vs test left to you — a comment is against your usual no-comments rule; a test is the comment-free option.)
- **qa-team + paul:** low-volatility git reads (`getPrStatus`, branch metadata) and the sidebar PR badge lose focus-refresh; covered by 30-60s sibling polls + invalidation. Accepted tradeoff of the global flip.

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | APPROVE, low risk. Listener correctly retained for the polling gate; gate keeps its override. |
| 👤 paul | Ship it. Nits: pin the default (done), protect the override, gut-check other focus-only queries. |
| 📐 xp | Clean, simplicity rules balanced well. One nit: call-site rationale for the override. |
| 🛡 security-audit | No findings. Permission/org freshness preserved by the explicit override + subscriptions. |

---

pauldambra:
Closing: investigation with a real-data repro (766 workspaces copied from a production instance) showed the UI freeze is caused by the non-virtualized sidebar task list (`TaskListView` renders every workspace via `.map()`) plus per-conversation work, blocking the main thread 1-4s on task load and scaling linearly with conversation count. Focus-refetch / file-watcher churn never produced a long task. This change is unrelated to the actual freeze, so closing to keep focus on the real fix (virtualize `TaskListView` + bound per-row work).

PR #2799, opened 2026-06-20
perf(file-watcher): coalesce query invalidations within a window

pauldambra:
## Problem

While the agent writes files (or on a fetch/merge), the renderer churns. `useRepoFileWatcher` reacted to every file-watcher event **synchronously**, and a single `git-state-changed` event invalidates **10 query families** (`invalidateGitBranchQueries`) while `working-tree-changed` invalidates 4 more. A burst of changes therefore fired those repo-global invalidations once per event with no coalescing — each invalidation cascading into refetch + re-render work on the main thread.

This is Phase 2 of the renderer-performance pass (Phase 1: #2798, the window-focus refetch storm).

## Change

Batch watcher events into a fixed 250ms window:
- per-file invalidations dedupe by path within the window;
- the repo-global git invalidations collapse to a single call per flush.

A burst of N `git-state-changed` events now triggers **one** round of branch-query invalidation instead of N. The window is fixed (not reset on each event) so a continuous stream of changes still flushes every 250ms rather than starving until activity stops.

The coalescing logic is extracted into a pure helper, `createFileWatcherCoalescer`, so it is unit-testable without React; the hook becomes glue that feeds classified events in and performs the invalidations on flush.

## Measurement

The reduction is deterministic, so it's proven by a parameterised unit test (`fileWatcherCoalescer.test.ts`) rather than a load-dependent live count:

- 3 repeated `git-state-changed` -> **1** branch invalidation (was 3)
- repeated `working-tree-changed` -> **1** invalidation
- repeated `fileChanged` on the same path -> deduped to one invalidation
- a mixed burst (files + git-state + working-tree + delete) -> **one** flush
- a continuous stream flushes once per window (no starvation)
- nothing flushes after `dispose()`

## Behaviour notes

- Invalidations now land up to 250ms later. File-watcher events are already async/external, so this is imperceptible and well within normal query staleness.
- `file-deleted` tab-closing is also batched into the window (a deleted file's tab closes ~250ms later).

### DIFF
--- a/packages/ui/src/features/file-watcher/fileWatcherCoalescer.test.ts
+++ b/packages/ui/src/features/file-watcher/fileWatcherCoalescer.test.ts
@@ -109,12 +109,25 @@ describe("createFileWatcherCoalescer", () => {
     expect(calls.invalidateGitBranch).toBe(2);
   });
 
-  it("does not flush after dispose", () => {
+  it("flushes pending work on dispose so a teardown mid-window drops nothing", () => {
     const { actions, calls } = makeActions();
     const c = createFileWatcherCoalescer(actions, 250);
     c.gitStateChanged();
+    c.fileChanged("a.ts");
+    c.dispose();
+    expect(calls.invalidateGitBranch).toBe(1);
+    expect(calls.invalidateFile).toEqual(["a.ts"]);
+    vi.advanceTimersByTime(1000);
+    expect(calls.invalidateGitBranch).toBe(1);
+    expect(calls.invalidateFile).toEqual(["a.ts"]);
+  });
+
+  it("dispose with no pending work does nothing", () => {
+    const { actions, calls } = makeActions();
+    const c = createFileWatcherCoalescer(actions, 250);
     c.dispose();
     vi.advanceTimersByTime(1000);
     expect(calls.invalidateGitBranch).toBe(0);
+    expect(calls.invalidateFile).toEqual([]);
   });
 });
--- a/packages/ui/src/features/file-watcher/fileWatcherCoalescer.test.ts
+++ b/packages/ui/src/features/file-watcher/fileWatcherCoalescer.test.ts
@@ -1,0 +1,120 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+  createFileWatcherCoalescer,
+  type FileWatcherActions,
+  type FileWatcherCoalescer,
+} from "./fileWatcherCoalescer";
+
+function makeActions() {
+  const calls = {
+    invalidateFile: [] as string[],
+    closeTabsForFile: [] as string[],
+    invalidateGitBranch: 0,
+    invalidateGitWorkingTree: 0,
+  };
+  const actions: FileWatcherActions = {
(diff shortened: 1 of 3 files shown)


pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team, paul-reviewer, xp-reviewer, security-audit (all on Opus; Fable unavailable)

Clean, well-tested pure-helper refactor. One MEDIUM addressed in a follow-up commit; the rest are non-blocking preferences left to the author.

### Convergent MEDIUM (qa-team + paul + xp) — fixed in follow-up commit

`dispose()` cleared the pending timer **without flushing**. The coalescer is recreated (and disposed) whenever `repoPath`/`taskId`/`queryClient`/`closeTabsForFile`/`cacheKeys` change identity, and on unmount. So a burst buffered mid-window was silently dropped on a `taskId` change or an unmount-mid-burst, leaving git/file views stale — the exact failure the watcher exists to prevent. **Fix applied:** `dispose()` now flushes pending work before clearing. Safe across a `repoPath` switch because the dispose closure captures the old `repoPath`, so the flush invalidates the correct repo. Test updated to pin flush-on-teardown (it previously encoded the drop as intended) + a no-op-when-empty case.

### Convergent NIT (paul + xp) — fixed in follow-up commit

`invalidateFile` used `repoPath ?? ""` while its sibling git actions guard `if (repoPath)`. The `""` branch was unreachable (`onEvent` early-returns on `!repoPath`) and would have built a nonsense cache key. Now guarded consistently.

### Deferred to author (non-blocking preferences)

- **Leading edge (paul):** trailing-only window means a single hand-save waits the full 250ms before other views reflect it. Fine for agent bursts; consider leading+trailing if felt latency on lone saves matters.
- **Observability (paul):** a perf change with no signal of the actual coalescing ratio (events-in vs flushes-out). A debug-level count would confirm the win in prod.
- **Configurable window (paul):** 250ms is a hardcoded default; could be host-tunable.
- **Branch name (xp):** branch says "debounce" but it's a fixed-window batch — the PR title's "coalesce" is the accurate word. Cosmetic.
- **Extra test (xp, optional):** a case for two separate bursts within one window / the changed-vs-deleted same-path interaction.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Sound design, fixed-window is right; the dispose-drop was the one real bug. MEDIUM. |
| 👤 paul | Clean coalescer, loved the parameterised test; think about dispose dropping the buffer. Approve. |
| 📐 xp | Genuinely simple, abstraction earns its keep and resists over-DRYing; flush-on-dispose or justify. |
| 🛡 security-audit | No findings. Renderer-only, no new sink, timer lifetime correct. |

---

pauldambra:
Closing: investigation with a real-data repro (766 workspaces copied from a production instance) showed the UI freeze is caused by the non-virtualized sidebar task list (`TaskListView` renders every workspace via `.map()`) plus per-conversation work, blocking the main thread 1-4s on task load and scaling linearly with conversation count. Focus-refetch / file-watcher churn never produced a long task. This change is unrelated to the actual freeze, so closing to keep focus on the real fix (virtualize `TaskListView` + bound per-row work).

PR #2803, opened 2026-06-20
perf(sidebar): memoize TaskRow so navigation re-renders 2 rows, not all

pauldambra:
when i'm using ph-code it regularly freezes

i copied my local prod app database into the dev instance and had the robot poke at it via CDP

-----

## Problem

The sidebar task list froze the UI on every task load with a real-scale dataset. `TaskRow` was not memoized, and each call site passed fresh inline closures (`onClick={(e) => onTaskClick(task.id, e)}`, ...), so an `activeTaskId` change on navigation re-rendered **every** row — and each row re-runs two queries (`useWorkspace`, `useTaskPrStatus`) plus a store subscription.

This was found by loading a **production database (766 workspaces, 5.9 GB of conversation logs)** into a dev instance and CPU-profiling task loads over CDP. It blocked the renderer main thread for **1–3.3s on every task load** — a hard freeze with no loaders, because it's synchronous render work, not async fetching. (Earlier theories — focus-refetch storms, file-watcher churn — were measured and ruled out; they never produced a long task.)

## Change

Memoize `TaskRow` and pass it the parent's already-`useCallback`'d handlers, binding `task.id` inside the row. Now an `activeTaskId` change re-renders only the two rows whose `isActive` flipped, instead of all of them. `TaskListView`'s external props are unchanged.

## Measurement

Blocked main-thread time on task load, 766-workspace dataset, same CDP harness before/after:

| task | before | after |
| --- | --- | --- |
| 1fbb7d49 | 2742ms | **539ms** |
| 4ebc00bf | 2361ms | **515ms** |
| 7f55577b | 3282ms | **665ms** |
| 0a06a281 | 2079ms | **347ms** |
| 1f9f73a8 | 942ms | **176ms** |
| be480903 | 984ms | **384ms** |

The measured drop confirms `task` object identity is stable enough for `memo` to bite.

## Scope / follow-ups

This fixes the **sidebar** half of the freeze (every task load). A second, independent cause remains for **large cloud conversations** — `buildCloudEventSummary`/`getDiffStats` (`@posthog/core/task-detail/cloudToolChanges`) reprocess the entire session event log on every update (O(n) per tick) — which I'm tackling separately. Virtualizing the sidebar list would further cut the first-mount cost but is a larger change; memoization captures the recurring re-render freeze.

### DIFF
--- a/packages/ui/src/features/sidebar/components/TaskListView.test.tsx
+++ b/packages/ui/src/features/sidebar/components/TaskListView.test.tsx
@@ -1,0 +1,111 @@
+import type { TaskData } from "@posthog/core/sidebar/sidebarData.types";
+import { render } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { renderCounts } = vi.hoisted(() => {
+  globalThis.ResizeObserver ??= class {
+    observe() {}
+    unobserve() {}
+    disconnect() {}
+  };
+  return { renderCounts: new Map<string, number>() };
+});
+
+vi.mock("@posthog/ui/features/sidebar/components/items/TaskItem", () => ({
+  TaskItem: ({ taskId }: { taskId: string }) => {
+    renderCounts.set(taskId, (renderCounts.get(taskId) ?? 0) + 1);
+    return null;
+  },
+}));
+
+vi.mock("@posthog/ui/features/workspace/useWorkspace", () => ({
+  useWorkspace: () => undefined,
+}));
+vi.mock("@posthog/ui/features/sidebar/useTaskPrStatus", () => ({
+  useTaskPrStatus: () => ({ prState: null, hasDiff: false }),
+}));
+vi.mock("@posthog/ui/features/sidebar/archivingTasksStore", () => ({
+  useArchivingTasksStore: (
+    selector: (s: { archivingTaskIds: Set<string> }) => unknown,
+  ) => selector({ archivingTaskIds: new Set<string>() }),
+}));
+vi.mock("@posthog/ui/features/folders/useFolders", () => ({
+  useFolders: () => ({ folders: [] }),
+}));
+vi.mock("@posthog/ui/router/useAppView", () => ({
+  useAppView: () => ({ type: "task-detail" }),
+}));
+vi.mock("@posthog/ui/router/useOpenTask", () => ({ openTaskInput: vi.fn() }));
+vi.mock("@posthog/ui/assets/hedgehogs", () => ({ builderHog: "" }));
+vi.mock("@posthog/ui/features/sidebar/sidebarStore", () => {
+  const state = {
+    organizeMode: "recent",
+    sortMode: "updated",
+    collapsedSections: new Set<string>(),
+    toggleSection: () => {},
+    loadMoreHistory: () => {},
+    resetHistoryVisibleCount: () => {},
+    folderOrder: [] as string[],
+    reorderFolders: () => {},
+  };
+  return {
+    useSidebarStore: Object.assign(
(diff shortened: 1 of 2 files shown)


pauldambra:
> [!NOTE]
>
> Clean-room review (reviewers given only the diff): qa-team, paul-reviewer, xp-reviewer, security-audit — all on Opus.

## Verdict: 💬 APPROVE WITH NITS — one real follow-up

All four approved; behaviour is preserved (the `task.id` binding moved faithfully into `TaskRow`, `isPinned` threading and `onEditSubmit(task.id, task.title, ...)` intact), and `task` identity is stable across navigation (`useSidebarData`'s `taskData` memo excludes `activeView`).

### Convergent finding (qa-team + paul + xp + security note) — memo is fragile, not fully effective

`TaskRow` is now `React.memo`'d, but two of the seven handlers it receives — `handleTaskClick` (`SidebarMenu.tsx:164`) and `handleTaskContextMenu` (`:225`) — are **plain functions, not `useCallback`'d**, and both close over `taskMap` (`:71`), which is rebuilt as a fresh `new Map` every render. So on any `SidebarMenu` re-render the shallow prop compare fails for every row and all rows re-render.

**Why the measured 5x win is still real:** navigation does not re-render `SidebarMenu` itself (only `TaskListView` via the `activeTaskId` prop), so on the task-load/switch path the handlers keep identity and the memo holds — which is the freeze that was reported. The gap is `SidebarMenu`'s own re-renders (sidebar data refresh, its own state), where the memo is currently defeated.

**Complete fix (deferred — needs care, not a blind edit):** memoize `taskMap` with `useMemo(..., [allTasks])`, then `useCallback` both handlers. Note `handleTaskContextMenu` forward-references `handleTaskArchive`/`handleTaskTogglePin`/`handleArchivePrior` (defined later), so it must be reordered below them to avoid a temporal-dead-zone error in the deps array. Deferring to a focused follow-up rather than risk a stale-closure bug here.

### Minor notes (non-blocking)
- Each `TaskRow` still runs `useWorkspace` + `useTaskPrStatus` + an archiving-store selector per row — memo stops parent-driven re-renders but per-row subscriptions still fire independently (out of scope; virtualization would bound this).
- No render-count test pins "2 rows, not all", so a future unmemoized prop could silently re-defeat it.
- security-audit: **no findings** — `key={task.id}` + reading bound values from the `task` prop means no wrong-id-to-action hazard.

| Reviewer | Verdict |
| --- | --- |
| 🔍 qa-team | Approve w/ nits (MEDIUM): memoize taskMap + useCallback the two handlers. |
| 👤 paul | Right direction; confirm/stabilize the two handlers before relying on the win. |
| 📐 xp | Correct, minimal, duplication actually reduced; wrap the two handlers. |
| 🛡 security-audit | No findings. |

---

pauldambra:
Closing in favor of virtualization.

Verified on a production-scale dataset (766 workspaces) with a render-count instrument on the running app: even after memoizing TaskRow *and* stabilizing the two un-`useCallback`'d handlers, ~350 rows still re-render on navigation. The memo is defeated at the data layer, not the handler layer:

- `task` objects churn identity (`useSidebarData` rebuilds the task arrays/objects when `activeView` changes) — 325 rows get a new `task` prop on navigation.
- `onArchive`/`onEditSubmit` also churn (they depend on `allSidebarTasks`, rebuilt from those same arrays).

Making the memo hold would require structural-sharing stable task identities through `useSidebarData` *and* stabilizing every handler dep — a fragile, multi-layer refactor. Virtualizing the list (render only visible rows via `@tanstack/react-virtual`) makes identity churn irrelevant and is the robust fix. Pursuing that instead.

PR #2804, opened 2026-06-20
perf(task-detail): cache cloud diff stats by content identity

pauldambra:
## Problem

Opening a large **cloud** task froze the renderer (100% CPU, no loaders — synchronous work). `extractCloudToolChangedFiles` line-diffs every changed file via `getDiffStats`, and it re-runs whenever the session `events` array changes identity. A big session log loads/streams in chunks, so **every changed file was re-diffed on every chunk** — O(files × updates).

Found by loading a production database (766 workspaces, 5.9 GB of logs, biggest single conversation 449 MB / 203k events) into a dev instance and CPU-profiling task loads over CDP. `getDiffStats` was the single biggest frame — **28% / ~2,088ms** on the worst load.

## Change

The diff content objects are stable across these recomputes (they're the session events' own objects), so cache `getDiffStats` results in a `WeakMap` keyed on the diff object. Repeated recomputes become O(new diffs) instead of O(all files). Output is unchanged — the cache only avoids recompute.

## Measurement

Same CDP profile, before/after:
- `getDiffStats`: **~2,088ms (28% of the worst load) → ~0** (dropped out of the hot path entirely).
- A representative large cloud task: **4,101ms → 1,282ms** blocked main-thread time on load.

A parameterised test pins the diff-stat values (added/removed across new/modified/removal) and that reusing the same diff object returns identical results.

## Scope

This is the **cloud-conversation** half of the freeze. The **sidebar** half (re-rendering all task rows on navigation) is a separate PR (memoizing `TaskRow`). Residual cost on the most extreme 449 MB task is now diffuse (log parsing, file-search indexing) with no single dominant frame — a possible future follow-up is incrementalizing `buildCloudEventSummary`, but the dominant freeze is resolved.

### DIFF
--- a/packages/core/src/task-detail/cloudToolChanges.test.ts
+++ b/packages/core/src/task-detail/cloudToolChanges.test.ts
@@ -62,6 +62,61 @@ describe("extractCloudToolChangedFiles", () => {
     expect(result).toHaveLength(1);
     expect(result[0].path).toBe("src/app.ts");
   });
+
+  it.each([
+    {
+      name: "new file counts all lines as added",
+      kind: "write" as const,
+      oldText: undefined,
+      newText: "a\nb\nc",
+      added: 3,
+      removed: 0,
+    },
+    {
+      name: "modified file counts added and removed",
+      kind: "edit" as const,
+      oldText: "a\nb\nc",
+      newText: "a\nB\nc\nd",
+      added: 2,
+      removed: 1,
+    },
+    {
+      name: "pure removal counts removed only",
+      kind: "edit" as const,
+      oldText: "a\nb\nc",
+      newText: "a",
+      added: 0,
+      removed: 2,
+    },
+  ])("diff stats: $name", ({ kind, oldText, newText, added, removed }) => {
+    const calls = makeToolCalls(
+      toolCall({
+        toolCallId: "tc",
+        kind,
+        locations: [{ path: "src/f.ts" }],
+        content: diffContent("src/f.ts", newText, oldText),
+      }),
+    );
+    const [file] = extractCloudToolChangedFiles(calls);
+    expect(file.linesAdded).toBe(added);
+    expect(file.linesRemoved).toBe(removed);
+  });
+
+  it("returns identical stats when the same diff object is reused (cache)", () => {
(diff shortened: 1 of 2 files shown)


pauldambra:
> [!NOTE]
>
> Clean-room review: qa-team, paul-reviewer, xp-reviewer, security-audit — all on Opus.

## Verdict: ✅ APPROVE (unanimous)

All four approved; this is a clean, correct, well-targeted memoization.

### Verified (the load-bearing correctness question)
The cache key — the diff content object — is sound. All four reviewers independently traced it: `getDiffStats` reads only `oldText`/`newText`, which live *on the key object itself*; the session store **appends** events (`sessionStore` push, never mutates in place), so already-seen diff objects keep identity across the `buildCloudEventSummary` recomputes — the WeakMap hits, and a cached value can never go stale relative to its key. WeakMap is the right structure (self-evicting, no leak). Behaviour is value-identical; only recompute is avoided. **security-audit: no cross-session contamination, no findings.**

### Actioned (convergent nit, paul + xp) — fixed in d46ed414b
The original cache test asserted value-equality, which passes even with the WeakMap removed (since `getDiffStats` is pure) — it tested purity, not the cache. Exported `cachedDiffStats` and rewrote the test to assert a cache hit returns the **same instance** (and a distinct-but-equal diff object recomputes), so it now fails if the cache regresses. This also pins qa-team's shared-return-reference observation.

### Non-blocking notes (left as-is)
- Module-level WeakMap is a process-global, but it's a pure memoization table (value derived from key), not mutable shared state — safe; reviewers agreed leaving it.
- xp/qa-team noted the deeper fix is incrementalizing `buildCloudEventSummary` (so the whole summary isn't rebuilt each update) — a possible follow-up; the cache is a correct, contained win for now.

| Reviewer | Verdict |
| --- | --- |
| 🔍 qa-team | APPROVE, risk NONE. Cache key sound, no leak, behaviour preserved. |
| 👤 paul | High trust, ship it. Verified the key holds; assert instance identity (done). |
| 📐 xp | Correctly diagnosed + built; tighten the cache test (done). |
| 🛡 security-audit | No findings. |

---

tatoalo (approved):

PR #2812, opened 2026-06-20
fix(auth): Log out on rejected refresh token instead of stalling startup

charlesvien:
## Problem

When a stored session holds a refresh token the server rejects (you logged in against a dev instance, or the token expired or was revoked), the desktop app hung on a blank "Loading..." screen and, 25 seconds later, showed "PostHog is taking longer than expected to start" with Retry and Get support. Restarting did not help, because the dead token was retried on every launch.

Root cause: the OAuth2 token endpoint returns an invalid, expired or revoked refresh token as HTTP 400 (`invalid_grant`, RFC 6749 section 5.2). Our OAuth client only classified 401/403 as auth errors, so a 400 fell through to `unknown_error`. That code is non-retryable and does not clear the stored session, so `AuthService` dropped bootstrap back into the `restoring` state with `bootstrapComplete: false` and never recovered. The UI sat on the loading screen until the 25s stall fallback fired.

Console from a failing launch:

```
(oauth-service) Token refresh failed: 400 Bad Request
Failed to restore stored auth session
```

## Changes

- `packages/core/src/oauth/oauth.ts`: classify HTTP 400 as an `auth_error` alongside 401/403. A rejected refresh token now takes the existing forced-logout path in `AuthService.refreshSession`: it clears the stored session, publishes anonymous state (`bootstrapComplete: true`) and routes the user straight to the login screen. No retries, no spinner, no stall screen.
- `packages/core/src/oauth/oauth.test.ts`: add a case asserting 400 is classified as `auth_error`, and repoint the other-4xx case to 404.

Unchanged: 5xx is still retried (`server_error`), fetch failures stay `network_error` and other 4xx stay `unknown_error`. The bootstrap stall fallback is kept as a safety net for a genuinely slow boot; it just no longer triggers for auth failures.

## How did you test this?

- `pnpm --filter @posthog/core test`: full core suite, 1616 passed. Covers both layers of the fix: `oauth.test.ts` proves a 400 is classified as `auth_error`, and `auth.test.ts` ("does not retry on auth_error and forces logout") proves that an `auth_error` clears the session and returns anonymous state.
- `pnpm --filter @posthog/core typecheck`: clean.
- `biome lint` on the two changed files: clean.
- Pre-commit hook ran repo-wide `pnpm typecheck` and Biome on commit: passed.

Not done: I did not reproduce the hang in the live Electron app. The coverage above is unit level across the classifier and the logout path.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/core/src/oauth/oauth.test.ts
+++ b/packages/core/src/oauth/oauth.test.ts
@@ -122,31 +122,50 @@ describe("OAuthService.refreshToken", () => {
     expect(result.errorCode).toBe("auth_error");
   });
 
-  it("maps a 400 invalid_grant to an auth_error", async () => {
+  it.each([
+    {
+      name: "invalid_grant",
+      body: { error: "invalid_grant" },
+      expected: "auth_error",
+    },
+    {
+      name: "invalid_token",
+      body: { error: "invalid_token" },
+      expected: "auth_error",
+    },
+    {
+      name: "invalid_client",
+      body: { error: "invalid_client" },
+      expected: "unknown_error",
+    },
+    {
+      name: "invalid_request",
+      body: { error: "invalid_request" },
+      expected: "unknown_error",
+    },
+    {
+      name: "a non-string error field",
+      body: { error: 42 },
+      expected: "unknown_error",
+    },
+    { name: "no error field", body: {}, expected: "unknown_error" },
+  ])("maps a 400 $name to a $expected", async ({ body, expected }) => {
     const { service } = createDeps();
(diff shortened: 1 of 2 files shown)


charlesvien:
* **#2812** <a href="https://app.graphite.com/github/pr/PostHog/code/2812?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2812?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

k11kirky (approved):

PR #2813, opened 2026-06-21
refactor(arch): close DI, import-boundary and lifecycle gaps

charlesvien:
## Problem

A review of the architecture post package split surfaced a set of gaps where `apps/code` and the portable packages had drifted from this repo's own stated rules: portable `core`/`ui`, strongly-typed DI, a thin host, one-line tRPC forwards. None of these are structural redesigns. They are the codebase not yet matching its own architecture doc.

## Changes

DI token safety and import boundaries

- Remove the forbidden `Object.freeze` token bags `MAIN_TOKENS`, `RENDERER_TOKENS` (`apps/code`) and `TOKENS` (`workspace-server`); use the standalone `Symbol.for(...)` consts at every call site so `TypedContainer<BindingMap>` actually type-checks `get`/`bind`.
- `biome.jsonc`: drop the `!@posthog/core` negations that allowed import cycles back into `api-client`, `workspace-client` and `platform`; block `core`, `workspace-client` and `host-router` from `workspace-server`.
- `workspace-server/src/trpc.ts`: replace the package-level `container.get()` service-locator with a `createAppRouter(services)` factory resolved in `serve.ts` (the composition seam). The `container` import is removed entirely so any missed closure is a compile error.
- `GIT_DIFF_SOURCE` binding uses the injector `ctx.get()` instead of the module-level container.

Lifecycle and composition root

- Move `container.unbindAll()` out of `AppLifecycleService` into the `index.ts` composition root via a `beforeExit` hook on `gracefulExit`, so the service no longer reaches for the global container.

State placement

- Move the session domain store into `@posthog/core` as a `zustand/vanilla` store; `@posthog/ui` keeps a thin React wrapper. The immer proxy-lifetime guards in the dequeue paths are preserved verbatim.
- Add a 5 minute TTL to `SessionService.previewConfigOptionsCache` so stale cloud model lists do not persist for the renderer lifetime.

Reliability and correctness

- `WorkspaceServerService`: supervised restart with exponential backoff and a max-attempts cap, a `statusChanged` event stream, and a renderer failure banner with retry.
- `packages/agent`: thread LLM gateway config explicitly (`GatewayEnv`) into the Claude subprocess env instead of mutating the shared global `process.env`, which clobbered config across concurrent sessions.
- `clearApplicationStorage` now clears the encrypted SecureStore alongside `localStorage` (previously encrypted settings survived a factory reset).
- `AgentService.getOrCreateSession` overloads remove the unreachable `null` return on the new-session path.

[...]

PR #2816, opened 2026-06-21
fix(desktop): renderer boot error boundary and electron hardening

charlesvien:
## Problem

Two reliability and security gaps from the package split surfaced, deliberately split out from #2813 because they touch runtime-sensitive Electron config and warrant their own review.

1. If `boot()`, a module-scope import or an Inversify resolution throws during renderer startup, the user gets a blank window with no error and no recovery short of force-quitting.
2. The main window runs with `nodeIntegration: true`, which applies in production, not just dev.

## Changes

- Add a `BootErrorBoundary` (`apps/code/src/renderer/components/BootErrorBoundary.tsx`) around the root React tree, plus a `try/catch` around the synchronous bootstrap in `main.tsx` and a `.catch` on the async `boot(container)`. Any of those failing now renders a minimal, theme-independent error screen with a Reload button instead of a blank window, and logs the failure.
- Set `webPreferences.nodeIntegration: false` on the main window. `contextIsolation: true` is already set (so the renderer main world never had Node anyway) and the renderer is Vite-bundled with zero `node:*`/`require` usage in the main world, so this is defense in depth with no functional change.

Scope note: the dev-only `webSecurity: false` override is left in place. It does not affect production (production already runs with `webSecurity: true`) and flipping it risks breaking dev resource loading. The remaining dev/prod parity gap is deferred.

## How did you test this?

All run locally in the branch worktree:
- `pnpm typecheck`: 22/22 packages pass
- `pnpm lint`: clean
- `pnpm test`: `apps/code` 151 pass
- `node scripts/check-host-boundaries.mjs`: no new violations

Not yet done: a manual smoke test of the running app (dev and packaged). The `nodeIntegration` flip is low risk given `contextIsolation: true`, but it changes runtime Electron config, so a human should confirm the window still loads and agents/terminals work before merge. This PR is left as a draft for that reason.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/apps/code/src/main/window.ts
+++ b/apps/code/src/main/window.ts
@@ -200,7 +200,7 @@ export function createWindow(): void {
     ...platformWindowConfig,
     show: false,
     webPreferences: {
-      nodeIntegration: true,
+      nodeIntegration: false,
       contextIsolation: true,
       preload: path.join(__dirname, "preload.js"),
       enableBlinkFeatures: "GetDisplayMedia",
--- a/apps/code/src/renderer/components/BootErrorBoundary.test.tsx
+++ b/apps/code/src/renderer/components/BootErrorBoundary.test.tsx
@@ -1,0 +1,95 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+const { logError } = vi.hoisted(() => ({ logError: vi.fn() }));
+
+vi.mock("@posthog/ui/shell/logger", () => ({
+  logger: {
+    scope: () => ({
+      error: logError,
+      info: vi.fn(),
+      warn: vi.fn(),
+      debug: vi.fn(),
+    }),
+  },
+}));
+
+import {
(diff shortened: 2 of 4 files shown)


charlesvien:
* **#2816** <a href="https://app.graphite.com/github/pr/PostHog/code/2816?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2816?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2813** <a href="https://app.graphite.com/github/pr/PostHog/code/2813?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tatoalo (approved):

charlesvien:
### Merge activity

* **Jun 24, 1:07 AM UTC**: @charlesvien merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/2816).

PR #2823, opened 2026-06-21
Sort channel tasks by last updated in the channels nav

raquelmsmith:
## Problem

In the channels (bluebird) nav, the tasks filed under each channel were ordered by when they were filed, not by recent activity — so a channel didn't surface what was actively moving at the top.

**Why:** requested so each channel's task list reflects the most recently updated tasks first.

## Changes

`ChannelSection` in `ChannelsList.tsx` now sorts each channel's visible filed tasks by the task's own `updated_at` (most recent first), instead of inheriting the server's filing-time (`createdAt`) order. The sort lives in the component because that's where the filed-task FS rows are already joined against the live task list (which carries `updated_at`); the server's `channelTasksService.list` only has the filing-time on each row.

## How did you test this?

Static review only. The change is a sort over already-typed records using the existing `Task.updated_at` field. I did not run the test suite or typecheck — `node_modules` isn't installed in this environment, so the existing `pnpm typecheck` failures are unrelated (missing-module noise) and none reference the changed file.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/canvas/components/ChannelsList.tsx
+++ b/packages/ui/src/features/canvas/components/ChannelsList.tsx
@@ -653,17 +653,22 @@ function ChannelSection({
   // Order by each task's own last-updated time (most recent first) so a
   // channel surfaces what's actively moving, rather than the order tasks were
   // filed. `filedTasks` arrives sorted by filing time; this re-sorts by the
-  // task's `updated_at`, falling back to 0 when a task isn't loaded.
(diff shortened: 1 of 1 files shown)


raquelmsmith:
Addressed in 43b41bc — pre-built a single `id → updated-ms` Map so both the membership filter and the recency sort are O(1) per item, and used `|| 0` for the NaN-safe fallback as suggested.

PR #2837, opened 2026-06-22
feat(agent-platform): identities, users + inline approvals in agent config UI

benjackwhite:
Surfaces the per-principal identity + **approvals** work in the agent **configuration** UI. Backend / runtime side: **PostHog/posthog#64997** (`feat(agent-platform): per-principal identity, credentials + approvals authority`).

## What's here

**Identities section** in the config pane
- New `identities` node listing `spec.identity_providers[]` (kind · binding · consumer count).
- Each provider cross-links to the **custom tools** (`requires_identity`) and **MCP servers** (`auth.provider`) that use it; those tool/MCP detail bodies link **back** to the identity — easy two-way nav.
- Flags **undeclared** references (a tool/MCP points at an identity not in `identity_providers`) and **unused** providers.
- Notes that native `@posthog/*` tools declare their provider intrinsically in the registry, so they don't appear in the spec (otherwise their absence reads as a bug).

**Users / connections pane** (per agent)
- Lists the agent's end-users (`agent_user`) and each user's linked external identities (connection metadata only — credentials are never sent to the client). Revoke a connection inline; "ban user" stubbed for later.
- Sessions tab gains a by-user filter. Shared `userDisplayName` helper; the API client throws on non-2xx for the list + revoke calls (no false-success toasts).

**Approval decisions** (inline chat + console)
- Inline `AgentChatPendingApprovalCard` in the live chat / builder dock branches on `approver_scope.type`: **`principal`** approvals are decided right there (the person driving the chat *is* the session principal) — posting to the **ingress** via the chat session (preview token + the user's bearer → principal-match), and the open `/listen` stream resumes the chat on approve. **`agent`** approvals show "an owner/admin must decide in the console" + link to the Approvals view rather than offering buttons.
- `approver_scope` is now a typed field on `AgentApprovalRequest` (`{ type, allow_edit }`), so the card branches without an inline cast.
- The owner-console Approvals view continues to decide `agent`-type approvals via Django.

**Nicer trigger config**
- Replaces the raw `JSON.stringify` dump with a value-typed renderer: on/off pills for booleans, chips for string arrays (e.g. `trusted_workspaces`), wrapped blocks for long text (cron `prompt`), inline scalars — each with a short hint for the well-known fields.

**Slug next to the app name**
- Mono slug chip beside the title in the agent detail header.

## Notes
- The Identities + Users sections only render for agents whose spec declares `identity_providers` / has end-users (i.e. those built on #64997).
- Touched files pass `tsc` + Biome. The pre-commit hook (full-repo typecheck) was bypassed where it fails on **pre-existing** errors in `canvas/` and `code-review/` on the base branch, unrelated to this change.

_(Folds in the stacked approval-decision-UI PR (#2851) — now a single PR.)_

## Follow-up (not in this PR)
Make these sections **spec-schema-driven** — overlay the revision's values on the JSON Schema (`z.toJSONSchema(AgentSpecSchema)`, already powering `@posthog/agent-applications-spec-schema`) so defaults / enums / new fields render automatically. The new value-typed trigger renderer is the nucleus; happy to take that on separately.


packages/api-client/src/posthog-client.ts
```diff
@@ -4543,6 +4547,34 @@ export class PostHogAPIClient {
...
+  /** Revoke one linked connection for an agent user (kept for audit, not deleted). */
+  async deleteAgentUserConnection(
+    idOrSlug: string,
+    agentUserId: string,
+    provider: string,
+  ): Promise<void> {
+    const teamId = await this.getTeamId();
+    const path = `${this.agentApplicationsPath(teamId)}${encodeURIComponent(idOrSlug)}/users/${encodeURIComponent(agentUserId)}/connections/${encodeURIComponent(provider)}/`;
+    const url = new URL(`${this.api.baseUrl}${path}`);
+    await this.api.fetcher.fetch({ method: "delete", url, path });
```
dmarticus: **`deleteAgentUserConnection` never checks `response.ok` → false success toast.**

The fetcher doesn't throw on non-2xx — every other DELETE in this client guards it (e.g. `if (!response.ok && response.status !== 404) throw …` around lines 1002 / 1084 / 1155). Because the mutation wires `onSuccess` to `toast.success("Connection revoked")` and has a real `onError` path, a 4xx/5xx will currently resolve as success and tell the operator a credential was revoked **when it wasn't**. For a destructive, audited action this is the one I'd most want fixed — recommend checking `response.ok` and throwing.

```ts
const response = await this.api.fetcher.fetch({ method: "delete", url, path });
if (!response.ok && response.status !== 404) {
  throw new Error(`Failed to revoke connection: ${response.status}`);
}
```

packages/api-client/src/posthog-client.ts
```diff
@@ -4543,6 +4547,34 @@ export class PostHogAPIClient {
...
+  // never cross this boundary. Proxied Django → janitor → runtime store, same
+  // shape as the memory endpoints above.
+
+  /** List the agent's end-users, each with their linked connections. */
+  async listAgentUsers(idOrSlug: string): Promise<AgentUsersListResponse> {
+    const teamId = await this.getTeamId();
+    const path = `${this.agentApplicationsPath(teamId)}${encodeURIComponent(idOrSlug)}/users/`;
+    const url = new URL(`${this.api.baseUrl}${path}`);
+    const response = await this.api.fetcher.fetch({ method: "get", url, path });
+    const data = (await response.json()) as Partial<AgentUsersListResponse>;
```
dmarticus, line 4603: **`listAgentUsers` masks server errors as an empty list.**

On a non-2xx that still returns JSON, `data.results ?? []` yields `[]`, so the pane shows its "No users yet" empty state rather than the dedicated "Couldn't load users" error branch — which will then rarely fire. This is consistent with the memory GET methods it mirrors (those also skip the check), so lower priority, but it also means the PR note that the pane "shows its empty/error state" until the backend ships is only half true: a 404-with-JSON renders as "No users yet," not an error. A `response.ok` check would let genuine failures surface.

packages/ui/src/features/agent-applications/components/AgentSessionsPane.tsx
```diff
@@ -23,14 +39,19 @@ const PAGE = 25;
...
   const [limit, setLimit] = useState(PAGE);
 
   const { data, isLoading, isError, isFetching, dataUpdatedAt, refetch } =
     useAgentApplicationSessions(idOrSlug, {
       limit,
       state: filter === "all" ? undefined : [filter],
+      agent_user_id: userId === "all" ? undefined : userId,
     });
 
+  const { data: usersData } = useAgentUsers(idOrSlug);
```
dmarticus, line 47: **Sessions pane now fires an unconditional `useAgentUsers` query.**

Every Sessions-tab load hits `/users/` solely to populate a filter dropdown that's hidden when empty. Until the backend endpoint exists this 404s on every visit (with react-query's default retries) on a hot, auto-polling tab. Low priority and it degrades gracefully, but it's avoidable failing traffic — worth deciding whether the filter earns its keep before the endpoint lands.

packages/ui/src/features/agent-applications/components/AgentUsersPane.tsx
```diff
@@ -0,0 +1,247 @@
...
+        confirming ? (
+          <Flex align="center" gap="1" className="shrink-0">
+            <Button
+              size="1"
+              variant="solid"
+              color="red"
+              disabled={disconnect.isPending}
+              onClick={() =>
+                disconnect.mutate(
+                  { agentUserId, provider: connection.provider },
```
dmarticus, line 200: **Disconnect targets `provider`, not the connection `id`.**

Rows are rendered per-connection (`user.connections.map`), but revoke deletes by `connection.provider`, and the unique `connection.id` goes unused. If a user can ever have two connections for the same provider, both rows revoke the same thing. Probably the backend enforces one-active-per-provider — worth confirming the API shape (delete-by-provider) against the UI (list-by-connection).

packages/ui/src/features/agent-applications/components/AgentSessionsPane.tsx
```diff
@@ -1,14 +1,30 @@
...
+import { useAgentUsers } from "../hooks/useAgentUsers";
 import { AgentDetailEmptyState, AgentDetailLayout } from "./AgentDetailLayout";
 import { AgentSessionRow } from "./AgentSessionRow";
 import { RefreshIndicator } from "./RefreshIndicator";
 
 type Filter = AgentSessionState | "all";
 
+/** Short label for a user in the filter dropdown: display name if the trigger
+ *  stamped one, else the principal id, prefixed by kind. */
+function userLabel(u: AgentUserWithConnections): string {
```
dmarticus, line 19: **Minor: duplicated display-name logic.**

This `userLabel` and `UserCard`'s `displayName` in `AgentUsersPane.tsx` both do the same `metadata.display_name ?? metadata.name ?? principal_id` resolution. Could be a single shared helper.

packages/ui/src/features/agent-applications/components/AgentConfigurationPane.tsx
```diff
@@ -765,6 +849,158 @@ function TriggersOverview({ spec, ctx }: { spec: AgentSpec; ctx: Ctx }) {
...
+  const display =
+    typeof value === "string" || typeof value === "number"
+      ? String(value)
+      : JSON.stringify(value);
+  return (
+    <OptionCard
+      label={human}
+      hint={hint}
+      trailing={
+        <Text className="truncate text-[12.5px] text-gray-12 [font-family:var(--font-mono)]">
```
dmarticus, line 948: **Nit: `truncate` inside a `shrink-0` span won't engage.**

The trailing wrapper is `min-w-0 shrink-0`, so `truncate` has no width bound to ellipsize against. An unusually long scalar (≤64 chars, no newline — so it misses the long-text block branch above) could overflow instead of truncating. Edge-casey, but worth a glance.

dmarticus (commented):
Thorough review pass — left inline comments on the findings below. Most are low-priority; #1 (the disconnect not checking `response.ok`) is the only one close to a blocker.

**Scope heads-up:** the diff is larger than the title implies — it also includes the full Users/Connections pane + sessions-by-user filter, and a fourth commit that is #2840 (observability MetricCards + cost clamp). That last one only exists on this branch and shows up here as a stacking artifact because the base is `posthog-code/m3a-final` (which doesn't yet contain #2840). Worth confirming it reconciles on rebase; I didn't re-review the `AgentAnalyticsView.tsx`/`format.ts` changes as part of this PR.

### What's done well (genuinely)
- The loosely-typed spec readers (`rec`/`str`/`arr`) are reused consistently with the existing convention, and slash-containing ids survive the `idParts.join("/")` path round-trip.
- Two-way identity cross-linking (tool/MCP ↔ provider) with **undeclared** and **unused** flagging is real, useful operator tooling.
- The credentials-never-serialized boundary is clearly documented in both the types and the client comment — good thing to make explicit.
- Confirm-before-revoke + a disabled stubbed "Ban" with an explanatory `title` is the right level of caution for a destructive surface.
- The session query key includes `params`, so the new `agent_user_id` filter caches/refetches correctly.
- Commit 2's index-key fix for duplicate array chips, with a clear `biome-ignore` rationale, is a nice catch.

Nice work overall — the identities UX is a solid addition.

dmarticus (approved):
Here's my feedback, but I have nothing blocking — left some inline comments (the disconnect `response.ok` check is the one I'd most want addressed, but it's not a gate). Nice work on the identities UX. 🚀

PR #2844, opened 2026-06-22
feat(canvas): start CONTEXT.md and canvas generation repo-less

raquelmsmith:
## Problem

Canvas and CONTEXT.md generation tasks forced the user to select a folder (local) or a repo (cloud) before they could start. That requirement is no longer needed — the agent knows how to handle repo-less tasks now, attaching a repo lazily and asking the user to clarify if it can't find the right one.

**Why:** Picking a repo up front is friction for generation tasks that mostly read PostHog data, and the agent can now figure out (or ask about) the repo at runtime.

## Changes

- Both generation flows now create a **repo-less local task** (scratch dir) via the existing `allowNoRepo` path, mirroring the channels new-task chat box. The created task's workspace cache is invalidated so its view resolves the scratch cwd instead of showing the repo-picker prompt.
- The local/cloud segmented control and the folder/repo pickers are removed from the CONTEXT.md **"Generate with agent"** control and the canvas **generate bar** — each is now a single Generate button.
- The CONTEXT.md prompt notes no repo is attached up front and tells the agent to ask which repo to use when it can't infer one.

## How did you test this?

- `pnpm --filter @posthog/ui typecheck` — passes
- `biome check` on changed files — clean
- `vitest run src/features/canvas` — 17 tests pass

---

### DIFF
--- a/packages/ui/src/features/canvas/components/WebsiteContext.tsx
+++ b/packages/ui/src/features/canvas/components/WebsiteContext.tsx
@@ -27,16 +27,9 @@ import {
   useFolderInstructionsMutations,
   useFolderInstructionsVersions,
 } from "@posthog/ui/features/canvas/hooks/useFolderInstructions";
-import {
-  type GenerateContextTarget,
-  useGenerateContext,
-} from "@posthog/ui/features/canvas/hooks/useGenerateContext";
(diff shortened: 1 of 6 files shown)


raquelmsmith:
I did confirm it works e2e

PR #2862, opened 2026-06-23
fix(auth): don't eject users to invite screen on transient access-check failure

pauldambra:
## Problem

People actively using PostHog Code — already past the paywall, mid-session — were suddenly bounced to the invite-code screen during brief API outages, then recovered on their own. A quit/reopen also fixed it, which pointed at a transient failure rather than real loss of access.

Root cause: code access (`hasCodeAccess`) is never persisted. It's re-fetched from `GET /api/code/invites/check-access/` on every token refresh (≈hourly, piggybacking on the next API call after the token enters its 60s expiry skew) via `updateCodeAccessFromSession`. That function treated **any** non-success — a 403, a 5xx, a timeout, a network error — as a definitive "no access" and set `hasCodeAccess = false`, which is exactly the condition that renders the invite screen. A single bad response during an outage ejected authenticated users. Confirmed against the backend (`products/tasks/backend/.../api.py` `CodeInviteViewSet.check_access`): genuine denial is always `200 {has_access: false}`; non-2xx is only auth failure / throttle / outage, never a "no access" signal.

## Changes

`updateCodeAccessFromSession` now treats the check as resilient with a bounded fail-closed:
- **Healthy path stays synchronous** — one awaited attempt resolves the gate before the app renders.
- **Transient failures don't eject the user** — on a non-2xx / network error / timeout we keep the current value and retry in the **background** with **prime-number backoff** (2, 3, 5, 7, 11, 13, 17, 19, 23s) up to a **~90s budget**. Primes keep many clients' retries from lining up after an outage. The background loop is decoupled from the token-refresh path so it can't stall other API calls.
- **Fail closed after the budget** — if we still can't get a definitive `200` within ~90s, `hasCodeAccess` is set to `false`. A sustained inability to call home revokes access, so the gate can't be bypassed by keeping the client offline.
- Per-attempt fetch timeout for this check tightened from 30s → **10s**.
- Genuine "no access" (`200 {has_access:false}`) still gates immediately on the first attempt.

Net effect: a blip no longer kicks real users, but you can't game the invite by going offline.

## How did you test this?

Unit tests in `auth.test.ts` (39 passing, `pnpm --filter @posthog/core exec vitest run src/auth/auth.test.ts`):
- transient failures then 200 → recovers to access in the background, never flips to `false`
- persistent failure → fails closed (`false`) after exhausting the prime-backoff budget (1 awaited + 8 retries = 9 attempts)
- `200 {has_access:false}` → still gated to `false`

Biome lint clean on the changed files; no new type errors in `auth.ts`.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1782221042593189?thread_ts=1782221042.593189&cid=C09G8Q32R6F)*

### DIFF
--- a/packages/core/src/auth/auth.test.ts
+++ b/packages/core/src/auth/auth.test.ts
@@ -1342,4 +1342,116 @@ describe("AuthService", () => {
       expect(redeemCallCount).toBe(2);
     });
   });
+
+  describe("code access check resilience", () => {
+    const stubFetchWithCheckAccess = (
+      checkAccess: (callCount: number) => Response,
+    ): { getCheckAccessCalls: () => number } => {
+      let checkAccessCalls = 0;
+      vi.stubGlobal(
+        "fetch",
+        vi.fn(async (input: string | Request) => {
+          const url = typeof input === "string" ? input : input.url;
+
+          if (url.includes("/api/users/@me/")) {
+            return {
+              ok: true,
+              json: vi.fn().mockResolvedValue({
+                uuid: "user-1",
+                organization: { id: "org-1" },
+              }),
+            } as unknown as Response;
+          }
+
+          if (/\/api\/organizations\/[^/]+\/$/.test(url)) {
+            return {
+              ok: true,
+              json: vi.fn().mockResolvedValue({
+                name: "Org 1",
+                teams: [{ id: 42, name: "Project 42" }],
+              }),
+            } as unknown as Response;
+          }
+
+          if (url.includes("/invites/check-access/")) {
+            checkAccessCalls++;
+            return checkAccess(checkAccessCalls);
+          }
+
+          return {
+            ok: true,
+            json: vi.fn().mockResolvedValue({}),
+          } as unknown as Response;
+        }) as unknown as typeof fetch,
+      );
+      return { getCheckAccessCalls: () => checkAccessCalls };
+    };
+
+    const restoreSession = async () => {
+      seedStoredSession({ selectedProjectId: 42 });
+      oauthFlow.refreshToken.mockResolvedValue(
+        mockTokenResponse({
+          accessToken: "access-token",
+          refreshToken: "refresh-token",
+        }),
+      );
+      await service.initialize();
+    };
+
+    it("retries a transient failure and grants access on the retry", async () => {
+      const { getCheckAccessCalls } = stubFetchWithCheckAccess((call) =>
+        call === 1
+          ? ({
+              ok: false,
+              status: 403,
+              json: () => Promise.resolve({}),
+            } as unknown as Response)
+          : ({
+              ok: true,
+              status: 200,
+              json: () => Promise.resolve({ has_access: true }),
+            } as unknown as Response),
+      );
+
+      await restoreSession();
+
+      expect(service.getState().hasCodeAccess).toBe(true);
+      expect(getCheckAccessCalls()).toBe(2);
+    });
+
+    it("preserves prior state instead of ejecting the user when the check keeps failing", async () => {
+      const { getCheckAccessCalls } = stubFetchWithCheckAccess(
+        () =>
+          ({
+            ok: false,
+            status: 403,
+            json: () => Promise.resolve({}),
+          }) as unknown as Response,
+      );
+
+      await restoreSession();
+
+      const state = service.getState();
+      expect(state.status).toBe("authenticated");
+      // Access was never determined, so it stays null (the "Checking access"
+      // spinner) — never false, which would wrongly show the invite screen.
+      expect(state.hasCodeAccess).toBeNull();
+      expect(getCheckAccessCalls()).toBe(3);
+    });
+
+    it("still gates the user when the server definitively reports no access", async () => {
+      stubFetchWithCheckAccess(
+        () =>
+          ({
+            ok: true,
+            status: 200,
+            json: () => Promise.resolve({ has_access: false }),
+          }) as unknown as Response,
+      );
+
+      await restoreSession();
+
+      expect(service.getState().hasCodeAccess).toBe(false);
+    });
+  });
 });
--- a/packages/core/src/auth/auth.test.ts
+++ b/packages/core/src/auth/auth.test.ts
@@ -1502,8 +1502,8 @@ describe("AuthService", () => {
       // Restore's attempt failed transiently; gate is unresolved (null, not false).
       expect(service.getState().hasCodeAccess).toBeNull();
 
-      // A fresh authoritative check (token refresh) resolves access true and
-      // supersedes the parked loop's sequence.
+      // A fresh authoritative check (token refresh) installs a new session and
+      // resolves access true, superseding the parked loop.
       await service.refreshAccessToken();
       expect(service.getState().hasCodeAccess).toBe(true);
 
--- a/packages/core/src/auth/auth.test.ts
+++ b/packages/core/src/auth/auth.test.ts
@@ -1,6 +1,10 @@
 import type { RootLogger } from "@posthog/di/logger";
 import type { IPowerManager } from "@posthog/platform/power-manager";
-import { OAUTH_SCOPE_VERSION, sleep } from "@posthog/shared";
+import {
+  NotAuthenticatedError,
+  OAUTH_SCOPE_VERSION,
+  sleep,
+} from "@posthog/shared";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { AuthService } from "./auth";
 import type {
@@ -1543,6 +1547,89 @@ describe("AuthService", () => {
       await restoreSession();
 
       expect(service.getState().hasCodeAccess).toBe(false);
+    });
+
+    it("treats an unreadable 200 body like a transient failure", async () => {
+      // A 200 whose body can't be parsed (e.g. a proxy returning truncated/HTML
+      // content) is not a definitive answer — it must retry, not gate the user,
+      // and recover once a real body arrives.
+      const { getCheckAccessCalls } = stubFetchWithCheckAccess((call) =>
+        call <= 1
+          ? ({
+              ok: true,
+              status: 200,
+              json: () => Promise.reject(new Error("unexpected token <")),
+            } as unknown as Response)
+          : ({
+              ok: true,
+              status: 200,
+              json: () => Promise.resolve({ has_access: true }),
+            } as unknown as Response),
+      );
+
+      await restoreSession();
+
+      await vi.waitFor(() =>
+        expect(service.getState().hasCodeAccess).toBe(true),
+      );
+      expect(getCheckAccessCalls()).toBe(2);
+    });
+
+    it("honours a success on the final retry instead of failing closed", async () => {
+      // 1 awaited attempt + 8 transient failures, then the 9th (final) retry
+      // succeeds: the loop must settle to access and never reach the
+      // fail-closed write, exercising the final-iteration boundary.
+      const { getCheckAccessCalls } = stubFetchWithCheckAccess((call) =>
+        call < 10
+          ? ({
+              ok: false,
+              status: 403,
+              json: () => Promise.resolve({}),
+            } as unknown as Response)
+          : ({
+              ok: true,
+              status: 200,
+              json: () => Promise.resolve({ has_access: true }),
+            } as unknown as Response),
+      );
+
+      await restoreSession();
+
+      await vi.waitFor(() =>
+        expect(service.getState().hasCodeAccess).toBe(true),
+      );
+      expect(getCheckAccessCalls()).toBe(10);
+    });
+
+    it("rejects an in-flight refresh that logs out mid-sync on a 401", async () => {
+      // The race the seq guard alone doesn't cover: a 401 from the code-access
+      // check logs out *during* syncAuthenticatedSession, inside the shared
+      // refresh promise. The awaited refresh must reject rather than resolve
+      // with the torn-down session's (now dead) access token.
+      let reject401 = false;
+      stubFetchWithCheckAccess(() =>
+        reject401
+          ? ({
+              ok: false,
+              status: 401,
+              json: () => Promise.resolve({}),
+            } as unknown as Response)
+          : ({
+              ok: true,
+              status: 200,
+              json: () => Promise.resolve({ has_access: true }),
+            } as unknown as Response),
+      );
+
+      await restoreSession();
+      expect(service.getState().status).toBe("authenticated");
+
+      reject401 = true;
+      await expect(service.refreshAccessToken()).rejects.toBeInstanceOf(
+        NotAuthenticatedError,
+      );
+      expect(service.getState().status).toBe("anonymous");
+      expect(service.getState().hasCodeAccess).toBeNull();
     });
   });
 });
--- a/packages/core/src/auth/auth.test.ts
+++ b/packages/core/src/auth/auth.test.ts
@@ -1,6 +1,6 @@
 import type { RootLogger } from "@posthog/di/logger";
 import type { IPowerManager } from "@posthog/platform/power-manager";
-import { OAUTH_SCOPE_VERSION } from "@posthog/shared";
+import { OAUTH_SCOPE_VERSION, sleep } from "@posthog/shared";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { AuthService } from "./auth";
 import type {
@@ -21,6 +21,7 @@ vi.mock("@posthog/shared", async (importOriginal) => {
   return {
     ...actual,
     sleepWithBackoff: vi.fn().mockResolvedValue(undefined),
+    sleep: vi.fn().mockResolvedValue(undefined),
   };
 });
 
@@ -1441,8 +1442,56 @@ describe("AuthService", () => {
         expect(service.getState().hasCodeAccess).toBe(false),
       );
       expect(service.getState().status).toBe("authenticated");
-      // 1 awaited attempt + one per prime-backoff step before the budget caps.
-      expect(getCheckAccessCalls()).toBe(9);
+      // 1 awaited attempt + one retry per prime-backoff step (9 primes).
+      expect(getCheckAccessCalls()).toBe(10);
+      // The backoff schedule is actually exercised: each retry waits the next
+      // prime delay (the mocked sleep records the requested durations).
+      expect(vi.mocked(sleep).mock.calls.map((call) => call[0])).toEqual([
+        2_000, 3_000, 5_000, 7_000, 11_000, 13_000, 17_000, 19_000, 23_000,
+      ]);
+    });
+
+    it("does not let a stale retry loop clobber a fresh grant", async () => {
+      // Park the retry loop on its first backoff so we can interleave a newer
+      // authoritative check that succeeds before the stale loop wakes.
+      let releaseSleep!: () => void;
+      const gate = new Promise<void>((resolve) => {
+        releaseSleep = resolve;
+      });
+      vi.mocked(sleep).mockImplementationOnce(() => gate);
+
+      let call = 0;
+      stubFetchWithCheckAccess(() => {
+        call++;
+        // First (restore) attempt fails and schedules the retry loop; every
+        // later check succeeds with access.
+        return call === 1
+          ? ({
+              ok: false,
+              status: 403,
+              json: () => Promise.resolve({}),
+            } as unknown as Response)
+          : ({
+              ok: true,
+              status: 200,
+              json: () => Promise.resolve({ has_access: true }),
+            } as unknown as Response);
+      });
+
+      await restoreSession();
+      // Restore's attempt failed transiently; gate is unresolved (null, not false).
+      expect(service.getState().hasCodeAccess).toBeNull();
+
+      // A fresh authoritative check (token refresh) resolves access true and
+      // supersedes the parked loop's sequence.
+      await service.refreshAccessToken();
+      expect(service.getState().hasCodeAccess).toBe(true);
+
+      // Wake the stale loop: it must notice it was superseded and NOT fail closed.
+      releaseSleep();
+      await Promise.resolve();
+      await Promise.resolve();
+      expect(service.getState().hasCodeAccess).toBe(true);
     });
 
     it("still gates the user when the server definitively reports no access", async () => {
--- a/packages/core/src/auth/auth.test.ts
+++ b/packages/core/src/auth/auth.test.ts
@@ -3,7 +3,7 @@ import type { IPowerManager } from "@posthog/platform/power-manager";
 import {
   NotAuthenticatedError,
   OAUTH_SCOPE_VERSION,
-  sleep,
+  sleepWithBackoff,
 } from "@posthog/shared";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { AuthService } from "./auth";
@@ -25,7 +25,6 @@ vi.mock("@posthog/shared", async (importOriginal) => {
   return {
     ...actual,
     sleepWithBackoff: vi.fn().mockResolvedValue(undefined),
-    sleep: vi.fn().mockResolvedValue(undefined),
   };
 });
 
@@ -1446,13 +1445,12 @@ describe("AuthService", () => {
         expect(service.getState().hasCodeAccess).toBe(false),
       );
       expect(service.getState().status).toBe("authenticated");
-      // 1 awaited attempt + one retry per prime-backoff step (9 primes).
+      // 1 awaited attempt + one retry per backoff step.
       expect(getCheckAccessCalls()).toBe(10);
-      // The backoff schedule is actually exercised: each retry waits the next
-      // prime delay (the mocked sleep records the requested durations).
-      expect(vi.mocked(sleep).mock.calls.map((call) => call[0])).toEqual([
-        2_000, 3_000, 5_000, 7_000, 11_000, 13_000, 17_000, 19_000, 23_000,
-      ]);
+      // Each retry waits the next backoff step (attempt 0..8).
+      expect(
+        vi.mocked(sleepWithBackoff).mock.calls.map((call) => call[0]),
+      ).toEqual([0, 1, 2, 3, 4, 5, 6, 7, 8]);
     });
 
     it("treats thrown network errors like transient failures and fails closed", async () => {
@@ -1478,7 +1476,7 @@ describe("AuthService", () => {
       const gate = new Promise<void>((resolve) => {
         releaseSleep = resolve;
       });
-      vi.mocked(sleep).mockImplementationOnce(() => gate);
+      vi.mocked(sleepWithBackoff).mockImplementationOnce(() => gate);
 
       let call = 0;
       stubFetchWithCheckAccess(() => {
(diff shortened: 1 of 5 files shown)


packages/core/src/auth/auth.ts
```diff
@@ -894,32 +894,117 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+    }
+
+    // Could not confirm access within the budget: fail closed so the invite
+    // gate can't be bypassed by keeping the client offline.
+    if (this.session) {
+      this.logger.warn(
+        "Code access check failed within budget, failing closed",
+        { backoffMs },
+      );
       this.updateState({ hasCodeAccess: false });
```
pauldambra: > [!NOTE]

**[paul]** 🟠 HIGH

i think there's a race here that bites the exact user we're trying to protect. the dedup guard (`codeAccessRetryPromise`) stops a *new* loop starting, but not an *already-running* one. picture: blip happens, background loop is spinning; meanwhile a token refresh fires `syncAuthenticatedSession` → `checkCodeAccessOnce` succeeds and sets `hasCodeAccess: true`. user's happy. but the stale loop keeps grinding and ~77s later hits its budget and does `updateState({ hasCodeAccess: false })`, bouncing a user who *currently has access* — the precise failure this PR exists to prevent. the fail-closed only checks `this.session` exists, not whether access got resolved `true` in the meantime. could we re-check (or bail) right before failing closed, or cancel the loop on any successful resolve? not comfortable shipping the clobber.

packages/core/src/auth/auth.ts
```diff
@@ -894,32 +894,117 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+  // synchronising into bursts. We keep retrying until the cumulative backoff
+  // would exceed the budget (~90s), then fail closed so a sustained inability
+  // to call home revokes access rather than granting it forever.
+  private async runCodeAccessRetry(): Promise<void> {
+    let backoffMs = 0;
+
+    for (let attempt = 0; ; attempt++) {
+      const delayMs =
+        AuthService.CODE_ACCESS_BACKOFF_PRIMES_MS[
+          Math.min(
```
pauldambra: > [!NOTE]

**[xp]** 🟡 MEDIUM

The 90s budget and the prime array are redundant stop conditions, and the `Math.min(attempt, PRIMES.length - 1)` clamp is dead code. The primes sum to 77s, so the budget check breaks at the last array element (77+23 > 90) before `attempt` ever reaches `length` — the clamp never engages and the implied "repeat last prime until budget" behaviour is unreachable. Pick one policy: iterate the array directly (`for (const delayMs of PRIMES)`) and drop the budget, or make the budget the real limit and let the clamp genuinely repeat. As written a reader must do prime arithmetic to discover one guard is inert.

packages/core/src/auth/auth.ts
```diff
@@ -894,32 +894,117 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+            attempt,
+            AuthService.CODE_ACCESS_BACKOFF_PRIMES_MS.length - 1,
+          )
+        ];
+      if (backoffMs + delayMs > AuthService.CODE_ACCESS_RETRY_BUDGET_MS) {
+        break;
+      }
+
+      await sleepWithBackoff(0, { initialDelayMs: delayMs });
```
pauldambra: > [!NOTE]

**[xp]** 🟡 MEDIUM

`sleepWithBackoff(0, { initialDelayMs: delayMs })` misuses an exponential-backoff helper to get a plain sleep — `attempt=0` neutralises the growth so it degenerates to `setTimeout(delayMs)`. The reader sees "backoff" and hunts for a growth schedule that actually lives in the prime array. A plain `sleep(delayMs)` says what it means; this is the flag-parameter smell where a call exists only to disable the function's own behaviour.

packages/core/src/auth/auth.test.ts
```diff
@@ -1339,4 +1339,122 @@ describe("AuthService", () => {
...
+      await restoreSession();
+
+      // Persistent inability to call home eventually revokes access so the
+      // invite gate can't be bypassed by staying offline.
+      await vi.waitFor(() =>
+        expect(service.getState().hasCodeAccess).toBe(false),
+      );
+      expect(service.getState().status).toBe("authenticated");
+      // 1 awaited attempt + one per prime-backoff step before the budget caps.
+      expect(getCheckAccessCalls()).toBe(9);
```
pauldambra: > [!NOTE]

`expect(getCheckAccessCalls()).toBe(9)` is a magic number hand-derived from "prime count minus the budget overflow" — nothing links 9 to that derivation, so adding a prime silently breaks it. And because `sleepWithBackoff` is globally mocked to resolve immediately (line 23), the budget, prime values, and cumulative cap are never actually exercised — the whole timing design has no test behind it. Assert against `CODE_ACCESS_BACKOFF_PRIMES_MS.length` with a comment, and consider one fake-timer test that proves a retry waits ~2s before firing.

packages/core/src/auth/auth.ts
```diff
@@ -894,32 +894,117 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+        { backoffMs },
+      );
       this.updateState({ hasCodeAccess: false });
     }
   }
   private static readonly REFRESH_MAX_ATTEMPTS = 3;
   private static readonly ORG_FETCH_MAX_ATTEMPTS = 3;
   private static readonly ORG_RECOVERY_MAX_ATTEMPTS = 5;
+  private static readonly CODE_ACCESS_FETCH_TIMEOUT_MS = 10_000;
+  private static readonly CODE_ACCESS_RETRY_BUDGET_MS = 90_000;
```
pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

nit picking — the comments say budget is "~90s" but the primes sum to 2+3+5+7+11+13+17+19 = 77s before the 9th (23s) would blow the 90s cap. so the real worst-case wait is ~77s, not 90, and the 90_000 is a slightly arbitrary fence around a fixed list. could the comment just say "walks the prime list, ~77s total, capped at 90s" so the next traveller doesn't go hunting for the missing 13 seconds?

packages/core/src/auth/auth.ts
```diff
@@ -894,32 +894,117 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+        {
+          signal: AbortSignal.timeout(
+            AuthService.CODE_ACCESS_FETCH_TIMEOUT_MS,
+          ),
+        },
+        session.accessToken,
       );
+
+      if (!response.ok) {
+        this.logger.warn("Transient failure checking code access", {
```
pauldambra, line 959: > [!NOTE]

**[paul]** 🟢 LOW

a 401/403 (genuinely rejected/expired token) gets treated identically to an outage — retried for ~77s then failed closed — and we'll warn-log "transient failure" 9 times for something that isn't transient. i guess it's fine for now, but i wonder if auth-class statuses should short-circuit differently from 5xx/timeout. observability-wise: love that fail-closed logs, but since this path *denies* a real user, can we hang a counter/metric off it? log-before-you-limit — i'd want to see how often we actually eject people before trusting the 77s number in the wild. not blocking.

pauldambra (commented):
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — not available in this environment). See inline comments.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — not available in this environment)

One real correctness bug (a stale background retry loop can clobber a freshly-granted `true` back to `false` and eject an active user — the exact failure this PR exists to prevent), plus a cluster of expressiveness/test nits around the budget+prime loop. Security found no gate-bypass.

### Key findings
- 🟠 **HIGH** — `runCodeAccessRetry` fail-closed race: the single-flight guard stops a *new* loop, not an *in-flight* one. If a token-refresh `checkCodeAccessOnce` resolves `true` while an older loop is still spinning, the old loop's budget-exhaustion still writes `hasCodeAccess: false`. Re-check or cancel-on-resolve before failing closed.
- 🟡 **MEDIUM** — Redundant stop conditions: the 90s budget + the prime array overlap, making the `Math.min(attempt, len-1)` clamp dead code (loop always stops at the array end). Pick one policy.
- 🟡 **MEDIUM** — `sleepWithBackoff(0, {initialDelayMs})` misused as a plain sleep; reads as backoff but the growth lives in the prime array.
- 🟡 **MEDIUM** *(convergent: paul + xp)* — `toBe(9)` is a magic number coupled to the prime list, and the mocked `sleepWithBackoff` means the budget/timing is never actually exercised.
- 🟢 **LOW** — "~90s" is really ~77s of backoff; 401/403 ride the same retry path as outages, with no metric on fail-closed denials.

### Convergence
- paul + xp independently flagged the test's `toBe(9)` magic number / untested timing — highest-confidence nit.
- paul + xp both circled the budget-vs-prime-array design (paul on the comment accuracy, xp on the redundant guard / dead clamp).

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 👤 paul | right shape (sync happy-path + background retry + fail-closed), but close the stale-loop race before merge; rest are nits. |
| 📐 xp | correctness holds; main weakness is expressiveness — collapse the two stop conditions to one and give the timing a real test. |
| 🛡 security-audit | no bypass: every success reflects the live server answer, fail-open is a bounded ~90s retention of a genuine prior grant, and it deterministically fails closed. |

---

pauldambra:
> [!NOTE]

Triaged the QA Swarm review and pushed fixes in `1c984415`:

- 🟠 **HIGH race** — *fixed*. Each authoritative check bumps a sequence; the background retry loop bails if superseded (or the session ended) before it can write state, so a stale loop can no longer clobber a fresh `true` → `false`.
- 🟡 **Redundant budget + dead clamp** — *fixed*. Dropped the 90s budget; the loop now just walks the prime schedule (~77s), removing the unreachable `Math.min` clamp.
- 🟡 **`sleepWithBackoff(0,…)` misuse** — *fixed*. Added a real `sleep()` to `@posthog/shared` and use it.
- 🟡 **Magic `9` / untested timing** — *fixed*. The count is now derived/commented, and a new assertion checks the actual prime-backoff schedule is exercised; added a regression test for the race.
- 🟢 **"~90s" comment** — *fixed* (now states ~77s).
- 🟢 **401/403 share the outage path + no fail-closed metric** — *deferred* (judgement calls): short-circuiting auth-class statuses and adding a denial counter are worth doing but belong in a follow-up, not this fix.

security-audit found no gate-bypass, so nothing to action there.

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Addressed in `fe45fb05`: a **401** is now a hard stop — we log the user out (re-auth) instead of retrying with a rejected token and dropping them on the fail-closed invite screen. A **403** stays in the transient/retry path, since during an outage it isn't a reliable "access revoked" signal. Added a test covering the 401 logout.

The observability half (a counter/metric on fail-closed denials — "log before you limit") is left as a follow-up; it needs a metrics sink that is out of scope for this fix. Resolving the thread on that basis.

pauldambra:
* **#2944** <a href="https://app.graphite.com/github/pr/PostHog/code/2944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2862** <a href="https://app.graphite.com/github/pr/PostHog/code/2862?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2862?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/core/src/auth/auth.ts
```diff
@@ -910,32 +911,108 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
-      this.logger.warn("Failed to update code access state", { error });
-      this.updateState({ hasCodeAccess: false });
+      this.logger.warn("Failed to reach code access check", { error });
+      return false;
+    }
+  }
+  // Background retry loop with prime-number backoff. Stepping through primes
+  // (rather than a doubling schedule) keeps many clients' retries from lining
+  // up into bursts after a shared outage. The prime sequence is the whole
+  // budget — it sums to ~77s of waiting — after which we fail closed so a
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/performance]** 🟢 LOW

the prime sequence `[2,3,5,7,11,13,17,19,23]` sums to **100s**, not the "~77s" this comment (and the one on line 1006) claims — 77s is the sum of only the first 8 primes, but the loop sleeps on all 9 before the final check + fail-closed. the commit message says "~90s". three different numbers describe the same budget. worth correcting to ~100s so the real fail-closed window is clear (and note the true wall-clock is even longer once you add up to 10s of `AbortSignal.timeout` fetch time per attempt — so don't let ~77s/~100s be load-bearing anywhere downstream).

packages/core/src/auth/auth.ts
```diff
@@ -910,32 +911,108 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+    for (const delayMs of AuthService.CODE_ACCESS_BACKOFF_PRIMES_MS) {
+      await sleep(delayMs);
+      if (this.codeAccessCheckSeq !== seq || !this.session) return;
+      if (await this.checkCodeAccessOnce()) return;
     }
+
+    if (this.codeAccessCheckSeq !== seq || !this.session) return;
+    // Could not confirm access within the retry window: fail closed so the
+    // invite gate can't be bypassed by keeping the client offline.
+    this.logger.warn("Code access check failed within retry window, failing closed");
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/data-integrity]** 🟡 MEDIUM

this is the user-impacting transition — an active user gets flipped to the invite screen — but it only logs at `warn`, and the recover-after-transient path logs nothing. when this fires in the wild we'll be asked "how many real users got bounced by a transient outage vs genuinely lost access?" and we can't answer it: the 401-logout, the fail-closed, and the eventual-recovery are three distinct outcomes worth counting separately. under a sustained partial outage this fix degrades quietly into the very ejection it's meant to prevent, ~100s later. consider an `error`-level log or a captured event on the fail-closed path so it's alertable. also worth thinking about the UX: the binary `hasCodeAccess` flips under the renderer with no "reverifying access" state — and since this is a gate change, rolling it behind a flag / to a % first would de-risk it. not blocking.

packages/core/src/auth/auth.ts
```diff
@@ -910,32 +911,108 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+        session.accessToken,
       );
+
+      if (response.status === 401) {
+        // The token is rejected — retrying can't fix it. Hard-stop and log out
+        // so the user re-authenticates instead of being retried into a
+        // fail-closed invite screen. (A 403 stays in the transient path below:
+        // during an outage it's not a reliable "your access is gone" signal.)
+        this.logger.warn("Code access check returned 401, logging out");
+        await this.logout();
```
pauldambra, line 954: > [!NOTE]

**[convergent: qa-team/reliability + xp]** 🟡 MEDIUM

`checkCodeAccessOnce()` is awaited from inside `syncAuthenticatedSession` (via `updateCodeAccessFromSession`), which runs inside the shared `refreshPromise`. on a 401 it `await this.logout()`, nulling `this.session` and publishing anonymous state mid-refresh; control then returns to `syncAuthenticatedSession` which continues, and `refreshPromise` still resolves with the now-logged-out session — so a concurrent `getValidAccessToken()` caller can receive a token for a session that was just torn down. worth confirming that transient is harmless, or guarding `syncAuthenticatedSession` against `this.session` having been cleared during the awaited check. (related smell: a method named `checkCodeAccessOnce` that reads as a query is also performing a `logout()` command — you could return the 401 as a third outcome and let the caller own the logout, keeping this method to "classify the response".)

pauldambra (commented):

packages/core/src/auth/auth.ts
```diff
@@ -910,32 +911,108 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+  // (rather than a doubling schedule) keeps many clients' retries from lining
+  // up into bursts after a shared outage. The prime sequence is the whole
+  // budget — it sums to ~77s of waiting — after which we fail closed so a
+  // sustained inability to call home revokes access rather than granting it
+  // forever. `seq` guards against a newer authoritative check (or logout)
+  // having superseded this loop while it slept.
+  private async runCodeAccessRetry(seq: number): Promise<void> {
+    for (const delayMs of AuthService.CODE_ACCESS_BACKOFF_PRIMES_MS) {
+      await sleep(delayMs);
+      if (this.codeAccessCheckSeq !== seq || !this.session) return;
```
pauldambra: > [!NOTE]

**[xp]** 🟡 MEDIUM

the staleness predicate `this.codeAccessCheckSeq !== seq || !this.session` is duplicated here and again before the fail-closed write (line 995) — extract `isStale(seq)`, which also names the concept the comment is explaining. beyond DRY there's a subtle ordering question it hides: the loop checks staleness *after the sleep, before the check*, but not *after* `checkCodeAccessOnce` returns "retry" inside the loop. so a fresh grant landing *during* the final in-flight check is only caught by the post-loop guard (B). the existing supersession test parks on the *first* sleep, so it doesn't exercise that window — adding a case where the fresh grant lands during the LAST `checkCodeAccessOnce` would prove guard B actually closes it rather than just narrowing it.

packages/core/src/auth/auth.ts
```diff
@@ -910,32 +911,108 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+      void this.runCodeAccessRetry(seq);
+    }
+  }
+  // Returns true when the check reached a definitive end and the caller must
+  // NOT retry: either a 200 (which updates the gate from `has_access`) or a
+  // 401 (the token itself is rejected, so we hard-stop and log out). Returns
+  // false only for transient failures worth retrying — a 403 (which during an
+  // outage looks transient), 429, 5xx, or a thrown network error. Genuine
+  // "no access" is a 200 with `has_access: false`, never a non-2xx status.
+  private async checkCodeAccessOnce(): Promise<boolean> {
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

`checkCodeAccessOnce` returns a bare `boolean` whose meaning ("definitive, don't retry" vs "transient, retry") only lives in the 6-line comment above it — that's primitive obsession the type could carry instead. something like `type CodeAccessOutcome = "settled" | "retry"` makes `if (resolved !== "settled")` and `if (await this.checkCodeAccessOnce() === "settled") return` read themselves, and the first half of the doc comment evaporates (rename over comment, per the repo's no-comment house style). the remaining "genuine no-access is 200 + has_access:false, never a non-2xx" sentence is real *why* — keep that.

packages/core/src/auth/auth.ts
```diff
@@ -910,32 +911,108 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+            AuthService.CODE_ACCESS_FETCH_TIMEOUT_MS,
+          ),
+        },
+        session.accessToken,
       );
+
+      if (response.status === 401) {
+        // The token is rejected — retrying can't fix it. Hard-stop and log out
+        // so the user re-authenticates instead of being retried into a
+        // fail-closed invite screen. (A 403 stays in the transient path below:
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/security]** 🟢 LOW

the whole transient-vs-definitive split rests on "a 403 is transient; genuine revocation is always a 200 `{has_access:false}`, never a 403". that's a reasonable read, but it means a genuinely-revoked user whose server returns 403 now gets ~100s of grace before being gated, where before they were ejected immediately. safe *only* as long as that backend contract holds. worth one line of confirmation with whoever owns `check-access/` (and ideally a backend test pinning the 200/403 semantics) so the coupling can't be silently broken later.

packages/core/src/auth/auth.ts
```diff
@@ -910,32 +911,108 @@ export class AuthService extends TypedEventEmitter<AuthServiceEvents> {
...
+      }
+
+      if (!response.ok) {
+        this.logger.warn("Transient failure checking code access", {
+          status: response.status,
+        });
+        return false;
+      }
+
       const data = (await response.json().catch(() => ({}))) as {
```
pauldambra: > [!NOTE]

**[qa-team/generalist]** 🟢 LOW

a 200 with a malformed/empty body resolves `has_access` to `undefined`, so `data.has_access === true` is false and the user is immediately gated (returns true, no retry). a transient proxy returning 200 with a truncated/HTML body would thus eject the user for that cycle rather than being treated as transient. low impact since the next hourly refresh recovers — but if you want symmetry with the network-error path, treating an unparseable 200 as transient (`return false`) would route it through the retry loop instead of failing closed for the cycle.

packages/shared/src/backoff.ts
```diff
@@ -29,3 +29,10 @@ export function sleepWithBackoff(
   const delay = getBackoffDelay(attempt, options);
   return new Promise((resolve) => setTimeout(resolve, delay));
 }
+
+/**
+ * Sleep for a fixed number of milliseconds.
+ */
+export function sleep(ms: number): Promise<void> {
```
pauldambra: > [!NOTE]

**[convergent: xp + qa-team + paul]** ⚪ NIT

`sleep(ms)` duplicates the `new Promise((resolve) => setTimeout(resolve, …))` body already in `sleepWithBackoff` just above. it's a legitimately distinct primitive (fixed delay vs computed backoff), but you can say the setTimeout idea once by having `sleepWithBackoff` delegate: `return sleep(getBackoffDelay(attempt, options))`. (tiny naming aside: a bare `sleep` isn't really *backoff* — fine to leave it here, but a general `async`/`time` primitives home would read slightly better.)

pauldambra:
> [!NOTE]
>

A careful, well-tested resilience fix to a hot-path auth gate. The synchronous-healthy / background-retry split is the right shape, the monotonic `codeAccessCheckSeq` is a clean lock-free supersession primitive, the security posture is correct, and the 7 new tests assert the *why* (no clobber, backoff actually exercised) not just the *what*. **No CRITICAL or HIGH findings.** The follow-ups are observability- and polish-shaped, none blocking.

### Key findings

- 🟡 **MEDIUM** — Fail-closed eviction only logs at `warn`, with no recovery signal — under a sustained partial outage this degrades quietly into the very user-ejection it prevents, ~100s later. Wants an alertable log/event (and ideally a flag/% rollout + a "reverifying" UI state so the flip isn't silent). `auth.ts:998`
- 🟡 **MEDIUM** — `await this.logout()` on a 401 fires inside the shared `refreshPromise`; a concurrent `getValidAccessToken()` caller can get a token for a session that was just torn down. `auth.ts:960`
- 🟡 **MEDIUM** — Duplicated staleness guard worth extracting, and it hides a final-iteration ordering window: a fresh grant landing during the *last* `checkCodeAccessOnce` isn't covered by the existing supersession test. `auth.ts:991`
- 🟢 **LOW** — Backoff comments say "~77s" but the primes sum to **100s** (commit says "~90s") — three numbers for one budget. `auth.ts:984`
- 🟢 **LOW** — 403-is-transient rests entirely on the backend contract "revocation is always a 200, never a 403" — worth confirming/pinning. `auth.ts:957`
- 🟢 **LOW** — `boolean` return wants a named type (`"settled" | "retry"`), which deletes the paraphrasing comment. `auth.ts:937`
- 🟢 **LOW** — A malformed/empty 200 body fails the user closed for that cycle instead of being treated as transient. `auth.ts:971`
- ⚪ **NIT** — `sleep` duplicates `sleepWithBackoff`'s `setTimeout` body; let `sleepWithBackoff` delegate. `backoff.ts:36`

### Convergence

- **Fail-closed observability / silent eviction** — flagged independently by paul and qa-team/data-integrity.
- **403-is-transient backend-contract coupling** — paul and qa-team/security.
- **Budget number wrong (~77s vs 100s)** — paul and qa-team/performance (arithmetic verified: 2+3+5+7+11+13+17+19+23 = 100).
- **`await logout()` mid-refresh** — qa-team/reliability and xp.
- **`sleep` duplicates `sleepWithBackoff`** — xp, qa-team, and paul.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Careful, well-tested fix; no CRITICAL/HIGH; main gaps are observability on the fail-closed path and a mid-refresh `logout()` race. |
| 👤 paul | Really nice change and strong tests; ship it once the silent invite-screen flip has some UI feedback + observability, and the 403 contract is confirmed. |
| 📐 xp | Passes all four rules with only minor once-and-only-once / primitive-obsession nicks; most of the added comments are genuine *why* and should stay. |
| 🛡 security-audit | **0 findings** — the gate can only open on a genuine backend allow; no failure path sets it open, the 401-logout invalidates the session, and seq/`!session` guards close the race. |

---

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Fixed in `bb49f240b` — corrected the `~77s` comment to `~100s` on this line (the prime sequence 2+3+5+7+11+13+17+19+23 = 100).

packages/shared/src/backoff.ts
pauldambra: > [!NOTE]

Fixed in `bb49f240b` — made `sleepWithBackoff` delegate to `sleep` so the `Promise`/`setTimeout` body is expressed once: `return sleep(getBackoffDelay(attempt, options))`.

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Fixed in `372b1d618`. The 401 path no longer hands a torn-down session to awaiting callers: `refreshAndSync` now rejects with `NotAuthenticatedError` when the code-access check logs us out mid-sync (`this.session !== session`). Added a test that a refresh whose check 401s rejects rather than resolving with a dead token.

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Done in `372b1d618` — extracted `codeAccessCheckIsStale(seq)` for the duplicated supersession guard. The post-loop guard (B) already closes the final-iteration window; added a test that honours a success on the 9th (final) retry instead of failing closed, exercising that boundary.

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Done in `372b1d618` — `checkCodeAccessOnce` now returns a named `CodeAccessCheckOutcome` (`"settled" | "retry"`) instead of a bare boolean, and the paraphrasing half of the doc comment is gone.

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Fixed in `372b1d618` — an unreadable 200 body (`json()` rejects) is now treated as a transient failure (`retry`) instead of gating the user, symmetric with the network-error path. Added a test covering recovery after an unparseable 200.

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Confirmed against the backend rather than changing the client: `CodeInviteViewSet.check_access` (posthog `products/tasks/backend/presentation/views/api.py`) always returns `Response({"has_access": bool})` at HTTP 200 — genuine denial is `200 {has_access:false}`, and non-2xx only ever comes from the auth/permission framework or an outage, never as a deliberate "no access". So treating 403 as transient is correct per the contract. Resolving.

packages/core/src/auth/auth.ts
pauldambra: > [!NOTE]

Update: reverted in \`5ee6da135\`. On reflection this was treated as speculative (YAGNI) — an authenticated API returning a 200 with an unparseable body is unlikely enough not to warrant a dedicated transient branch. A 200 that does not cleanly parse to `{has_access: true}` now simply gates (settled), and the response is parsed via a zod schema rather than a hand cast.

pauldambra:
Closing this as **superseded by #2931** ("treat transient code-access checks as indeterminate, not denied"), which merged to `main` first and fixes the same root cause: a transient/non-2xx/offline code-access check no longer denies access and bounces an active user to the invite screen. #2931's `checkCodeAccess()` returns `boolean | null` — authoritative only on a 2xx with an explicit boolean `has_access`, indeterminate otherwise — retries with backoff, and keeps the prior value when inconclusive.

Since the headline fix is already shipped, this PR is redundant and conflicts with `main` in the same region, so I'm closing rather than landing a competing implementation.

Two behavioural differences this PR carried that #2931 deliberately does **not**, noted here in case they're worth a follow-up:

1. **Fail closed after a sustained budget** — #2931 keeps the prior value indefinitely while the check stays inconclusive; this PR eventually revoked access so the invite gate can't be held open by staying offline.
2. **401 → hard logout** — #2931 treats a 401 as just another indeterminate result; this PR logged the user out so they re-authenticate.

If either is actually wanted, it's a small follow-up layered on top of #2931's `checkCodeAccess`, not a reason to keep this branch open.

PR #2867, opened 2026-06-23
fix(inbox): keep detail back link pointing at the tab you came from

rafaeelaudibert:
## Problem

After archiving a report while viewing it, the back button flipped from **"Back to reports"** to **"Back to archive"**. The usual flow is: open Reports → open a report → maybe archive → go back. So you'd expect to land back in Reports, not Archive.

This wasn't a label being swapped — the architecture flips it implicitly. Archiving sets the report's status to `suppressed`, and `InboxReportDetailGate` redirects (`replace: true`) from `/code/inbox/reports/$reportId` to `/code/inbox/dismissed/$reportId`. That route mounts `DismissedReportDetail`, which hard-codes `backTo="/code/inbox/dismissed"` / "Back to archive". So the back link followed the report's *new state* instead of the path you took in.

The redirect is load-bearing (it keeps archived reports off the triage view), so it's kept — but it now **remembers where you came from**.

## Change

- **New `useInboxBackTarget` hook** — owns the `InboxListRoute` / `InboxBackTarget` types, augments TanStack Router's `HistoryState` with an `inboxBackOrigin` field, and resolves the back link from history state (validating the untyped state at the boundary), with a fallback.
- **`InboxReportDetailGate`** — stamps the origin (`{ to, label }`, always a pipeline route at that point) into navigation `state` on the to-Archive redirect. Gains optional display-only `backLinkTo`/`backLinkLabel`; `backTo` stays the route identity that drives the redirect and engagement tracking.
- **`DismissedReportDetail`** — reads the recorded origin and uses it for both the missing-shell link and the header link; defaults to "Back to archive".
- **`InboxDetailFrame`** — `backTo` widened to the shared `InboxListRoute` union.
- **Unit test** for the history-state validator.

## Behavior

| Path in | Back link |
| --- | --- |
| Reports → open → archive | **Back to reports** → Reports list |
| Pulls → open → archive | **Back to pull requests** |
| Runs → open → archive | **Back to runs** |
| Archive tab / deep link / refresh | **Back to archive** (fallback) |

## Verification

⚠️ I could **not** run `pnpm typecheck`/tests locally — the sandbox's network is throttled to single-digit KiB/s and `pnpm install` gets OOM-killed before `@tanstack/react-router` and `vitest` are installed. Verified by manual review. Two things worth a CI/reviewer eye:
- The `declare module "@tanstack/react-router" { interface HistoryState }` augmentation (documented TanStack v1 pattern; mirrors the existing `Register` augmentation in `router.ts`).
- `useLocation({ select })` typing.

### DIFF
--- a/packages/ui/src/features/inbox/components/DismissedReportDetail.tsx
+++ b/packages/ui/src/features/inbox/components/DismissedReportDetail.tsx
@@ -8,6 +8,10 @@ import { Button } from "@posthog/quill";
 import type { SignalReport } from "@posthog/shared/types";
 import { InboxDetailFrame } from "@posthog/ui/features/inbox/components/InboxDetailFrame";
 import { InboxReportDetailGate } from "@posthog/ui/features/inbox/components/InboxReportDetailGate";
+import {
+  type InboxBackTarget,
+  useInboxBackTarget,
+} from "@posthog/ui/features/inbox/hooks/useInboxBackTarget";
 import { useInboxRestoreReport } from "@posthog/ui/features/inbox/hooks/useInboxRestoreReport";
 import { copyInboxReportLink } from "@posthog/ui/features/inbox/utils/copyInboxReportLink";
 import { Spinner } from "@radix-ui/themes";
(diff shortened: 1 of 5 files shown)


rafaeelaudibert:
* **#2867** <a href="https://app.graphite.com/github/pr/PostHog/code/2867?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2867?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2885, opened 2026-06-23
feat(sessions): inline model swap on plan-approval card

dmarticus:
## Summary
- Adds a small `Model: [picker ▾]` row above the plan-approval card so users can swap models (e.g. Sonnet → Opus) without leaving the flow.
- Re-uses the existing `ModelSelector` (same `sessionService.setSessionConfigOption` path as the task-input picker) and the existing `useSessionTaskId` context — no new wiring, no new state.
- Picker disappears once the plan is approved/rejected; the execution decision is locked at that point.

## Test plan
- [x] On a Claude session, propose a plan via the agent and confirm a `Model` row appears above the blue plan card.
- [x] Swap the model from the row; confirm the dropdown updates and the swap takes effect for the upcoming turn.
- [x] Approve the plan; confirm the model row disappears and the existing "Plan approved" status renders as before.
- [x] Reject the plan; confirm the model row disappears and the "(Plan rejected)" status renders as before.
- [x] On a Codex session (no model select option), confirm the row simply doesn't render — `ModelSelector` already short-circuits to `null`.


dmarticus:
note that ideally we should have this part be auto-derived from the gateway, but I guess failing that I don't hate the idea of someone being able to choose their model (i mean it might cost us a helluva lot more down the line).

daniloc (approved):
hooray

PR #2889, opened 2026-06-24
feat(agents): editable agent.md and SKILL.md on draft revisions

dmarticus:
## Summary

- Adds a per-file **Edit / Save** affordance to `agent.md` and `skills/<id>/SKILL.md` in the configuration pane — gated to **draft** revisions. Ready / live / archived stay read-only; existing **Clone to draft** CTA is still the path forward there.
- Adds a **Paste markdown bundle…** dialog to the revision bar (drafts only) for the bulk-migration use case — paste a `--- path ---` fenced blob, preview new vs update per file, import. Parser is pure + unit-tested.
- Tool `source.ts` / `schema.json` remain read-only this round.

## Why

Today the bundle is GET-only. Porting a multi-file agent (e.g. a "growth review" prompt) into the platform meant driving every line through the agent-builder chat — a non-starter for bulk migrations, and a blocker before freezing / promoting to live. User feedback flagged this directly.

## Server contract (needed before merge)

Two new endpoints on `…/revisions/<id>/`, both **draft-only** (409 otherwise):

- `PUT  …/bundle/file/`     body `{ path, content }` — `path ∈ {agent.md, skills/<id>/SKILL.md}`
- `POST …/bundle/import/`   body `{ agent_md?, skills?: [{id, description?, body}] }` — merge-by-id, no implicit deletion

Both return the updated `AgentRevision`. **Paired PR on the Django repo to follow.** Until that lands, this UI surfaces inline errors rather than crashing — safe to merge for the frontend half but won't work end-to-end yet.

## Test plan

- [ ] [Backend PR](https://github.com/PostHog/posthog/pull/65693) merged and deployed
- [ ] On a draft, open `Instructions` (agent.md) → Edit button visible; ready / live / archived hide it
- [ ] Edit agent.md → Save → rendered view updates, network shows `PUT …/bundle/file/`
- [ ] Open a skill body → same edit flow; works on a skill whose body is empty
- [ ] Click **Paste markdown bundle…** on a draft → paste agent.md + two skill blocks (one existing id, one new) → preview shows mixed update / new badges → confirm → all three reflect after refresh
- [ ] Freeze the draft → promote → confirm the new live revision serves the edited content


daniloc (approved):
yerp, with a caveat:

should this view also SOURCE skill store content and pipe to the display? maybe functionally it doesn't matter if the skill version is pinned

PR #2896, opened 2026-06-24
feat(agent): natively resume prior session on cloud resume

tatoalo:
## Problem

we used to resume restored context by flattening the prior conversation into a markdown summary and replaying it as a synthetic `You are resuming a previous conversation…` prompt. That is lossy (token-budget truncation, tool results clipped) and the agent continued from a recap rather than the real thread.

## Changes

Resume the real session instead of replaying a summary

### DIFF
--- a/packages/agent/src/sagas/resume-saga.test.ts
+++ b/packages/agent/src/sagas/resume-saga.test.ts
@@ -1,5 +1,6 @@
 import type { SagaLogger } from "@posthog/shared";
 import { afterEach, beforeEach, describe, expect, it, type vi } from "vitest";
+import { POSTHOG_NOTIFICATIONS } from "../acp-extensions";
 import type { PostHogAPIClient } from "../posthog-api";
 import { ResumeSaga } from "./resume-saga";
 import {
@@ -8,6 +9,7 @@ import {
   createGitCheckpointNotification,
   createMockApiClient,
   createMockLogger,
+  createNotification,
   createTaskRun,
   createTestRepo,
   createToolCall,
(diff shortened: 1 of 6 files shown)


packages/agent/src/sagas/resume-saga.test.ts
```diff
@@ -553,6 +555,87 @@ describe("ResumeSaga", () => {
...
+        runId: "run-1",
+        repositoryPath: repo.path,
+        apiClient: mockApiClient,
+      });
+
+      expect(result.success).toBe(true);
+      if (!result.success) return;
+      expect(result.data.sessionId).toBeNull();
+    });
+  });
```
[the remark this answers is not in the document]
tatoalo: Done — collapsed the three cases into a single `it.each` and added the sdk-prefixed `_${RUN_STARTED}` case while at it.

packages/agent/src/server/agent-server.ts
```diff
@@ -1365,26 +1447,72 @@ export class AgentServer {
...
+        ? false
+        : await this.applyResumeGitCheckpoint(payload);
+
+      const pendingUserPrompt = await this.getPendingUserPrompt(taskRun);
+      const prompt: ContentBlock[] = pendingUserPrompt?.length
+        ? pendingUserPrompt
+        : [
+            {
+              type: "text",
+              text: "Continue from where you left off. The user is waiting for your response.",
```
skoob13, line 1480: Is there a case when we need a resume message?

skoob13 (approved):
Thanks, PostHog AI is unblocked:love_letter:

packages/agent/src/server/agent-server.ts
tatoalo: The old plimbing is still needed for the fallback path where native resume is unavailable, for example no prior session id, JSONL hydration failed. Kinda defensive but still worth keeping for the moment imho

PR #2900, opened 2026-06-24
feat(agent-applications): model policy config UI + session model

benjackwhite:
## Problem

The agent config pane couldn't show or edit the new `spec.models` (the auto/manual model picker + `optimize_for`) — the `model` node still rendered the legacy `spec.model` string, so new specs showed "not set". There was also no way to see which model a session actually used, or to browse the served models and their costs.

## Changes

Surface and edit `spec.models` in the agent config pane, save it back to a draft, and show the model a session used.

- **Types** (`agent-platform-types`): add `AgentModelPolicy` (auto `level` / manual `models[]`), `AgentModelOptimizeFor`, and `ModelCatalog`; `spec.model` becomes optional (legacy fallback).
- **Model section** (`AgentModelConfig`): an interactive policy editor — `mode` / **`optimize for`** / `level` / `reasoning` as dropdowns with per-option descriptions + icons; an "auto level resolves to" preview with live pricing; and a searchable **model browser** of every served model with its cost profile (input/output/cache $/Mtok, context window, provider; sort by name / cheapest / priciest). Manual mode supports add-from-browser and **drag-to-reorder** (`@dnd-kit`).
  - **`optimize for`** (cost | availability) — the session-stability control mirroring `spec.models.optimize_for`. **Cost** (default) pins the first working model for the whole session (warm prompt cache, no mid-session failover); **Availability** fails over to the next model if the session's model goes down (survives outages, re-reads context cold). Applies to both auto and manual.
- **Save** (`useApplyAgentSpec`): a **create-draft-and-apply** path — patches the spec on a draft in place, or branches a fresh draft from a non-draft revision first (and archives the orphan best-effort if the patch fails). Dirty banner + reset; "Save to new draft" when the source isn't a draft.
- **Catalog** (`useModelCatalog`): fetches the real catalog (`GET …/agent_applications/models/` → janitor → gateway), with a known-levels fallback while loading.
- **Session detail** (`AgentSessionDetailBody`): a **Model** KPI showing the model(s) that answered (one, or `primary +N` on fallback).

Default policy is `auto / medium / cost`.

## How did you test this?

- `pnpm --filter @posthog/ui --filter @posthog/shared typecheck` — clean.
- `pnpm biome check` on changed files — clean.
- `useApplyAgentSpec` unit tests (PATCH-in-place, clone-then-PATCH, orphaned-draft archive-on-failure) pass.
- Drove the running dev app over CDP and screenshotted the model section (auto + manual), the dropdowns, and the manual drag-to-reorder list against a live agent. (Screenshot above predates the `optimize for` row — will refresh.)

## Automatic notifications

- [ ] Publish to changelog?


packages/ui/src/features/agent-applications/components/AgentModelConfig.tsx
```diff
@@ -0,0 +1,747 @@
...
+    const filtered = needle
+      ? models.filter(
+          (m) =>
+            m.model.toLowerCase().includes(needle) ||
+            m.provider.toLowerCase().includes(needle),
+        )
+      : models;
+    const sorted = [...filtered];
+    if (sort === "name") sorted.sort((a, b) => a.model.localeCompare(b.model));
+    if (sort === "cheapest") sorted.sort((a, b) => a.input - b.input);
```
dmarticus: Nit: sorting `cheapest` / `priciest` by `input` alone can mis-rank models where input is cheap but output is expensive (reasoning models in particular). A blended cost (e.g. `input + output`, or weighted by typical output share) would match what users actually optimize for. Worth a tooltip on the segmented control if you keep input-only, so the heuristic is explicit.

packages/ui/src/features/agent-applications/components/AgentModelConfig.tsx
```diff
@@ -0,0 +1,747 @@
...
+  models,
+  byId,
+  onChange,
+}: {
+  models: AgentModelEntry[];
+  byId: Map<string, ModelCatalogEntry>;
+  onChange: (next: AgentModelEntry[]) => void;
+}) {
+  const ids = models.map((m) => m.model);
+  const handleDragOver: DragDropEvents["dragover"] = (event) => {
```
dmarticus, line 349: Nit: reordering on `dragover` fires `setManual` on every pointer move that crosses a target row — works, but unconventional and can thrash state on longer lists. The more common @dnd-kit shape is to keep visual order in `state` during the drag and commit the final array on `dragend`. Probably fine for the short lists this surface will see; flagging in case manual lists grow.

packages/ui/src/features/agent-applications/hooks/useApplyAgentSpec.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+// Capture the mutationFn the hook hands to react-query so we can exercise the
+// create-draft-vs-patch branching directly, without a live QueryClient.
+let mutationFn: (vars: {
+  revision: { id: string; state: string };
+  spec: unknown;
+}) => Promise<unknown>;
+
+vi.mock("@tanstack/react-query", () => ({
+  useMutation: (opts: { mutationFn: typeof mutationFn }) => {
```
dmarticus: Nit: mocking `useMutation` itself lets the tests exercise the closure's branching cleanly, but `onSuccess`'s invalidation keys never get touched — if the query keys here ever drift from the keys the corresponding fetch hooks use, tests stay green and the UI silently stops invalidating. One integration-style test through a real `QueryClient` (or asserting the key strings against the shared `agentApplicationsKeys` factory) would close that gap.

dmarticus (approved):
Verified the spec.model / spec.entrypoint loosening — no other consumers in packages/ or apps/, so safe. Pushed a fix for the Tailwind size collision on the mono Model KPI (e3417045b). LGTM otherwise — three non-blocking nits inline.

PR #2905, opened 2026-06-24
feat(canvas): toast when a backgrounded canvas generation finishes

adamleithp:
## What

image

--- 

Screenshot 2026-06-24 at 16 41 01

Surfaces a global application toast — with a **View canvas** link — the moment a freeform canvas generation finishes, so the user is called back to it from **anywhere in the app**, not just while sitting on the canvas.

- **Tracking** is in-memory, registered at generation start (`canvasGenerationTrackerStore`), rather than the dashboard's `generationTaskId` (which is never cleared for freeform canvases).
- **Completion detection** reuses the same signal the canvas view uses — `isCanvasGenerating` (live ACP session `isPromptPending` for local runs, `cloudStatus` for cloud). Extracted into tested helpers and shared with `FreeformCanvasView`, removing its duplicated inline logic.
- **Armed latch** (`hasCanvasGenerationStarted`): a task can only toast after concrete "has started running" evidence, so the brief create→connect gap can't fire a false completion.
- **Mount point**: `CanvasGenerationToaster` (renders `null`) is a sibling of `<RouterProvider>` in `App.tsx`, so it stays mounted across every route. `toast.error` extended to carry an action link.

Toast: `completed` → success "X is ready" + View canvas; `failed` → error + link; `cancelled` → silent.

## Testing

- 33 unit tests across the status helpers (local-session, cloud, create→connect race, status resolution).
- Full canvas suite green, typecheck + lint clean.

## Known limitation

Tracks generations started in **this client session** — a quit/reload mid-generation, or another client's generation, won't toast. Persisting that needs a backend "recently completed" signal (out of scope).

### DIFF
--- a/packages/ui/src/features/canvas/freeform/FreeformCanvasView.tsx
+++ b/packages/ui/src/features/canvas/freeform/FreeformCanvasView.tsx
@@ -17,8 +17,10 @@ import {
   EmptyMedia,
   EmptyTitle,
 } from "@posthog/quill";
-import { isTerminalStatus } from "@posthog/shared/domain-types";
-import { isCanvasGenerationRunning } from "@posthog/ui/features/canvas/freeform/canvasGenerationStatus";
+import {
+  isCanvasGenerating,
+  isCanvasGenerationRunning,
+} from "@posthog/ui/features/canvas/freeform/canvasGenerationStatus";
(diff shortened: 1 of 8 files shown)


adamleithp:
Closed, tasks already have notifications, this seems redundant

PR #2908, opened 2026-06-24
feat(tasks): bundle local skills for cloud runs

tatoalo:
## Problem

for cloud tasks, local skill invocations only sent the slash command text. User-local and repo-local skills were available in the local editor but not in the cloud sandbox

## Changes

- let's bundle uploadable local skills into zip artifacts with SHA-256 metadata and send them through the existing cloud artifact upload flow.

- install `skill_bundle` artifacts in the agent server before prompt delivery, inject the bundled skill instructions for the invoked turn, and suppress the raw slash command from the model-facing prompt so local skills behave like actual skills rather than unsupported commands

- refresh slash command metadata when a follow-up command is submitted, so newly created local skills can be selected and invoked without starting a brand-new task.

closes https://github.com/PostHog/code/issues/2260

### DIFF
--- a/apps/code/src/renderer/di/bindings.ts
+++ b/apps/code/src/renderer/di/bindings.ts
@@ -55,7 +55,11 @@ import {
   GITHUB_CONNECT_CLIENT,
   type GithubConnectClient,
 } from "@posthog/core/onboarding/identifiers";
-import { CLOUD_ARTIFACT_READ_FILE_AS_BASE64 } from "@posthog/core/sessions/cloudArtifactIdentifiers";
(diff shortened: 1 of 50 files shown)


tatoalo:
Superseded by a stacked split for easier review (same change, rebased onto current main):

- #2922 — contracts & metadata types
- #2923 — local skill bundler (+ tRPC route)
- #2924 — sandbox install/extract/apply
- #2925 — cloud prompt plumbing
- #2926 — composer integration & DI wiring

Closing in favour of the stack.

PR #2920, opened 2026-06-25
fix(canvas): don't flash empty state while canvas is loading

raquelmsmith:
## Problem

When viewing a canvas that has content, it briefly flashes "This canvas is empty…" before the canvas loads in. The empty-state message keyed only on the working-copy `code` being empty, with no guard for the record still being fetched from the server.

**Why:** A non-empty canvas shouldn't show an "empty" message just because the result hasn't come back from the server yet.

## Changes

In `FreeformCanvasView`, gate the empty state on the record actually being loaded:

- Capture the `dashboards.get` record's `isLoading` (the query was already running).
- Show a loading spinner while the record is still fetching, or in the one-render gap before a record that already has `code` syncs into the thread.
- The "This canvas is empty…" message now only appears once the record has loaded and genuinely has no code.

The loading state is a quill `<Empty>` with the same spinner the generating state uses.

## How did you test this?

- Typechecked the canvas feature (`@posthog/ui`) — no errors in the changed file.
- Biome lint clean on the edited file.

(Note: full repo typecheck has pre-existing errors from unbuilt sibling packages, unrelated to this change.)

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/canvas/freeform/FreeformCanvasView.tsx
+++ b/packages/ui/src/features/canvas/freeform/FreeformCanvasView.tsx
@@ -71,7 +71,7 @@ export function FreeformCanvasView({
   // The generation-task association lives in the canvas record's meta. Poll it
   // while a task is running so the published code + the cleared association show
   // up without a manual refresh (WebsiteDashboard re-syncs the working copy).
-  const { data: dashboard } = useQuery(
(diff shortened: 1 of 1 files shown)


raquelmsmith:
tested, works

PR #2923, opened 2026-06-25
feat(tasks): bundle local skills into sha256-verified zip artifacts

tatoalo:
## Problem

for cloud tasks, local skill invocations only sent the slash command text. User-local and repo-local skills were available in the local editor but not in the cloud sandbox

## Changes

- let's bundle uploadable local skills into zip artifacts with SHA-256 metadata and send them through the existing cloud artifact upload flow.

- install `skill_bundle` artifacts in the agent server before prompt delivery, inject the bundled skill instructions for the invoked turn, and suppress the raw slash command from the model-facing prompt so local skills behave like actual skills rather than unsupported commands

- refresh slash command metadata when a follow-up command is submitted, so newly created local skills can be selected and invoked without starting a brand-new task.

closes https://github.com/PostHog/code/issues/2260

PR #2924, opened 2026-06-25
feat(tasks): install and apply skill bundle artifacts in the sandbox

tatoalo:
## Problem

for cloud tasks, local skill invocations only sent the slash command text. User-local and repo-local skills were available in the local editor but not in the cloud sandbox

## Changes

- let's bundle uploadable local skills into zip artifacts with SHA-256 metadata and send them through the existing cloud artifact upload flow.

- install `skill_bundle` artifacts in the agent server before prompt delivery, inject the bundled skill instructions for the invoked turn, and suppress the raw slash command from the model-facing prompt so local skills behave like actual skills rather than unsupported commands

- refresh slash command metadata when a follow-up command is submitted, so newly created local skills can be selected and invoked without starting a brand-new task.

closes https://github.com/PostHog/code/issues/2260

### DIFF
--- a/packages/agent/src/adapters/claude/claude-agent.ts
+++ b/packages/agent/src/adapters/claude/claude-agent.ts
@@ -459,6 +459,10 @@ export class ClaudeAcpAgent extends BaseAcpAgent {
       promptReplayed = true;
     }
 
+    if (commandMatch && !isLocalOnlyCommand) {
+      await this.refreshSlashCommandsForPrompt(commandMatch[1]);
+    }
+
     if (this.session.promptRunning) {
       const isSteer = isSteerMeta(params._meta);
       if (isSteer) {
@@ -2164,6 +2168,7 @@ export class ClaudeAcpAgent extends BaseAcpAgent {
 
   private async sendAvailableCommandsUpdate(): Promise<void> {
     const commands = await this.session.query.supportedCommands();
+    this.session.knownSlashCommands = collectKnownSlashCommands(commands);
     const available = getAvailableSlashCommands(commands);
     await this.client.sessionUpdate({
       sessionId: this.sessionId,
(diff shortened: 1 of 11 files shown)


tatoalo:
* **#2926** <a href="https://app.graphite.com/github/pr/PostHog/code/2926?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2925** <a href="https://app.graphite.com/github/pr/PostHog/code/2925?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2924** <a href="https://app.graphite.com/github/pr/PostHog/code/2924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/2924?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2923** <a href="https://app.graphite.com/github/pr/PostHog/code/2923?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2922** <a href="https://app.graphite.com/github/pr/PostHog/code/2922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/agent/src/adapters/claude/claude-agent.ts
```diff
@@ -2175,6 +2180,24 @@ export class ClaudeAcpAgent extends BaseAcpAgent {
     this.updateBreakdownCategory("skills", estimateSkillsTokens(available));
   }
 
+  private async refreshSlashCommandsForPrompt(command: string): Promise<void> {
+    const commandName = command.slice(1);
+    if (this.session.knownSlashCommands?.has(commandName)) {
+      return;
+    }
+
```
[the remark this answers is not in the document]
tatoalo: Intentional: `sendAvailableCommandsUpdate` is the single place that fetches `supportedCommands()` and pushes them to the client, so caching what we just advertised in `knownSlashCommands` keeps one authoritative source. Splitting it would create a second code path that could drift from what the client was actually told. Keeping as-is.

packages/agent/src/adapters/claude/conversion/acp-to-sdk.ts
```diff
@@ -80,6 +80,18 @@ function transformMcpCommand(text: string): string {
   return text;
 }
 
+function isLocalSkillCommandChunk(
```
k11kirky, line 83: This feels kinda risky to me what if a the string is "hello please run /skill_name - thank you" could the tokens split the skill name?

tatoalo: chunk is an ACP prompt content block, not model tokens, and the matcher is anchored to a slash command at the start of the trimmed block so that ^ example would not fly

It also can't reach this as a "skill" at all since `localSkillName` is only set when the leading text block parses as /<name> and <name> exactly equals an installed skill_bundle artifact's skill_name. The name is captured  and compared by full equality so ther;s no tokenization, no prefix matches, and the skill name never feeds the regex

k11kirky (approved):

k11kirky (approved):

PR #2929, opened 2026-06-25
feat(agent-applications): manage MCP connections in the agent builder

benjackwhite:
Builder UI for the agent-platform MCP layer — **shared connections**, the explicit **agent/principal credential model**, and **per-agent tool permissions**. Pairs with the runtime in **PostHog/posthog#66007**. An owner connects a remote MCP server once via PostHog Code's native `mcp_store` flow, attaches it to an agent, and every asker uses that one shared credential (or, for `principal`, each asker connects as themselves).

## What you can do
- **Add an MCP to an agent** from a connected server, or **connect a new one** via a **popout modal** (`AddCustomServerDialog`, shared with the agent-builder `connect_mcp` punch-out — native OAuth/DCR or api-key flow).
- **Choose the credential model** for an MCP entry via an explicit dropdown — **Agent-level** (a shared `connection`) or **Principal-level** (a per-asker linked identity, surfaced as a fingerprint preview card with a jump to the identity detail). Writes `spec.mcps[].kind` and clears the opposite credential.
- **Edit per-tool permissions** against the connection's **live tool catalog** (`ToolPermissionList`): a connection-wide default (allow / approve / deny) plus per-tool overrides, mapped to the spec's `default_tool_approval` / `level`.
- **Remove an MCP** entry (drops the `mcps[]` ref; the shared installation is untouched).
- All edits persist via draft-branch-then-PATCH (a fresh draft is branched off a non-draft revision, then selected).

## Changes
- **`mcp-server-manager/`** (new shared module): `useMcpConnect` (installations query + connect/reauthorize over the host OAuth callback) and the moved `AddCustomServerForm`. The standalone MCP-servers scene now consumes the shared `mcpKeys`/`createOAuthCallback` (no behaviour change — core mcp-servers tests still green).
- **agent-applications**: editable `McpsOverview` (add-from-connection + connect-new) and `McpBody` — a compact header (auth-mode dropdown + **remove-server**), then the mode-specific config (agent → connection picker; principal → identity selector + preview card) and the `ToolPermissionList` catalog.
- **Legacy `integrations` removed (deliberate).** This PR drops the `integrations` config section (`IntegrationsOverview`/`IntegrationBody`, the `cfg:integrations` node, `SECTION_INFO` entry) **and** the `AgentSpec.integrations?: string[]` field in `packages/shared` — a total removal, not a soft UI-only regression. Nothing reads the field and the agent-platform schema already dropped it; team-OAuth use cases are served by `mcps[].connection` / `identity_providers[]`. (Closes @dmarticus's review note.)
- **Editing foundation, built on `main`**: `api-client.updateAgentRevisionSpec` + `useApplyAgentSpec` (+ test). Intentionally ported compatible with the in-flight `agent-model-policy-ui` branch so the two reconcile cleanly on merge.
- **Drive-by**: cast `spec.entrypoint` in `ModelBody` — fixes a **pre-existing** `@posthog/ui` typecheck error (`AgentSpec`'s `[key: string]: unknown` index signature), unrelated to this change but blocking a clean typecheck.

## Validation
`@posthog/api-client` + `@posthog/ui` typecheck clean (full `pnpm typecheck` green via pre-commit); biome clean; core `mcp-servers` suite (41) + new `useApplyAgentSpec` suite (5) pass.

## Notes / follow-ups
- **Base branch:** targets `main` (the `ben/agent-supported-client-tools` base PR was merged; this branch absorbed master and reconciled the `supported_client_tools` producer, keeping the `connect_mcp` client tool).
- Overlaps the `agent-model-policy-ui` editing work (ported compatibly).
- Component-level tests for the new MCP panels are a follow-up (hooks/logic are covered).

## ⚠️ Outstanding / known gaps before this is production-good

UI add/edit/remove was driven against the live app and works. Remaining concerns:

1. **Tool catalog needs a verified connection.** `ToolPermissionList` now renders the connection's **live tool catalog** with per-tool allow/approve/deny (deny = hidden), so choosing which tools to expose is solved for connection-backed MCPs — but the catalog only populates once the connection is verified/has listed its tools; until then it's empty. Non-connection (principal / BYO) entries still fall back to the legacy already-selected list.
2. **`needs_reauth` not surfaced.** The `connectionMissing` hint only covers "connection isn't in this project". There's no UI for an **expired/revoked** shared connection (the runtime returns `needs_reauth`; see PostHog/posthog#66007 gap #3–4) — the owner gets no prompt to reconnect.
3. **Stale-spec agents can't be edited.** Example agents seeded with an older `models` (plural) spec shape fail the write schema (`'model' is a required property`), so *any* MCP edit 400s. Re-seed to refresh. Pre-existing data issue, not introduced here, but users will hit it on old agents.
4. **Runtime is unproven end-to-end** — see the matching section in PostHog/posthog#66007. The UI can attach a connection, but it hasn't been observed driving a real agent→MCP tool call.

## Agent context
Authored from a Claude Code session (Opus 4.8).


packages/ui/src/features/agent-applications/components/AgentConfigurationPane.tsx
```diff
@@ -192,158 +202,130 @@ function buildTree(spec: AgentSpec, setKeys: string[]): FileTreeNode {
...
+        icon: toolIcon(str(r.kind)),
+        trailing:
+          r.requires_approval === true ? (
+            <LockKeyIcon size={11} className="text-amber-10" />
+          ) : undefined,
+      };
+    }),
+  });
 
+  // Top-level authorable sections always render — even with no entries — so the
```
dmarticus, line 328: The list of always-render sections here conspicuously drops `integrations` — and the rest of the PR removes `IntegrationsOverview`/`IntegrationBody`, the `cfg:integrations` tree node, and the `SECTION_INFO` entry. But `AgentSpec.integrations?: string[]` is still in `packages/shared/src/agent-platform-types.ts:70`.

Net: if any shipped agent has `spec.integrations: ["github"]` today, the field is preserved on the backend but the UI silently stops showing or editing it — a soft regression that's easy to miss because nothing typechecks-broken.

Two clean ways out:
- If no real specs use it: drop the type field in the same PR so the removal is total.
- If some do: keep a read-only render path until the field is properly retired (deprecation note + migration).

Mostly want to make sure this was a deliberate "nobody uses integrations" call rather than collateral from the MCP refactor — the PR description doesn't mention it.

dmarticus (commented):
Just one question on an unstated removal — no blockers on the draft itself, the gaps you list cover it.

packages/ui/src/features/agent-applications/components/AgentConfigurationPane.tsx
benjackwhite: It was never used hence killing it

dmarticus (approved):

PR #2931, opened 2026-06-25
fix(auth): treat transient code-access checks as indeterminate, not denied

k11kirky:
## Problem

The invite-code screen is gated by an in-memory `hasCodeAccess` flag that is re-fetched from `GET /api/code/invites/check-access/` every time the session syncs. `updateCodeAccessFromSession` treated **any** non-success as "not invited":

- a transient throw (network blip, timeout) → `hasCodeAccess: false`
- a non-`true` body (e.g. `{}` returned by a 401) → `hasCodeAccess: false`

This check re-runs mid-session on every token refresh, 401-retry, system sleep/wake, and network reconnect (via `syncAuthenticatedSession`). So a single transient blip — an expired token at check time, a flaky network moment, a slow API response — overwrote a previously-valid `true` with `false`, and `App.tsx` immediately rendered the invite screen. Restarting "fixed" it because init re-ran the check and it succeeded.

In short: a recoverable error was being treated as a hard lockout, and the flag had no indeterminate state to fall back on.

## Fix

Introduce a discriminated `CodeAccessOutcome` (`resolved` vs `indeterminate`) and split the logic into a `checkCodeAccess(session)` helper:

1. **Only an authoritative answer changes state.** `false` (invite screen) is set *only* on a clean 2xx response carrying an explicit boolean `has_access`. A genuinely non-invited user (`{ has_access: false }`) still correctly sees the invite screen.
2. **Everything else is indeterminate and preserves the prior value.** Offline status, network errors, timeouts, non-2xx (incl. 401/403), and unparseable/incomplete bodies no longer clobber a known value — a confirmed `true` stays `true`, a pending `null` stays `null` (UI keeps showing "Checking access…") — and the next sync re-checks.
3. **Transient failures self-heal within the sync** via a bounded retry (`CODE_ACCESS_MAX_ATTEMPTS = 3`) using the existing `sleepWithBackoff` + `REFRESH_BACKOFF`, with a fast exit when offline.

### Deliberate non-change

I considered routing through `authenticatedFetch` (which does 401-refresh-and-retry), but that path calls `getValidAccessToken → initialize()` / `ensureValidSession`, and this method runs *inside* the bootstrap/refresh flow — re-entering that machinery would await an in-flight promise from within itself and **deadlock**. Because `syncAuthenticatedSession` always passes a freshly-minted token, a 401 here is anomalous and is handled as indeterminate (preserve + retry) rather than via a re-entrant refresh. This reasoning is documented in the method's docstring.

## Tests

Adds a `code access resilience` block (5 cases): explicit denial → `false`; network failure on re-check → keeps `true`; 401 on re-check → keeps `true`; inconclusive restore → stays `null` (never auto-denies); 2xx without `has_access` → stays `null`. All 41 auth tests pass; `biome lint` and `pnpm --filter @posthog/core typecheck` are clean.

> Part 1 of a stack. A follow-up PR builds the broader connectivity UX (persistent banner + reconnecting state, consistent git-action gating, auto-reconnect) that this fix is one instance of.

### DIFF
--- a/packages/core/src/auth/auth.test.ts
+++ b/packages/core/src/auth/auth.test.ts
@@ -151,6 +151,9 @@ describe("AuthService", () => {
         string,
         { name: string; projects: { id: number; name: string }[] }
       >;
+      // Overrides the /api/code/invites/check-access/ response (defaults to
+      // granting access). May throw to simulate a network error.
+      checkAccess?: () => Response;
     } = {},
   ) => {
     const accountKey = options.accountKey ?? "user-1";
@@ -186,6 +189,9 @@ describe("AuthService", () => {
           } as unknown as Response;
         }
 
+        if (options.checkAccess) {
+          return options.checkAccess();
+        }
         return {
           ok: true,
           json: vi.fn().mockResolvedValue({ has_access: true }),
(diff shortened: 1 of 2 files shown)


adboio (approved):
have not thoroughly tested, but this looks legit, hopefully fixes the issues we’ve been seeing 🤞💪💪

PR #2932, opened 2026-06-25
feat(connectivity): graceful offline UX — banner, action gating, auto-recovery

k11kirky:
> **Stacked on #2931** (`posthog-code/auth-code-access-resilience`). Review/merge that first; this PR's base is set to its branch, so the diff here is only the connectivity work. GitHub will retarget this to `main` automatically once #2931 merges.

## Why

The code-access fix (#2931) made one recoverable error (a transient invite-check failure) stop locking users out. This PR generalises that instinct: during intermittent connectivity, **block the *actions* that need the network — not the whole app** — keep everything offline-capable working, explain what's paused, and self-heal on reconnect.

The app already had the right instincts unevenly: agent-send is gated on `isOnline`, and detection is robust and server-side (workspace-server polls Google/Cloudflare with a 2-failure hysteresis + backoff, streamed over tRPC). The gaps were: no always-visible status, git remote actions firing blind offline, and no recovery trigger on the reconnect event itself. This PR closes those, reusing the existing machinery rather than rebuilding it.

### 1. Connectivity banner
`ConnectivityBanner` mounted in the shell (`__root.tsx`, all three layouts: main / channels / settings).
- **Offline:** amber strip, "You're offline — network actions are paused, reconnecting automatically", with a **Retry** that forces an immediate probe via a new `ConnectivityClient.checkNow()` (wired to the existing `connectivity.checkNow` route). While the probe runs it shows "Checking connection…".
- **Recovery:** brief green "Back online", then collapses.
- Animated height collapse, modeled on `UpdateBanner`. No re-debouncing — detection is already debounced server-side, so the banner renders the settled state.

### 2. Consistent git-action gating (reusing the "No internet connection" idiom)
- `computeGitInteractionState` (core, pure) gains an `isOnline` input; push / sync / publish and create-PR return the offline reason, which the existing `disabledReason` plumbing renders as a tooltip on the primary button **and** both dropdown variants — one edit, all sites. **Local actions (commit, new branch) stay enabled offline.**
- `usePrActions` (PR lifecycle: merge / close / convert draft / ready) and `useInboxCloudTaskRunner` (inbox Create PR / Discuss, scout Discuss) get an imperative offline guard that shows the offline toast instead of firing a doomed request — the same pattern `SessionView`'s send already uses.

### 3. Auto-recovery on reconnect
`NetworkReconnectContribution` subscribes to the connectivity store and, on an **offline→online** transition, calls a new `SessionService.recoverAfterReconnect()` (retry errored cloud streams + flush stranded cloud message queues). This is the same recovery the **window-focus** and **auth-restored** paths already perform — now also driven by the reconnect event, which previously fired neither. Local sessions need no new wiring: `reconcileLocalConnection` already re-runs when `isOnline` flips.

## Design choices
- **Disable specific actions, not a global modal.** A user can't fix wifi from inside a scrim, and reading a task / reviewing a diff / scrolling code all work offline. Trapping the whole app would be hostile; gating the unavailable actions (with a reason) and self-healing is what "graceful" means here.
- **No new store/transport plumbing.** Reuses `connectivityStore`, `useConnectivity`, the existing `checkNow` route, and the existing cloud-recovery service methods.

## Tests
- `gitInteractionLogic.test.ts`: offline gates push & create-PR; commit & new-branch stay enabled.
- `network-reconnect.contribution.test.ts`: recovers only on offline→online, once per cycle, never on online→offline or redundant updates.
- core + ui + apps/code typecheck clean; biome clean; touched core/ui suites green; host-boundary check shows no new violations.


tatoalo (approved):
looks good, approving to unblock ya

PR #2934, opened 2026-06-25
feat(notifications): focus-aware notification bus on quill toasts

adamleithp:
## What

A single **notification bus** that every app notification routes through, delivering by focus + what you're currently viewing:

| You are… | Result |
| --- | --- |
| viewing the exact target (task/canvas) | **suppress** (you can already see it) |
| app focused, but elsewhere | **in-app toast** with a click-through |
| app unfocused (another OS app) | **native OS notification** — click deep-links to the target |

image

Screenshot 2026-06-25 at 13 24 20

image

## How

- **`NotificationBus`** (renderer) replaces `TaskNotificationService`. A pure, table-tested `routeNotification()` decides the tier from a generic `NotificationTarget` union (`task | canvas`), defined in `@posthog/platform` (so core/host-router/ui share it without an internal-import violation).
- **Native click routing** generalized via a new `OpenTargetLinkService` + deep-link router `onOpenTarget`/`getPendingOpenTarget`. The renderer consumer drains pending **every (re)subscribe**, so a click landing during a reconnect/HMR gap still routes (was the "OS deep-link doesn't route" bug).
- **Producers migrated**: task prompt-complete + permission requests now flow through the bus (focused-elsewhere becomes a toast, not a native notification); canvas-generation-done re-emits through it (detection kept).
- **Toast primitive rebuilt on `@posthog/quill`** — provider-managed dismiss, stacking, and auto-dismiss (fixes the hand-rolled sonner-custom toast that couldn't be dismissed). All ~18 direct-sonner callers migrated; **`sonner` dropped** entirely.
- **New Settings → Notifications section**: defaults, an OS-permission banner, reset-to-defaults, and a per-tier **test harness** (in-app toast / deep-link toast / native / deep-link notification / dock badge / bounce). Canvas copy gated on the bluebird flag.

## Behavior change worth calling out

Task notifications that previously fired a native OS notification while the app was *focused-but-elsewhere* now show an in-app toast; native is reserved for when the app is backgrounded.

## Testing

- Pure `routeNotification` / `targetsEqual` table tests; zod↔platform `NotificationTarget` parity test; `useOpenTargetDeepLink` consumer test (warm task + canvas targets + pending-drain); `NotificationBus` tier-routing tests; core `NotificationService` target-click tests.
- Full workspace typecheck green; UI/core/host-router suites green; biome clean; host-boundary no new violations.


packages/ui/src/primitives/toast.tsx
```diff
@@ -1,175 +1,69 @@
...
+    title,
+    description: o.description,
+    timeout,
+    action: o.action,
+  };
+  if (o.id) {
+    quillToast.update(o.id, { type: level, ...fields });
+    return o.id;
+  }
+  return quillToast[level](fields);
```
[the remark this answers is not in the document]
adamleithp: Good catch — confirmed real, fixed in 32bef75e. `useArchiveTask` (`archive-undo-${taskId}`) and `auth.contribution` (`stranded-cloud-queue`) both create with a fresh stable id, so my `id`→`update` mapping no-op'd them. Reworked the wrapper to true **upsert** semantics via an id registry (caller id → quill id): first call creates + records the mapping, repeat updates the same toast, `dismiss(id)` resolves through it, entries self-clean on `onClose`. Added `toast.test.ts` covering create-with-id, update-on-repeat, dismiss, and id-frees-on-close.

packages/ui/src/features/settings/sections/NotificationsSettings.tsx
```diff
@@ -0,0 +1,383 @@
...
+          Notifications.
+        </Text>
+      )}
+
+      <Flex align="center" justify="between" className="mb-2 pt-2">
+        <Text className="font-medium text-sm">Defaults</Text>
+        <Button variant="soft" size="1" onClick={resetToDefaults}>
+          Reset to defaults
+        </Button>
+      </Flex>
```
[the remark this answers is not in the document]
adamleithp: Fixed in 32bef75e — generalized the copy (banner + permission toast) to OS-neutral wording ("Enable notifications for PostHog Code in your system settings"). This app does ship Linux/Windows makers, so the macOS-only path was genuinely wrong there.

packages/core/src/links/open-target-link.ts
```diff
@@ -0,0 +1,63 @@
...
+      this.log.info("Queueing open-target (renderer not ready)", {
+        kind: target.kind,
+      });
+      this.pending = target;
+    }
+
+    if (this.mainWindow.isMinimized()) {
+      this.mainWindow.restore();
+    }
+    this.mainWindow.focus();
```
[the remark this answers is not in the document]
adamleithp: Acknowledged — intentional and, as you note, consistent with the existing `TaskLinkService` single-pending design. The renderer subscription registers within ms of mount, so the only window is two distinct native clicks during cold start / a reconnect gap, which doesn't happen in practice (notifications are clicked one at a time). Promoting to a queue would mean changing `getPendingOpenTarget` to return an array across the tRPC boundary; not worth it for this edge. Leaving as-is.

packages/ui/src/features/notifications/routeNotification.ts
```diff
@@ -0,0 +1,37 @@
+import type { NotificationTarget } from "@posthog/platform/notifications";
+
+// Whether two targets point at the same thing (same kind + ids).
+export function targetsEqual(
```
adboio, line 15: [still reading code, maybe this is stupid, will update if so lol]

can/should we make this a bit more generic? like just give NotificationTarget some uniqueIdentifier property so we can just return `a.kind === b.kind && a.uniqueId === b.uniqueId` ? otherwise we'll have to update this if/when new target types are introduced

adboio: edit: just looked at the `NotificationTarget` type, both `task` and `canvas` are effectively the same thing (seems like identifier + "sub-identifier" (? not familiar w/ canvas)

in that case, could we make it a bit more generic and maybe not even split out the `kind` on the target?

packages/ui/src/features/notifications/notifications.ts
```diff
@@ -10,11 +16,36 @@ import {
...
+    duration?: number;
+  };
+  silent?: boolean;
+}
+
+// The single channel every app notification flows through. Reads focus + the
+// active route, decides suppress / toast / native (see routeNotification), and
+// dispatches accordingly. Native delivery + dock effects are gated by the user's
+// notification settings; the in-app toast always shows (it's non-intrusive and
+// only appears while the app is focused).
```
adboio, line 43: [nit] not just here specifically, but lots of bgi comments throughout, thoughts on stripping them down a bit and shipping this with a markdown file that explains the notif system instead ? no strong opinion here just personal pref

adboio (approved):
not tested too extensively but seems legit!! left a few comments, nothing blocking but lmk what you think :)

packages/ui/src/features/notifications/routeNotification.ts
adamleithp: Good call on the maintainability point — done in 64df5655a. Collapsed the per-kind equality chain into a single `targetKey(target)` identity function; `targetsEqual` is now just `targetKey(a) === targetKey(b)`, so adding a kind touches one exhaustive `switch` (compile error if you forget) instead of growing an if-chain that silently returns `false`.

On dropping `kind` / a flat `uniqueId`: I kept `kind` because it's load-bearing for *navigation*, not just equality — `deriveAction` in `notifications.ts` switches on it to pick `navigateToTaskDetail(taskId)` vs `navigateToChannelDashboard(channelId, dashboardId)`, which need different fields. And canvas identity is composite (`channelId` + `dashboardId`), so a single `uniqueId` would have to be a synthesized composite key anyway — which is exactly what `targetKey` now produces. So we get the generic comparison you wanted while keeping the typed per-kind fields navigation relies on.

packages/ui/src/features/notifications/notifications.ts
adamleithp: Fair — left them as concise inline comments for now rather than a separate markdown file. My reasoning: the routing decision (suppress/toast/native) is the one genuinely non-obvious bit and it's right next to the code it explains, where it's hardest to let drift out of sync. A standalone doc tends to rot once the code moves. Happy to pull it into a `notifications/README.md` if the comment density bugs you on a re-read though — no attachment.

PR #2955, opened 2026-06-26
fix(git): stop background sync-status polls from hitting the network

dmarticus:
## Summary

`useGitQueries` runs `getGitSyncStatus` on a 60s `refetchInterval` for every open repo/worktree. The server handler always called `git fetch` before returning (throttled to one fetch per 30s per directory). Over an SSH remote that means a `git fetch` → `ssh` → ssh-agent signing request roughly every minute per worktree, even when the user is idle. With Secretive (or any hardware-backed SSH agent), each signing request triggers a Touch ID / password prompt; concurrent worktrees produce overlapping prompts that cancel each other:

```
SecretAgent: sign a request from "PostHog Code" using secret "github"
SecretAgent: ... LACErrorCancelledByHigherPriorityAuthenticationKey ... "Canceled by another authentication."
```

This PR renames the existing `forceRefresh` input on `getGitSyncStatus` to `fetchFromRemote` and flips its semantics: the procedure no longer runs `git fetch` by default. Callsites that genuinely need an up-to-date view of `origin/*` opt in:

- `GitHost.getPrStateSnapshot` (PR-create flow) — `packages/core/src/git/git-host.ts:132`
- the agent service's explicit `fetchIfStale` adapter — `apps/code/src/main/di/container.ts:504`

Everything else — the UI sync-status poll, the post-mutation `getStateSnapshot`, `TaskPrStatusService` revalidations, the PR-create saga's `hasRemote` check — reads local refs only. They were either already reading purely local fields (`hasRemote`, `aheadOfDefault`, `currentBranch`, `isFeatureBranch`) or were happy with the local-only view of `aheadOfRemote`/`behind` that existing cache invalidations from commit/push/pull/branch-switch already keep fresh.

### Trade-off

Background "ahead of origin / behind origin" badges no longer self-update when an outside party pushes to upstream. They refresh whenever the user does a local action (commit, push, pull, branch switch — all already trigger invalidations) or opens the PR-create flow. The user-visible effect is missing teammate-side drift between idle interactions, which feels like an acceptable price for not hammering ssh-agent.

## Test plan

- [x] Added integration test in `packages/workspace-server/src/services/git/git.integration.test.ts`: another clone pushes to a shared bare remote; `getGitSyncStatus(work)` reports `behind: 0` by default, `getGitSyncStatus(work, true)` reports `behind: 1`.
- [x] `pnpm lint` clean.
- [x] Touched packages typecheck cleanly (pre-existing `claude-cli-sessions` / `taskCreationSaga` failures on `main` are unrelated).
- [x] Manual: open the app with several worktrees, leave idle for a few minutes, confirm ssh-agent / Secretive prompts stop.


packages/workspace-server/src/services/git/service.ts
```diff
@@ -445,12 +445,11 @@ export class GitService extends TypedEventEmitter<GitCloneEvents> {
...
   private async getGitSyncStatusInternal(
     directoryPath: string,
-    forceRefresh = false,
+    fetchFromRemote = false,
   ): Promise<GitSyncStatus> {
-    if (forceRefresh) {
-      this.lastFetchTime.delete(directoryPath);
+    if (fetchFromRemote) {
+      await this.fetchIfStale(directoryPath);
     }
```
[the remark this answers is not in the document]
dmarticus: Good catch — fixed in e96784093. The previous `forceRefresh: true` path cleared the throttle entry before calling `fetchIfStale`, so the fetch was guaranteed to run. My rename collapsed that into a plain `fetchIfStale` call, so the 30s throttle could swallow an opt-in fetch fired right after a push.

Extracted a `forceFetch` helper that always runs `git fetch` and stamps the clock; `fetchIfStale` delegates to it, and the `fetchFromRemote: true` path on `getGitSyncStatus` calls it directly. Strengthened the integration test with a back-to-back fetch assertion (well inside the throttle window) that catches the regression.

packages/workspace-server/src/services/git/service.ts
```diff
@@ -432,25 +432,35 @@ export class GitService extends TypedEventEmitter<GitCloneEvents> {
...
+   * exactly the moment it mattered.
+   */
+  private async forceFetch(directoryPath: string): Promise<void> {
+    try {
+      await gitFetch(directoryPath);
+      this.lastFetchTime.set(directoryPath, Date.now());
+    } catch {}
+  }
+
   private async fetchIfStale(directoryPath: string): Promise<void> {
```
jonathanlab, line 454: this now always force fetches right, e.g. still calling it `fetchIfStale` is a lie

packages/workspace-server/src/services/git/service.ts
```diff
@@ -432,25 +432,35 @@ export class GitService extends TypedEventEmitter<GitCloneEvents> {
...
+   * caller has explicitly asked for a fresh view of the remote (e.g.,
+   * `fetchFromRemote: true`) — otherwise a fetch triggered by a preceding
+   * mutation can silently swallow this one and leave the snapshot stale at
+   * exactly the moment it mattered.
+   */
+  private async forceFetch(directoryPath: string): Promise<void> {
+    try {
+      await gitFetch(directoryPath);
+      this.lastFetchTime.set(directoryPath, Date.now());
+    } catch {}
```
jonathanlab: Why are we silently catching here?

jonathanlab: I just saw that the old code did the same lol, but let's make this catch non-silent while we're at it

jonathanlab (approved):
LGTM, left some nits, thank u!

packages/workspace-server/src/services/git/service.ts
dmarticus: Good nit — the private `fetchIfStale` inside `GitService` still respects the throttle, so the name's accurate there. The actually-lying one was the agent adapter exposed as `GitDiffSource.fetchIfStale` (in `apps/code/src/main/di/container.ts`) — it goes through `getGitSyncStatus.query({ fetchFromRemote: true })`, which since the throttle-bypass fix unconditionally hits `forceFetch`. Renamed that interface method (and its caller in `git-pr.ts`) to `fetchFromRemote` in f4c40b935 so the public contract matches the behavior.

packages/workspace-server/src/services/git/service.ts
dmarticus: Done in f4c40b935 — `forceFetch` now writes the error to `process.stderr` before falling back to local refs (matches the existing pattern in `packages/git/src/signed-commit.ts` and `workspace-server/src/serve.ts`). I tried injecting `ROOT_LOGGER` first, but `@inject(...)` parameter decorators on `GitService` blow up workspace-client's typecheck — its tsconfig has no `experimentalDecorators` and it transitively imports `GitService`'s type via the AppRouter — so I went with the stderr write to avoid a DI restructure for a nit.

PR #2956, opened 2026-06-27
fix(command-center): prevent scroll on grid cell focus

K3N4Y:
## Problem

When clicking a command center grid cell, the browser scrolls the page to bring the newly-focused element into view. This is the default behavior of `.focus()` — it scrolls the focused element into the visible viewport, which jumps the user away from where they were.

## Changes

Added `{ preventScroll: true }` to the `.focus()` call in `CommandCenterGrid.tsx:84-85`. This tells the browser to focus the element without scrolling, matching the intent of the click handler (which delegates focus to an interactive element inside the cell without moving the viewport).

## How did you test this?

- `tsc --noEmit` clean
- `biome check` clean
- `vitest run` — 116 files, 1018 tests (UI), 179 files, 1755 tests (core), all passing
- Host boundary check clean

## Automatic notifications

- [ ] Publish to changelog?


charlesvien:
Showing 0 changed files
CleanShot 2026-07-01 at 18 51 32@2x

PR #2957, opened 2026-06-27
perf(renderer): faster big-chat open + bounded memory (#2162)

KarloAldrete:
## Problem

Issue #2162 — renderer CPU & memory on real, heavy sessions. Two concrete symptoms:

1. **Opening a big chat freezes the UI.** A 48k-event transcript is parsed, committed and rendered *entirely and synchronously* on open — ~1.1 s of main-thread jank before the first paint.
2. **A chat's memory is never freed.** `session.events` is append-only and stays resident after you navigate away. With a few big chats open (Command Center / "ADHD mode") the renderer heap climbs toward the `"memory-eviction"` crash reason the app already ships in `apps/code/src/main/index.ts`.

## Changes

Three focused changes on the open-time and memory path:

1. **Tail-first parse** — parse the tail of the ndjson first so the latest turns paint immediately, instead of `JSON.parse`-ing the whole ~110 MB log up front. The rest is parsed in non-blocking chunks (`parseSessionLogContentChunked`, yields between slices) and swapped in.
2. **Pre-freeze events** — `Object.freeze` each event at creation so immer skips its deep-freeze walk on commit (~240 ms → ~5 ms on a 48k-event session).
3. **Evict inactive transcripts** — free `session.events` ~20 s after a chat loses focus; rehydrate from disk (OS page cache) on return.

### Scope note

An earlier version of this branch also bundled (a) a windowed open with on-demand scroll-up history and (b) the sidebar / Command-Center re-render fixes. Both are now **out of this PR**:

- The **scroll-up jank** that the windowed open targeted is already fixed on `main` by **#2960** (Richard's `OVERSCAN` bump) — so the extra complexity of windowed scrollback isn't needed here.
- The **sidebar re-render** fix ships separately as **#2710**; the Command-Center variant can follow on its own.

This keeps the PR to the open-time + memory story, which is what the title promises.

### Benchmark (to be re-run on this reduced scope)

> ⚠️ The numbers below were measured on the earlier bundle. The open-time and evict-on-blur figures are driven by the changes that remain (tail-first parse, pre-freeze, eviction), so they should hold — but a fresh production A/B on exactly this scope is being re-run before these are treated as final. Note the steady-state RAM of a *focused* chat is **no longer** reduced (that came from the windowed open, now dropped).

Two production (minified) builds — `baseline` = `main`, `this PR` — same machine and user profile, opening the same real 48k-event chat. Renderer heap = live objects after forced GC; CPU summed across all Electron processes.

**Opening the chat (cold open) — tail-first parse + pre-freeze:**

| | baseline | this PR |
|---|--:|--:|
| Worst main-thread freeze | 1076 ms | **198 ms** (−82%) |
| Total main-thread busy | 2358 ms | **702 ms** (−70%) |
| CPU, all processes | 8.7 s | **1.6 s** (−82%) |

**Renderer heap when a chat loses focus — eviction:**

| | baseline | this PR |
|---|--:|--:|
| After navigating away | 196 MB | **53 MB** (−73%) |
| 4 big chats ("ADHD mode"), settled | 713 MB | **61 MB** (−91%) |

No regression in idle CPU, DOM node count, or GPU/VRAM.

## How did you test this?

- **Unit tests added with the changes:** `ensureEventsLoaded.test.ts`, `sessionLogs.chunked.test.ts`, `sessionStore.test.ts`, `useSessionEventsResidency.test.tsx`.
- `pnpm --filter @posthog/core test` → **1806 pass**; `pnpm --filter @posthog/ui test` → **1074 pass**; `pnpm typecheck` clean across all packages.
- **Production A/B benchmark** measured via Chrome DevTools Protocol + `/proc` + `/sys`, forced GC before each heap read (being re-run for the reduced scope — see note above).
- **Manual:** opened a 48k-event and a 96 MB chat; switching tasks evicts then rehydrates the transcript; streaming output unchanged.

## Automatic notifications

- [ ] Publish to changelog?

---


KarloAldrete:
Thanks for the review — addressed all three findings:

- **[P1] Scrollback gate could lock up** (`VirtualizedList.tsx`): the re-trigger gate was cleared only inside `if (delta > 0)`, so a load whose height the virtualizer hadn't measured yet — or one that prepended nothing renderable — left it stuck and blocked further scroll-up. The gate now releases whenever the anchor effect fires after a load, and `takeOlderEntries` skips older chunks that parse to zero entries so it never reports `hasOlder` alongside an empty slice. (007c8600)
- **Unused chunked parser**: removed `parseSessionLogContentChunked` and its only helper `parseLogLine` plus the test — dead since the windowed open parses the tail synchronously and loads older chunks on demand. (6edfb0ad)
- **Test style**: collapsed the repeated "changes when X changes" cases into an `it.each` table, matching `computeSidebarSessionSignature.test.ts`. (f027eea8)

Core (229) + UI sessions/command-center tests green, typecheck + biome clean.

KarloAldrete:
Superseded by #3063, which landed the pre-freeze + eviction work (and explicitly absorbs this PR and #2710). Closing to keep the queue clean.

PR #2964, opened 2026-06-27
Speed up channel/canvas/task loading on app reload

raquelmsmith:
## Problem

Reloading the Channels (project-bluebird) space felt slow. On every reload the
channel sidebar, a channel's canvases, and its filed tasks all blanked out and
refetched from scratch — the in-memory query cache is destroyed on reload and
nothing is shown until the network resolves. On top of that, each per-channel
list made a redundant extra round-trip to resolve the channel's folder path.

**Why:** make the Channels space feel instant on reload instead of waiting on a
cold cache + chatty round-trips. (Backend/Redis wouldn't help here — the cost is
the cold cache and the round-trip waterfall from the desktop app, not backend
compute, so the fix is client-side.)

## Changes

- **Persist the canvas query cache across reloads** (stale-while-revalidate). A
  shared `PersistQueryClientProvider` (desktop: async electron-store persister;
  web: `localStorage`) restores channels/dashboards/channel-tasks from disk
  instantly, then refetches in the background. Persistence is scoped to canvas
  queries by key, `gcTime` is raised to match the persist `maxAge`, and the
  on-disk blob is wiped on logout/project-switch (via `clearAuthScopedQueries`)
  so project-scoped data never leaks across sessions.
- **Drop a redundant per-channel round-trip.** `dashboards.list` /
  `channelTasks.list` resolved the channel's path with an extra `getEntry`
  before listing, even though the client already has it from `useChannels`.
  Thread the known `channelPath` through hook → router → service and skip the
  resolve (also halves `useTaskChannelMap`'s fan-out from N×2 to N).
- **Tune staleness.** `useChannels` now matches its 30s poll, and the web
  `QueryClient` gets the same defaults as desktop.

## How did you test this?

- Unit tests (ran locally, passing): service `list(channelId, channelPath)` skips
  `getEntry` when a path is supplied and falls back when it isn't
  (`dashboardsService.test.ts`, new `channelTasksService.test.ts`); the persist
  allowlist predicate (`queryPersistence.test.ts`).
- `typecheck` clean on the changed packages/apps (the only remaining error is a
  pre-existing `sonner` import in `FolderPicker.tsx`, unrelated to this change).
- `biome check` clean on all changed files; host-boundary check reports no new
  violations.
- Not yet manually verified in the running app — please pull and reload the
  Channels space to confirm the instant-from-cache paint.

## Automatic notifications

- [ ] Publish to changelog?

---


charlesvien:

PR #2986, opened 2026-06-29
fix(scouts): gate "Make a scout" on a personal GitHub repo

rubychilds:
## Problem

Clicking **Make a scout** on the agents page (`/code/agents`) showed a toast titled **"Failed to start scout authoring"** with the description *"Connect a GitHub repository before starting a scout authoring"*.

The root cause is an **integration-type mismatch**:

- "Make a scout" creates an auto-mode **cloud task** that runs under the user's *personal* GitHub identity, so it resolves its repo from **user-level** GitHub integrations (`useUserRepositoryIntegration`).
- The agents page's "Connected" status and Responders gating read **team/org-level** integrations.

A user with a team integration but **no personal GitHub installation** saw GitHub as "Connected", the button always enabled, and clicking it silently bailed in `useInboxCloudTaskRunner` because the resolved `cloudRepository` was `null`. The only place to add a personal installation is **Settings → GitHub**, which the agents page never pointed to.

## Fix

In `ScoutsFleetSection.tsx`, resolve the user-level repo and gate **only** the "Make a scout" chip on it:

- When no personal repo is available, the chip is replaced by an inline notice — *"Connect a personal GitHub repository to make a scout."* — with an **"Open GitHub settings"** button that deep-links to `/settings/github`.
- The two read-only question chips ("How is my scout fleet performing?", "What signals were emitted recently?") stay visible.
- While repos are still resolving, the chip is kept (no flash of the notice).

Screenshot 2026-06-29 at 11 07 48 AM

## Remaining
This PR prompts the user to connect their Github, but it doesn't remove the scout chips needing Github connection. for example. 'How is my scout fleet performing' etc. 

## Verification

- Reproduced and verified end-to-end in the running Electron app: with no personal install, the notice replaces the chip and "Open GitHub settings" lands on `/settings/github`.
- `pnpm --filter @posthog/ui typecheck` ✓ and Biome ✓.

---

### DIFF
--- a/packages/ui/src/features/scouts/components/ScoutsFleetSection.tsx
+++ b/packages/ui/src/features/scouts/components/ScoutsFleetSection.tsx
@@ -2,6 +2,7 @@ import type { Icon } from "@phosphor-icons/react";
 import {
   CaretDownIcon,
   CompassIcon,
+  GithubLogoIcon,
   PlusIcon,
   SparkleIcon,
 } from "@phosphor-icons/react";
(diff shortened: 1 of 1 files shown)


andrewm4894 (approved):

andrewm4894 (changes requested):
ty! i missed this - actually think i have a fix that should not need github

andrewm4894:
Superseded by #2988, which fixes the root cause: scout chats now run as **repo-less cloud tasks** rather than requiring a personal GitHub repo at all.

The GitHub requirement was purely a client-side gate in `useInboxCloudTaskRunner` — scout authoring is pure PostHog-MCP work and the backend already supports repo-less cloud tasks (nullable `Task.repository`; sandbox provisioning explicitly handles "Creating environment without repository"). #2988 adds an opt-in `allowMissingRepository` so a missing repo isn't an error, which also fixes the two read-only fleet question chips this PR left gated.

Suggest closing this in favour of #2988.

PR #2988, opened 2026-06-29
fix(scouts): run scout chats repo-less instead of gating on a GitHub repo

andrewm4894:
## Problem

"Make a scout" — and the read-only fleet question chips ("How is my scout fleet performing?", "What signals were emitted recently?") — create auto-mode **cloud tasks** through `useInboxCloudTaskRunner`, which **hard-required a personal GitHub repo + user-level integration**:

```ts
// useInboxCloudTaskRunner.ts
if (!cloudRepository) { toast.error(copy.errorTitle, { description: copy.missingRepository }); return; }
```

A user with only a **team-level** GitHub integration (no personal install) saw the button enabled, then hit a confusing **"Failed to start scout authoring / Connect a GitHub repository"** toast when the runner bailed on a `null` repository. This is the bug #2986 set out to address.

## Root cause

Scout chats are **pure PostHog-MCP work** — the `SCOUT_AUTHOR_PROMPT` scans the project over MCP and writes a `SKILL.md` into the skills-store. They **never touch repo code**. The repo requirement was purely a **client-side artifact** of the shared cloud-task runner.

The backend already supports repo-less cloud tasks (verified in `PostHog/posthog`):
- `Task.repository` is `null=True, blank=True`; `_build_task` gates GitHub-integration validation entirely behind `if repository:`.
- Sandbox provisioning explicitly handles the no-repo case: `get_sandbox_for_repository.py` → `elif not has_repo: emit_agent_log(... "Creating environment without repository")`, and `execute_sandbox/workflow.py` guards clone/checkout on `bool(prepared.repository ...)`. No repo → bare sandbox, no clone, no GitHub token. The agent still gets `POSTHOG_PERSONAL_API_KEY` + an OAuth token, which is all a scout-authoring task needs.

## Fix

Add an opt-in `allowMissingRepository` to `useInboxCloudTaskRunner`: when set, a missing repo is **not** an error and the task is created repo-less (no `repository`, no `githubUserIntegrationId`). Scout chats opt in; **Create-PR and Discuss keep the existing repo + integration gate**. When a personal repo *is* resolvable it's still passed through (harmless, keeps PR-authorship capability).

This fixes all three scout chips at once and removes the GitHub coupling entirely, rather than papering over it.

## Supersedes #2986

#2986 only addressed the symptom — it swapped the "Make a scout" chip for a "Connect a personal GitHub repository" notice (and left the two read-only chips still requiring GitHub). This PR removes the requirement at the source, so the connect-notice is no longer needed. **#2986 can be closed in favour of this.**

## Verification

- `pnpm --filter @posthog/ui typecheck` ✓ (also via pre-commit hook)
- Biome ✓
- New `useScoutChatTask.test.tsx`: covers the repo-less path (no personal repo → task still created with `repository: null`, no integration lookup) and the resolved-repo passthrough. Full `@posthog/ui` suite green (1024 tests).

---

### DIFF
--- a/packages/ui/src/features/inbox/hooks/useCreatePrReport.ts
+++ b/packages/ui/src/features/inbox/hooks/useCreatePrReport.ts
@@ -71,7 +71,9 @@ export function useCreatePrReport({
         reportUrl,
         feedback: feedbackRef.current,
       });
-      const targetRepo = ctx.cloudRepository.toLowerCase();
+      // Create-PR never runs repo-less, so the repo is always present here; the
+      // coalesce only satisfies the now-nullable context type.
+      const targetRepo = (ctx.cloudRepository ?? "").toLowerCase();
       const baseBranch = baseBranchOverrides
         ? (Object.entries(baseBranchOverrides).find(
             ([repo]) => repo.toLowerCase() === targetRepo,
@@ -81,7 +83,7 @@ export function useCreatePrReport({
         content: prompt,
         taskDescription: prompt,
         repository: ctx.cloudRepository,
-        githubUserIntegrationId: ctx.githubUserIntegrationId,
+        githubUserIntegrationId: ctx.githubUserIntegrationId ?? undefined,
         workspaceMode: "cloud",
         executionMode: "auto",
         adapter: ctx.adapter,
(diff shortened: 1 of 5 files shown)


sortafreel (approved):

skoob13 (approved):
Stamping this, but is there value of running scouts without a repo? Seems like we'll not deliver value.

andrewm4894:
> Stamping this, but is there value of running scouts without a repo? Seems like we'll not deliver value.

@skoob13 yeah so this would let you make a scout to do more non code stuff like eg the mcp feedback custom events etc - scout can still be useful even if dont have a github for them 

more so for stuff like non code tasks - eg account churn risk scouts or watch my dashboard scouts etc - lots of non pr based scouts use cases comming up internally and externally so want to try support them

PR #3000, opened 2026-06-29
fix(code-review): live diff line highlight + clean comment box background

MattPua:
## Problem

When dragging to select lines in a PR diff, two things were off:
- The selected lines didn't highlight until **after** releasing the mouse.
- The inline comment box rendered with a **blue** background (the selection highlight bleeding into it).

## Changes

Before:

https://github.com/user-attachments/assets/25cc2881-27d1-49fb-af83-5ed713d349f1

After

https://github.com/user-attachments/assets/7af57e8e-75ec-4d84-a528-37d4a4dfed53

- The diff was always in `@pierre/diffs` "controlled selection" mode (we passed `selectedLines={null}`, never `undefined`), so the highlight only painted when the prop changed — on selection *end*. Wired `onLineSelectionChange` to update the range live during the drag.
- Gave the comment box container an opaque `bg-(--diffs-bg)` so the selection highlight stops at the lines instead of tinting the box.

## How did you test this?

- Manually reloaded the app: lines now highlight live while dragging, and the comment box renders on a clean background.
- `pnpm --filter @posthog/core test diffAnnotations` (5 passing, updated for the new signature).
- Typecheck clean on all touched files.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/code-review/diffAnnotations.test.ts
+++ b/packages/core/src/code-review/diffAnnotations.test.ts
@@ -87,18 +87,21 @@ describe("buildCommentMergedOptions", () => {
     { hasOpenComment: false, expectedEnabled: true },
     { hasOpenComment: true, expectedEnabled: false },
   ])(
-    "with hasOpenComment=$hasOpenComment sets selection/gutter enabled=$expectedEnabled and routes both callbacks to the handler",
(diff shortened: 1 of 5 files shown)


jonathanlab (commented):
-

jonathanlab (approved):
🐐 , great stuff

PR #3014, opened 2026-06-30
chore(prompts): allow slack thread links in PR copy

VojtechBartos:
## Problem

The public-repo safety rule in the PR-writing prompts forbids any reference to Slack threads, but Slack thread links sit behind PostHog auth and are useful as origin context. Only the *contents* of a thread are private.

## Changes

Updated both PR-writing prompts to:

- Still forbid quoting or paraphrasing what was said in a Slack thread.
- Explicitly allow linking to the originating thread and channel references like "raised in #team-foo".

Touched:

- `packages/agent/src/server/agent-server.ts` — `publicRepoSafetyInstruction`
- `packages/core/src/git-pr/git-pr.ts` — body rules for `generatePrTitleAndBody`

## How did you test this?

Prompt-only change. No existing test snapshots the safety string, so nothing to update; manual diff review only.


Twixes (approved):
This makes sense, just why do we need to make this change both in https://github.com/PostHog/posthog/pull/66936 and here? That's a little fishy

VojtechBartos:
@Twixes we don't, just making sure it's updated across repos and behaving same way

PR #3015, opened 2026-06-30
fix(agent): require local commit ownership before attributing a PR to a run

VojtechBartos:
## Problem

The PR-attribution path in the agent server stamps `output.pr_url` onto the current TaskRun whenever a recently-created PR URL appears in any `session/update` text — regardless of whether *this* run actually opened the PR. Any sandbox that reads a PR URL from a tool response (e.g. a `signals/reports` listing whose `implementation_pr_url` field references a sibling run's PR, or a GitHub query that surfaces another bot's recent PR) gets attributed to that PR.

The downstream effects are nontrivial. `output.pr_url` is the field that:

- drives the Temporal babysitting orchestrator's CI follow-up loop (the agent gets re-entered every ~15 min with *"address CI feedback on the pull request you opened"*), and
- triggers a `_post_slack_update_for_pr` call that posts a "Pull request opened" card into whatever Slack thread is mapped to the run.

So a single misattribution can both surface a stranger's PR into the wrong Slack thread *and* keep the wrong agent session alive for hours debugging a PR it had no part in opening.

## Root cause

`maybeAttachCreatedPr` / `attachPrIfCreatedThisRun`:

```ts
const prUrl = findPrUrl(JSON.stringify(update));     // any URL in any text
…
if (!wasCreatedRecently(createdAt, Date.now())) return;   // 15-min window — the only gate
await posthogAPI.updateTaskRun(taskId, runId, { output: { pr_url: prUrl } });
```

Pre-#2765 the detector required the URL to come from a `Bash` tool's `gh pr create` stdout — a real "this run produced the PR" signal. #2765 traded that for a 15-minute time window, which fires whenever any sibling run's PR happens to be recent. That latent bug became frequent once babysitting was enabled for all `signal_report`-origin runs (lots of bot-authored PRs concurrently in the window), and started showing up as cross-thread "Pull request opened" surprises.

## What this changes

Reintroduce an ownership check that can't be spoofed by reading text in a tool response.

- `fetchPrCreatedAt` → `fetchPrAttributionMetadata`: also returns the PR's `headRefName`, `headRefOid`, and `author.login`.
- Before stamping `output.pr_url`, require the PR's head SHA to resolve to a commit object in this sandbox's local git database via a new `commitExistsLocally` helper (`git cat-file -e <sha>^{commit}`). A run that didn't produce (or fetch) that commit cannot have created the PR — SHAs can't be regex-matched into existence.
- Keep the 15-minute recency window as a cheap pre-filter so the git check only runs for plausible candidates.
- Also persist `output.head_branch` alongside `output.pr_url` so the row's shape matches what the GitHub webhook attribution backstop already writes.

`commitExistsLocally` fails closed on every error path (missing repo path, malformed SHA, git unavailable, abort signal). The cost of dropping a real attribution on the floor is much lower than the cost of taking a wrong one — that's how this bug compounded into multi-hour babysitting loops.

## Test plan

- [x] `pnpm vitest run packages/agent/src/server/agent-server.test.ts` — 78/78 pass, including the new ownership cases:
  - rejects when `commitExistsLocally` returns false (the failure mode in production)
  - rejects when the PR returns no head SHA (defensive)
  - existing cases adapted to the new metadata shape and `head_branch` write
- [x] `pnpm vitest run packages/git/src/queries.test.ts` — 30/30 pass
- [x] `pnpm biome check --write` on the three changed files
- [x] `pnpm typecheck` clean on the changed packages (ran via lint-staged on commit)

## Notes for reviewers

- The new test mocks `commitExistsLocally` via `vi.mock("@posthog/git/queries", …)` with a `vi.hoisted` factory — necessary because the existing pattern of overriding instance methods doesn't reach a module-level imported function.
- I considered also gating on `author.login === gitAuthorLogin` (cheap, no git call) but it's redundant once the SHA check is in place and gives a false sense of security when multiple runs share `posthog[bot]`. SHA membership is the single load-bearing check.
- Defense-in-depth I'm intentionally *not* including here, in scope for a follow-up: server-side rejection in the PostHog API for `output.pr_url` writes whose `head_branch` doesn't match `task_run.branch`. That belongs in `posthog/posthog`'s facade, not here.

### DIFF
--- a/packages/agent/src/server/agent-server.test.ts
+++ b/packages/agent/src/server/agent-server.test.ts
@@ -207,6 +207,14 @@ vi.mock("@anthropic-ai/claude-agent-sdk", async (importOriginal) => ({
   query: mockedClaudeSdk.query,
 }));
 
+const mockedGitQueries = vi.hoisted(() => ({
+  commitExistsLocally: vi.fn(async () => true),
+}));
+vi.mock("@posthog/git/queries", async (importOriginal) => ({
+  ...(await importOriginal()),
+  commitExistsLocally: mockedGitQueries.commitExistsLocally,
+}));
+
 interface TestableServer {
   getInitialPromptOverride(run: TaskRun): string | null;
   getClearedPendingUserState(run: TaskRun | null): string[] | null;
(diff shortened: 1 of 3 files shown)


VojtechBartos:
Fixed just by shortening the window https://github.com/PostHog/code/pull/3019

PR #3026, opened 2026-06-30
feat(notifications): scale completion sound speed with task length

frankh:
## Summary

Adds an opt-in **"Scale sound speed with task length"** toggle to **Settings → Notifications**. When enabled, the completion sound's playback rate reflects how long the just-completed turn took:

- **≤ 30s** → ~3× faster (higher pitch)
- **2–4 min** → normal speed (the "normal" band)
- **≥ 30 min** → ~3× slower (lower pitch)
- smooth log-interpolation across the two ramps, clamped to `[1/3, 3]`

Off by default, so existing behavior is unchanged unless the user turns it on. It uses `HTMLAudioElement.playbackRate`, so faster also means higher-pitched and slower means lower-pitched.

## How it works

- New `playbackRateForTaskDuration(durationMs)` helper in `sounds.ts`; `playCompletionSound` gains an optional `playbackRate` param.
- The just-completed turn duration is computed in `sessionService` from `acpMsg.ts - session.promptStartedAt` (Immer keeps the pre-update snapshot, so the start time is still readable), plumbed through `notifyPromptComplete` → `NotificationBus.notify` → `playCompletionSound`.
- New `scaleSoundWithTaskLength` setting added to the settings store, the `NotificationSettings` contract, the desktop settings provider, and the Notifications settings UI (shown only when a sound is selected; included in "Reset to defaults").

## Tests

- `sounds.test.ts`: parametrized coverage of the duration→rate mapping across anchor/boundary points.
- `notifications.test.ts`: asserts the scaled rate is applied when the toggle is on with a duration, and `1` when off or when no duration is available.

---

### DIFF
--- a/apps/code/src/renderer/desktop-services.ts
+++ b/apps/code/src/renderer/desktop-services.ts
@@ -202,6 +202,7 @@ container
         dockBounceNotifications: s.dockBounceNotifications,
         completionSound: s.completionSound,
         completionVolume: s.completionVolume,
+        scaleSoundWithTaskLength: s.scaleSoundWithTaskLength,
         customSounds: s.customSounds,
       };
     },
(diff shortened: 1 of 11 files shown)


pauldambra (approved):
do more weird

PR #3029, opened 2026-06-30
fix(cloud-task): lower session-log history page request size

tatoalo:
## Problem

On bootstrap the cloud-task watcher loads run history by paging `session_logs`. With the page size at 5,000 entries, the first page of a long run can be many MB and fail to transfer within the watcher's per-request fetch timeout, so it retries indefinitely and the run is stuck "Connecting to your cloud runner" even though it is healthy.

## Changes

Lower `SESSION_LOG_PAGE_LIMIT` from 5,000 to 1,000 so the first (and every) history page request stays small


tatoalo:
closing, since BE gate is enough

PR #3005, opened 2026-06-29
feat(command-menu): show keyboard shortcuts in command menu

MattPua:
## Summary

- Adds right-aligned keyboard shortcut hints to command menu items
- Adds **Go back** / **Go forward** navigation actions
- Reorders sections: Actions before Navigation

> **Danger level: 1/5**

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

PR #3030, opened 2026-06-30
fix(os-service): use pathToFileURL for cross-platform log folder path

MattPua:
## Summary

Follow-up to #3005. Fixes `showLogFolder` to produce a valid file URI on all platforms.

- `file://${path}` breaks on Windows (`C:\...` → `file://C:\...`, invalid)
- `pathToFileURL(path).href` (Node built-in) handles this correctly on all platforms
- Adds a test asserting the launcher is called with a `file://` URL

> **Danger level: 1/5**

---


MattPua:
Merged into #3031.

PR #3041, opened 2026-07-01
fix(agent): encode non-ascii gateway property headers

### LINKED ISSUE #65950: Bug report: posthog code - non-ascii slack task titles break header
When a task is triggered from Slack with a prompt containing non-ASCII characters, the run fails immediately with `Internal error: API Error: Header 'x-posthog-property-task_title' has invalid value: '<the prompt text>'`. This was hit with Danish characters (`å`, `æ`, `ø`), e.g. the prompt "har vi nogle logs på bil-fabrikken her de sidste par dage?", but it applies to any non-ASCII title (accents, CJK, emoji, etc.). The task title is set verbatim as the `x-posthog-property-task_title` HTTP header so the title can be lifted onto the in-sandbox agent's `$ai_generation` traces, but HTTP header values are ASCII-only (RFC 7230), so the request is rejected before it is sent and the whole task fails. Expected: a task with a non-ASCII title runs normally (the trace tag can be encoded, sanitized, or dropped, but it should never fail the run).

The task title originates from the raw Slack message text in `posthog/temporal/ai/slack_app/activities/task_creation.py` and is stored unmodified (correct, the DB and UI want the real title). The failure is at the header-construction boundary in the PostHog Code agent-server (`configureEnvironment`, where `x-posthog-property-*` properties are attached), which lives in the separate PostHog Code monorepo, not in `PostHog/posthog`. Same mechanism as the `ai_stage` property forwarded from `products/tasks/backend/models.py`.

## How to reproduce

1. Connect the Slack integration to PostHog Code.
2. Send a prompt to the bot containing a non-ASCII character, e.g. `har vi nogle logs på bil-fabrikken her de sidste par dage?` (note the `å`).
3. The task fails with `Task Failed :x: Internal error: API Error: Header 'x-posthog-property-task_title' has invalid value: '...'`.

Workaround for affected users: write the prompt in plain ASCII (e.g. "paa" instead of "på") or in English.

## Additional context

From: https://posthoghelp.zendesk.com/agent/tickets/61102 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/63013)

PostHog AI conversation ID: 06590392-0647-4573-a0a9-490acda642cc
Trace ID: 13bb2cb2-e6ca-4115-b4bd-59e15879953f

## Debug info

Session replay: https://us.posthog.com/project/sTMFPsFhdP1Ssg/replay/019efbf1-eb46-706a-b19f-6cad75c7e0fd?t=0

EU Cloud, project ID 152609.

MarxWellB:
## Problem

Slack-created PostHog Code tasks can fail when the task title contains non-ASCII characters. The original task title is forwarded as an `x-posthog-property-task_title` HTTP header value, and some HTTP clients reject header values containing characters outside the safe ASCII range.

The original title should remain unchanged for storage and UI display. The unsafe boundary is only when the title is serialized into gateway property headers.

Fixes PostHog/posthog#65950.

## Changes

* Preserve the existing newline-collapsing behavior to prevent header injection.
* Keep printable ASCII characters unchanged.
* Percent-encode all non-ASCII characters with `encodeURIComponent`.
* Add regression coverage for smart quotes, emoji, accented characters, CJK characters, and the Danish-style input described in the issue.

## Why encode instead of strip?

Stripping non-ASCII characters makes the header safe, but loses user context. Encoding keeps the value ASCII-safe while preserving the original information in a reversible representation.

Examples:

```txt
å  -> %C3%A5
é  -> %C3%A9
東京 -> %E6%9D%B1%E4%BA%AC
🚀 -> %F0%9F%9A%80
```

## How did you test this?

* [x] `pnpm --filter @posthog/agent test gateway.test.ts`
* [ ] `pnpm --filter @posthog/agent typecheck` — fails locally due unresolved internal workspace packages such as `@posthog/shared`, `@posthog/git`, and `@posthog/enricher`; no errors are reported in `gateway.ts` or `gateway.test.ts`.

### DIFF
--- a/packages/agent/src/utils/gateway.test.ts
+++ b/packages/agent/src/utils/gateway.test.ts
@@ -124,15 +124,21 @@ describe("buildGatewayPropertyHeaders", () => {
     },
   );
 
-  it("strips characters an HTTP header value cannot carry", () => {
-    expect(buildGatewayPropertyHeaders({ task_title: "don’t🚀ship" })).toBe(
-      "x-posthog-property-task_title: dontship",
-    );
-  });
-
-  it("keeps latin1 characters such as accents", () => {
-    expect(buildGatewayPropertyHeaders({ task_title: "café" })).toBe(
-      "x-posthog-property-task_title: café",
+  it.each([
+    {
+      description: "smart quote and emoji",
+      title: "don’t🚀ship",
+      expected: "don%E2%80%99t%F0%9F%9A%80ship",
+    },
+    {
+      description: "Danish accented, CJK, and emoji characters",
+      title: "har vi nogle logs på café 東京 🚀",
+      expected:
+        "har vi nogle logs p%C3%A5 caf%C3%A9 %E6%9D%B1%E4%BA%AC %F0%9F%9A%80",
+    },
+  ])("encodes non-ASCII characters in $description", ({ title, expected }) => {
+    expect(buildGatewayPropertyHeaders({ task_title: title })).toBe(
+      `x-posthog-property-task_title: ${expected}`,
     );
   });
 });
--- a/packages/agent/src/utils/gateway.ts
+++ b/packages/agent/src/utils/gateway.ts
@@ -28,14 +28,14 @@ export function resolveGatewayProduct({
 }
 
 /**
- * Make a value safe to embed in an HTTP header value. Collapses newlines to
- * spaces (the header block is newline-delimited) and drops characters outside
- * the valid header-byte range — control chars and code points above latin1
- * (emoji, smart quotes) — which an HTTP client (e.g. undici) would otherwise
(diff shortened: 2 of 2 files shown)


charlesvien:
Thanks for digging into this. Closing as superseded: main fixed the same bug in #3375 by transliterating header values to ASCII in `packages/shared/src/posthog-property-headers.ts`, and `gateway.ts` is now just a re-export of that module, so this branch can't be rebased onto the new location.

Percent-encoding also has two problems here: nothing decodes the values downstream, so titles would show up in analytics as `caf%C3%A9`, and `encodeURIComponent` throws on lone surrogates, turning a malformed title into a crash during session setup.

If preserving CJK/emoji fidelity matters, that would need a change in the shared module plus a decode step on the gateway side.

PR #3048, opened 2026-07-01
feat(slack): send agent turn parts alongside joined text on /relay_message

VojtechBartos:
## Problem

The Slack agent's initial reply posted the whole assistant turn including the "Let me pull DAU…" narration before each tool call. Slack should see only the final answer.

## Changes

`SessionLogWriter` already collects each assistant text block into `currentTurnMessages` before joining them at read time. Expose that raw list via `getAgentResponseParts()` and pipe it to the backend as `text_parts` alongside the existing joined `text` field.

The backend (PostHog/posthog#67297) posts the last non-empty `text_parts` entry when present, so only the post-last-tool-use text reaches Slack. `text` stays as the fallback for backends that don't understand `text_parts` both sides ship in either order.

Only the initial-reply site (`agent-server.ts`) sends `text_parts`; the Slack-question relay path is untouched (single formatted question, no interim narration to trim).

## How did you test this?

Automated only (agent, no manual testing claimed):

- `getAgentResponseParts` unit tests: returns each turn message as a separate entry; returns `undefined` for empty/unregistered sessions.
- Existing session-log-writer and posthog-api tests still pass.
- Tested locally

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Model: Claude Opus 4.7 (1M context).

### DIFF
--- a/packages/agent/src/server/agent-server.ts
+++ b/packages/agent/src/server/agent-server.ts
@@ -3023,11 +3023,20 @@ ${signedCommitInstructions}
       return;
     }
 
+    // Ordered assistant text blocks (one per message between tool calls).
+    // The backend picks the last entry — the post-last-tool-use answer — so
+    // Slack no longer sees the "Let me check…" narration. `message` stays as
+    // the joined fallback for backends that don't understand `text_parts`.
+    const messageParts = this.session.logWriter.getAgentResponseParts(
+      payload.run_id,
+    );
+
     try {
       await this.posthogAPI.relayMessage(
         payload.task_id,
         payload.run_id,
         message,
+        messageParts,
       );
     } catch (error) {
       this.logger.debug("Failed to relay initial agent response to Slack", {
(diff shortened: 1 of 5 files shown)


packages/agent/src/server/agent-server.ts
```diff
@@ -3026,11 +3023,20 @@ ${signedCommitInstructions}
       return;
     }
 
+    // Ordered assistant text blocks (one per message between tool calls).
+    // The backend picks the last entry — the post-last-tool-use answer — so
+    // Slack no longer sees the "Let me check…" narration. `message` stays as
+    // the joined fallback for backends that don't understand `text_parts`.
+    const messageParts = this.session.logWriter.getAgentResponseParts(
+      payload.run_id,
+    );
```
tatoalo, line 3032: can we gate this to slack only? this would run across cloud runs right? 🤔 

if this relay is still gated all good

VojtechBartos: yeah, the relay endpoint is gated only for slack aka if no slack thread task mapping exists, we skip.

tatoalo (approved):

PR #3049, opened 2026-07-01
fix(archive): don't let a broken worktree block archiving

frankh:
## Problem

Some worktree tasks can't be archived — every attempt just shows the generic `Failed to archive task` toast, forever.

Archiving a worktree task first captures a git **checkpoint** of the worktree (a restore point used on un-archive). That capture (`CaptureCheckpointSaga`) refuses to run and throws when the worktree is mid-operation or conflicted:

- `Cannot capture checkpoint while git operation is in progress` (e.g. an interrupted rebase/merge)
- `Cannot capture checkpoint with unresolved merge conflicts in the index`

When capture throws, the **entire** archive aborts and rolls back (`ArchiveService.archiveTask`), so the task can never be archived — and because the worktree stays stuck in the bad state, it fails on every retry. The real error is only logged; the UI shows the generic toast.

Confirmed against real data: a worktree left in an interrupted rebase with unmerged index entries reproduces this exactly.

## Fix

Checkpoint capture is best-effort recovery data — a failure to snapshot should not block the destructive archive. The flow already degrades gracefully when a worktree isn't a valid git repo (logs a warning, sets `checkpointId = null`, proceeds). This extends that same tolerance:

1. **Checkpoint capture is non-fatal.** On failure, log a warning, drop the restore point (`checkpointId = null`), and continue teardown. Safe because the saga rolls back its own internal steps before failing, so no partial checkpoint ref is left behind.
2. **Worktree teardown is hardened.** A `deleteWorktree` failure (git lock, filesystem error) now logs a warning and still records the archive.

Trade-off: a task archived this way has no restore point on un-archive — acceptable, since the worktree was already in an unrestorable/broken state. No manual DB/git surgery is needed to recover currently-stuck tasks; retrying Archive after this ships succeeds.

## Testing

- New integration test: archiving a worktree left in an in-progress merge conflict now succeeds with `checkpointId: null` and the worktree removed (rejects without the fix).
- `pnpm --filter @posthog/workspace-server test` — 28/28 archive integration tests pass.
- `pnpm --filter @posthog/workspace-server typecheck` — clean.
- `biome lint` — clean.

---

### DIFF
--- a/packages/workspace-server/src/services/archive/archive.integration.test.ts
+++ b/packages/workspace-server/src/services/archive/archive.integration.test.ts
@@ -395,6 +395,41 @@ describe("ArchiveService integration", () => {
         expect(archived.branchName).toBeNull();
         expect(ctx.archiveRepo.findAll()).toHaveLength(1);
       }));
+
+    it("archive succeeds when worktree has an in-progress merge conflict", () =>
+      withTestContext({}, async (ctx) => {
+        const { worktreePath } = await ctx.setupWorktree("detached");
+
+        const wt = (cmd: string) =>
+          execSync(`git ${cmd}`, {
+            cwd: worktreePath,
+            encoding: "utf8",
+            stdio: "pipe",
+          });
+
+        await fs.writeFile(path.join(worktreePath, "c.txt"), "base\n");
+        wt("add c.txt");
+        wt("commit -m base");
+        wt("tag base_commit");
+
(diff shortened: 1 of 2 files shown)


frankh:
I think it makes more sense to just leave dirty worktrees lying around than being completely unable to archive the task

jonathanlab (approved):

PR #3060, opened 2026-07-01
feat(notifications): agent spoken narration via speak tool

jonmcwest:
## What

Gives the agent a voice. A new `speak` local tool lets the agent narrate out loud — when it **finishes** your request, reaches a **meaningful new phase**, or is **blocked and needs you** (a question, decision, confirmation, or an error only you can resolve). You hear it even when you're in another tab.  
  
Demo https://posthog.slack.com/archives/C0AP5NXF8D7/p1782920963735449

## How it fits the architecture

The flow follows the same "pure side-effect off the event stream" pattern as completion/permission notifications:

- **`speak`** **tool** (`packages/agent`) — no-op agent-side (the agent subprocess/sandbox can't reach your speakers). It just surfaces a `tool_call` carrying the text + kind.
- **`SpeechQueueService`** (`packages/core`) — portable orchestration. Serializes utterances so parallel sessions never talk over each other, coalesces stale per-task narration, lets `needs_input` lines jump the queue (never dropped), and caps the queue so a burst can't back up minutes of speech. Depends only on the host-neutral `ISpeech` capability + injected settings/name providers — no tRPC, Node, or Electron.
- **`composeUtterance`** (`packages/core`) — the app owns the `PostHog Code task '…' —` prefix and the `Hey <name>,` greeting for needs-user lines, so the agent supplies only content. Idempotent against an agent that already added a prefix/greeting.
- **`ISpeech`** **adapter** (`apps/code`) — ElevenLabs playback with a macOS `say` fallback (audio tags like `[laughs]` are stripped by the fallback).
- **Settings** — users choose which kinds are spoken (and can mute `progress`), pick a voice, and set the API key in Notifications settings.

## Testing

- `packages/core` speech + composeUtterance: 18 tests pass
- `packages/ui` speechRouting: 12 tests pass
- `packages/agent` speak tool: 11 tests pass
- Full `pnpm typecheck` clean across all 22 packages
- Biome lint clean on touched files (no `noRestrictedImports` violations in core)

---

_Created with_ [_PostHog Code_](https://posthog.com/code?ref=pr)


jonmcwest:
Addressed both Greptile P2 comments in dd25b52:

1. **Orphaned `speakCalls` on teardown** — `speakCalls` is now keyed by `taskRunId` (inner map keyed by tool-call id), so `unsubscribeFromChannel` drops any speak calls still mid-stream for a departing run instead of leaking them until `reset()`.

2. **Greeting injection before the double-prefix guard** — the `^posthog code task` guard now runs *before* any greeting logic in `composeUtterance`, so an already-prefixed agent line can no longer be double-prefixed when `addressByName` is true. Added a regression test covering that case.

Also fixed the two failing CI checks: added `enqueueSpeech` to the cloud-task-update notification test harness deps, and allowlisted the host-resident ElevenLabs speech service (the API key intentionally stays in the host process, matching `secure-store`/`encryption`).

pauldambra (commented):
i'd be happy to approve this since i've seen it run but it's big enough we should _probably_ get team code to review it  
  
i think it has the natural next step of hooking up speech to text so you can reply when ph-ode talks to you

jonmcwest:
Rebased onto latest \`main\` to clear the merge conflict (only conflict was an import-ordering clash in \`apps/code/src/renderer/desktop-services.ts\` between the new \`Adapter\` type import from main and the \`@posthog/platform/speech\` import from this branch — both kept). Verified the rebase added no base-branch files to the PR diff, and `pnpm build` + full `pnpm typecheck` pass locally.

CI is now green: all required checks (build, quality, unit-test, integration-test, typecheck) pass. The non-required live `e2e` job flaked twice on unrelated `session-lifecycle` live-model tests (once `interrupts an in-flight turn`, once a 240s timeout on `plan mode ... codex` — a different test each run, neither touched by this PR), then passed cleanly on rerun.

The PR is mergeable with no conflicts; the only remaining gate is `REVIEW_REQUIRED`. @pauldambra flagged wanting team code to review — flagging that it is rebased, green, and ready for that review.

packages/ui/src/features/settings/sections/NotificationsSettings.tsx
```diff
@@ -360,6 +366,8 @@ export function NotificationsSettings() {
         </SettingRow>
       )}
 
+      <SpokenNotificationsSection />
```
pauldambra: is this feature flagged?  
we should probably just release to us for now?

pauldambra (approved):
let's wrap the settings in a flag otherwise :ship::ship::ship:

packages/ui/src/features/settings/sections/NotificationsSettings.tsx
jonmcwest: Good call — it was not flagged. I have put the whole Spoken notifications section behind a new `posthog-code-spoken-narration` feature flag (always on in dev builds), matching how canvas/autoresearch are gated. The master toggle already defaults to off, and the `speak` tool only produces audio when the user enables it in settings — so hiding that settings entry point means non-flagged users can never turn it on. We can now roll it out to the US cohort first via the flag.

<sub>🦉 via [talyn.dev](https://talyn.dev)</sub>

PR #3072, opened 2026-07-01
feat(autoresearch): agent metric-optimization mode with dashboard (GROW-103)

fercgomes:
## Summary

Implements [GROW-103](https://linear.app/posthog-team-growth/issue/GROW-103/posthog-code-autoresearch): **autoresearch** as a PostHog Code mode, modeled on [pi-autoresearch](https://github.com/davebcn87/pi-autoresearch) — a long-running agent loop that optimizes a metric. Under the hood it is the task's regular LLM session; a dedicated dashboard panel shows the run's progress.

Screenshot 2026-07-02 at 18 50 04

Screenshot 2026-07-02 at 18 50 20

### How it works

- **Protocol** — the kickoff prompt teaches the agent to end every turn with a fenced ` ```autoresearch ` report block (`metric: <number>`, `summary: <one line>`). Each iteration = one focused change + one measurement.
- **Auto-continue engine** (`packages/core/src/autoresearch/`) — `AutoresearchService` subscribes to the sessions store, detects turn completion (`isPromptPending` true→false), parses the report from the transcript, records the iteration, and sends the next continuation prompt (with best-so-far and recent history) until the **target is reached** or the **iteration budget is spent**. A turn without a report gets one reminder; a second lapse fails the run. Session errors and send failures fail the run with a reason. Runs can be paused (iterations still recorded, no auto-continue), resumed, and stopped.
- **Dashboard** (`packages/ui/src/features/autoresearch/`) — a new "Autoresearch" panel tab in task detail: status/direction badges, pause/resume/stop controls, stat cards (best / last / iterations / target), an SVG chart of the metric with best-so-far frontier and target line, an iterations table with deltas, a run-history selector, and a start-configuration dialog (metric, direction, optional target, budget, instructions). A `ChartLineUp` button in the task header (with a live-run indicator) opens the panel.
- **Architecture** — business logic lives in `@posthog/core` behind DI; the service reaches the session through a narrow `AutoresearchPromptClient` seam bound in `desktop-services.ts` (same pattern as `AGENT_PROMPT_SENDER`). The UI degrades gracefully on hosts that don't bind the service (`useServiceOptional`), so web stays a portability smoke test.

### Screenshots

The dashboard renders from the domain store; start a run from any task via the header chart button → **Start autoresearch**.

## Test plan

- [x] 80 new unit tests (`packages/core/src/autoresearch/*.test.ts`): config schema validation, stats/decision functions, report parsing + transcript extraction, prompt builders, and a full service harness driving fake session turns (iteration recording, continuation, target/budget completion, reminders, pause/resume/stop, session errors, send failures, unrelated-task isolation, dispose).
- [x] Full suites pass: core 1944, ui 1120, shared 423, app 23 files.
- [x] `pnpm --filter @posthog/core|@posthog/ui|code typecheck` clean.
- [x] Biome clean; zero `noRestrictedImports` in core; `node scripts/check-host-boundaries.mjs` reports no new violations.
- [ ] Manual smoke test in the desktop app (start a run against a real task, watch the dashboard update).

Note: `@posthog/git` and two workspace-server integration suites fail in the cloud sandbox because real `git commit` is blocked there — pre-existing and unrelated to this change.


packages/core/src/autoresearch/prompts.ts
```diff
@@ -0,0 +1,268 @@
...
+}
+
+const REPORT_BLOCK_REGEX = /```autoresearch\s*\n([\s\S]*?)```/g;
+
+/**
+ * Parse the agent's metric report from a reply. The last well-formed
+ * ```autoresearch fenced block wins, so an agent quoting the protocol and
+ * then reporting still parses correctly.
+ */
+export function parseMetricReport(text: string): AutoresearchReport | null {
```
DanielVisca, line 175: the metric is entirely self-reported — parseMetricReport over the agents own reply is the only source, and the host seam (AutoresearchSessionClient) has no way to run a measurement itself, so a single hallucinated finite value past the target ends the run as target-reached and sits as Best permanently. an agent under optimization pressure grading its own homework is the classic goodhart trap.. have you thought about a verifier seam (host-run command whose output cross-checks the reported number), even as a post-v1 follow-up? asking because Id much rather my eval harness compute the number than trust the block.

packages/core/src/autoresearch/autoresearch.ts
```diff
@@ -0,0 +1,952 @@
...
+
+      if (isSplitRun(current) && current.phase === "implement") {
+        this.beginMeasurePhase(current);
+        return;
+      }
+
+      const reminders = this.remindersSent.get(runId) ?? 0;
+      if (reminders >= 1) {
+        this.endRun(runId, "failed", {
+          endReason: "missing-report",
```
DanielVisca, line 611: noticed the task chat stays live during a run — a user prompt bumps the prompt-request cursor, so an ordinary chat reply without a report block burns the one reminder, and a second one ends the run as missing-report. is that intended? feels easy to hit without realizing (tests cover unrelated-task turns but not this one).

DanielVisca (approved):
ran the autoresearch suite locally, 136 passing, and the module layout follows the core conventions really cleanly. two questions inline but nothing blocking — Ive got an MCP eval metric ready to point at this the moment it lands

DanielVisca (approved):

tatoalo (approved):

PR #3075, opened 2026-07-02
perf(sessions): make opening a big task feel fast

richardsolomou:
## Problem

Opening a task with a large transcript felt slow — up to ~3.4s before anything showed in the task log, and hover/input froze during it. Two independent causes, both scaling with transcript size:

1. The diff highlighter's worker pool self-terminates when the last diff view unmounts, so opening a diff-heavy task after navigating away re-pays Shiki's one-time init (WASM regex engine + theme normalization, ~1.5–2.2s, off-main but it delays diff content).
2. The fast "tail-first" paint only read the last 1.5MB of the log. A heavily-resumed task appends megabytes of `config_option_update` / `current_mode_update` / resume blobs to the end, so the paint landed on pure noise, showed an empty transcript, and the user waited for the full read + build + render.

## Changes

- **Keep the diff worker pool warm app-wide.** Mount one `WorkerPoolContextProvider` at the app root so the pool's mount count never hits zero. Shiki inits once at startup; every subsequent diff render reuses the warm worker.
- **Seek real content in the fast paint.** Replace the fixed-tail `readLocalLogsTail` with a byte-windowed `readLocalLogsWindow` and page back past the config-noise until real messages are in view, then paint them. The full read still loads the complete transcript behind the paint.

Both are contained; no behavior change for small tasks (their tail already holds messages). Does not touch the separate ~1s of synchronous main-thread work (parse/build/render) — that's a bigger, off-main-parsing change left for later.

## How did you test this?

- Unit tests: `readLocalLogsWindow` (whole-file, tail-with-partial-line-drop, line-boundary, single-line-over-max, backwards paging reconstructs the file, missing log); `paintTailFirst` (paints when the tail holds content, pages back past config-noise, no-ops for missing host read / existing session / empty content).
- Typecheck + existing session suites pass.
- Measured on a heavily-resumed 20MB / 9k-event task via CDP CPU profiling: time-to-first-content **3364ms → ~1066ms** (~3x). The ~2.2s of off-main Shiki init drops out of the open path entirely; the fast paint then shows recent messages in ~1s (config-noise paging) — a typical task paints in ~150ms.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/apps/code/src/main/di/container.ts
+++ b/apps/code/src/main/di/container.ts
@@ -711,8 +711,11 @@ container.bind(LOGS_SERVICE).toDynamicValue((ctx) => {
     },
     readLocalLogs: (taskRunId: string) =>
       ws.localLogs.read.query({ taskRunId }),
-    readLocalLogsTail: (taskRunId: string, maxBytes: number) =>
-      ws.localLogs.readTail.query({ taskRunId, maxBytes }),
+    readLocalLogsWindow: (
+      taskRunId: string,
+      endOffset: number | null,
+      maxBytes: number,
+    ) => ws.localLogs.readWindow.query({ taskRunId, endOffset, maxBytes }),
     writeLocalLogs: (taskRunId: string, content: string) =>
       ws.localLogs.write.mutate({ taskRunId, content }),
   };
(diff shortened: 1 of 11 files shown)


richardsolomou:
Closing — verified benchmarks show no measurable open-speed win (equal warm, slightly slower cold due to serial config-noise paging), and the warm diff-worker pool would pin ~8 Shiki workers in memory for the whole session. The task-open speedup already shipped in #3063. Details in the thread.

PR #3088, opened 2026-07-02
feat(settings): add toggle to disable in-app toast notifications

pauldambra:
## Summary
- Adds a `toastNotifications` setting (default on) alongside the existing push/dock notification toggles
- Gates all non-error toasts on it in the shared toast wrapper's `emit()` — a single choke point every `toast.*` call in the app goes through, so no per-feature changes were needed
- Error toasts are exempt and always show
- Adds an "In-app toasts" toggle to the Notifications settings page, wired into "Reset to defaults"

## Test plan
- [x] `pnpm --filter @posthog/ui typecheck`
- [x] `pnpm --filter @posthog/ui test` (28/28 passing, including new parameterized suppression cases)
- [x] `biome check` on changed files

---

### DIFF
--- a/packages/ui/src/features/settings/settingsStore.ts
+++ b/packages/ui/src/features/settings/settingsStore.ts
@@ -111,6 +111,7 @@ interface SettingsStore {
   desktopNotifications: boolean;
   dockBadgeNotifications: boolean;
   dockBounceNotifications: boolean;
+  toastNotifications: boolean;
   completionSound: CompletionSound;
   completionVolume: number;
   scaleSoundWithTaskLength: boolean;
@@ -118,6 +119,7 @@ interface SettingsStore {
   setDesktopNotifications: (enabled: boolean) => void;
   setDockBadgeNotifications: (enabled: boolean) => void;
   setDockBounceNotifications: (enabled: boolean) => void;
+  setToastNotifications: (enabled: boolean) => void;
   setCompletionSound: (sound: CompletionSound) => void;
   setCompletionVolume: (volume: number) => void;
   setScaleSoundWithTaskLength: (enabled: boolean) => void;
@@ -193,6 +195,7 @@ export const NOTIFICATION_DEFAULTS = {
   desktopNotifications: true,
   dockBadgeNotifications: true,
   dockBounceNotifications: false,
+  toastNotifications: true,
   completionSound: "none" as CompletionSound,
   completionVolume: 80,
   scaleSoundWithTaskLength: false,
@@ -260,6 +263,7 @@ export const useSettingsStore = create<SettingsStore>()(
         set({ dockBadgeNotifications: enabled }),
       setDockBounceNotifications: (enabled) =>
         set({ dockBounceNotifications: enabled }),
+      setToastNotifications: (enabled) => set({ toastNotifications: enabled }),
       setCompletionSound: (sound) => set({ completionSound: sound }),
       setCompletionVolume: (volume) => set({ completionVolume: volume }),
       setScaleSoundWithTaskLength: (enabled) =>
@@ -394,6 +398,7 @@ export const useSettingsStore = create<SettingsStore>()(
         desktopNotifications: state.desktopNotifications,
         dockBadgeNotifications: state.dockBadgeNotifications,
         dockBounceNotifications: state.dockBounceNotifications,
+        toastNotifications: state.toastNotifications,
         completionSound: state.completionSound,
         completionVolume: state.completionVolume,
         scaleSoundWithTaskLength: state.scaleSoundWithTaskLength,
--- a/packages/ui/src/primitives/toast.test.ts
+++ b/packages/ui/src/primitives/toast.test.ts
@@ -109,6 +109,17 @@ describe("toast wrapper", () => {
     expect(quill.error).toHaveBeenCalled();
   });
 
+  it.each([
+    [undefined, undefined],
+    ["stable-id", "stable-id"],
+  ])(
+    "returns %s (not a fabricated id) when suppressed with detail id %s",
+    (id, expected) => {
+      settings.toastNotifications = false;
+      expect(toast.success("Title", id ? { id } : undefined)).toBe(expected);
+    },
+  );
+
   it("after a toast closes, its id frees up to create again", () => {
     toast.success("First", { id: "dup" });
     quill.success.mock.calls[0]?.[0]?.onClose?.(); // simulate the toast closing
--- a/packages/ui/src/primitives/toast.test.ts
+++ b/packages/ui/src/primitives/toast.test.ts
@@ -22,11 +22,18 @@ const quill = vi.hoisted(() => {
 
 vi.mock("@posthog/quill", () => ({ toast: quill }));
 
+const settings = vi.hoisted(() => ({ toastNotifications: true }));
+
+vi.mock("@posthog/ui/features/settings/settingsStore", () => ({
+  useSettingsStore: { getState: () => settings },
+}));
+
 import { toast } from "./toast";
 
 beforeEach(() => {
   vi.clearAllMocks();
   quill._reset();
+  settings.toastNotifications = true;
   // clearAllMocks resets the level fns to undefined returns; restore ids.
   let n = 0;
   for (const key of [
@@ -85,6 +92,21 @@ describe("toast wrapper", () => {
     expect(quill.error).toHaveBeenCalledWith(
       expect.objectContaining({ timeout: 0 }),
     );
+  });
+
+  it.each(["success", "info", "warning", "loading"] as const)(
+    "suppresses %s when toast notifications are disabled",
+    (level) => {
+      settings.toastNotifications = false;
+      toast[level]("Title");
+      expect(quill[level]).not.toHaveBeenCalled();
+    },
+  );
+
+  it("always shows error toasts, even when toast notifications are disabled", () => {
+    settings.toastNotifications = false;
+    toast.error("Offline");
+    expect(quill.error).toHaveBeenCalled();
   });
 
   it("after a toast closes, its id frees up to create again", () => {
--- a/packages/ui/src/primitives/toast.tsx
+++ b/packages/ui/src/primitives/toast.tsx
@@ -1,4 +1,5 @@
 import { toast as quillToast } from "@posthog/quill";
+import { useSettingsStore } from "@posthog/ui/features/settings/settingsStore";
 
 // Thin wrapper over quill's toast so the whole app shares one import and a
 // stable `(title, options)` signature. Quill (base-ui under the hood) owns
@@ -45,6 +46,11 @@ function emit(
   defaultTimeout?: number,
 ): string {
   const o = normalize(detail);
+  // Toasts can be disabled in settings; errors always show since they carry
+  // information the user needs regardless of that preference.
+  if (level !== "error" && !useSettingsStore.getState().toastNotifications) {
+    return o.id ?? "";
+  }
   // base-ui auto-dismisses any non-loading toast with `timeout > 0`; it has no
   // Infinity special-case (Infinity would fire immediately), so a request to
   // never auto-dismiss maps to `0`.
--- a/packages/ui/src/primitives/toast.tsx
+++ b/packages/ui/src/primitives/toast.tsx
@@ -44,12 +44,12 @@ function emit(
   title: string,
   detail: Detail | undefined,
   defaultTimeout?: number,
-): string {
+): string | undefined {
   const o = normalize(detail);
   // Toasts can be disabled in settings; errors always show since they carry
   // information the user needs regardless of that preference.
   if (level !== "error" && !useSettingsStore.getState().toastNotifications) {
-    return o.id ?? "";
+    return o.id;
   }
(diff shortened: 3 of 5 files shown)


packages/ui/src/primitives/toast.tsx
```diff
@@ -1,4 +1,5 @@
 import { toast as quillToast } from "@posthog/quill";
+import { useSettingsStore } from "@posthog/ui/features/settings/settingsStore";
```
pauldambra, line 2: > [!NOTE]

the primitive now imports a feature store — `toast.tsx` lives in `primitives/` but reaches up into `features/settings/settingsStore`, and a bunch of settings files already import the toast primitive, so there's a soft cycle (settings feature -> toast primitive -> settings store). It works because `getState()` is read lazily inside `emit()` rather than at module load, so no runtime blow-up — but it's a mild layering inversion within `@posthog/ui` (primitives are supposed to be the leaves). Worth a sentence of thought: would the gate sit more naturally the other way round, with the store exposing the preference and the primitive staying dumb? Not blocking either way, and it stays within the package so no import-boundary lint should fire — just flagging the inversion for awareness.

packages/ui/src/primitives/toast.tsx
```diff
@@ -45,6 +46,11 @@ function emit(
   defaultTimeout?: number,
 ): string {
   const o = normalize(detail);
+  // Toasts can be disabled in settings; errors always show since they carry
+  // information the user needs regardless of that preference.
+  if (level !== "error" && !useSettingsStore.getState().toastNotifications) {
```
pauldambra, line 51: > [!NOTE]

**[convergent: qa-team/generalist-a + qa-team/generalist-b + paul + xp]** 🟢 LOW

Four independent reviewers converged on this early-return branch from different angles:

- **Fabricated/unregistered id** — when suppressed, `emit()` returns `o.id ?? ""` without ever creating the quill toast or registering it in `idRegistry`. For the no-id case that's a made-up empty-string id; for the `{ id }` upsert case, the stable id is returned but *not* recorded, so if the setting is toggled back on and the same id reused, the wrapper treats it as a fresh create instead of an update.
- **No observability on suppression** — the swallow is silent. `SETTING_CHANGED` analytics on the toggle itself covers "how many people turned this off", which is probably the important half, but there's no signal at the swallow site itself if someone reports "the app went quiet."
- **Suppressed-upsert stranding** — toggling the setting off mid-flight leaves an already-visible stable-id toast (e.g. a `loading` -> `success` upsert chain) stuck with stale content, since the gate short-circuits before the update/replace path runs.
- **Is suppressing `loading` deliberate?** — a `toast.loading("Working…")` meant to be `dismiss`ed later now never creates the quill toast, so the "something is happening" affordance disappears while notifications are off. `dismiss` on the empty-string id is harmless, but worth confirming this is an intentional inclusion rather than incidental.

None of these bite today (current callers only ever pass the return value to `dismiss`, which no-ops safely on `""`), but they're all worth a one-line acknowledgement given how much independent attention this single early return drew. Not blocking.

packages/ui/src/features/settings/settingsStore.ts
```diff
@@ -193,6 +195,7 @@ export const NOTIFICATION_DEFAULTS = {
   desktopNotifications: true,
   dockBadgeNotifications: true,
   dockBounceNotifications: false,
+  toastNotifications: true,
```
pauldambra, line 198: > [!NOTE]

`rendererStorage`'s `getItem` is async, so zustand `persist` rehydrates after boot; until then the store serves the default `toastNotifications: true`, meaning a user who disabled toasts still gets any early-boot toasts for a brief moment. The failure direction is benign here (it can over-show, never wrongly suppress, and error toasts are exempt anyway) — informational only, worth knowing if this flag is ever used to gate something more intrusive.

packages/ui/src/primitives/toast.tsx
```diff
@@ -45,6 +46,11 @@ function emit(
   defaultTimeout?: number,
 ): string {
   const o = normalize(detail);
+  // Toasts can be disabled in settings; errors always show since they carry
```
pauldambra, line 49: > [!NOTE]

**[xp]** ⚪ NIT

The gate condition `level !== "error" && !useSettingsStore.getState().toastNotifications` reads clearly enough that an extracted `isSuppressed(level)` predicate would carry the intent in its name rather than needing the comment above it (house style: prefer rename/extract over comment). That said, the *why* ("errors carry information the user needs") is genuine domain rationale rather than a restatement of the code, so that half is worth keeping regardless. Minor lean, not a must.

pauldambra (commented):

pauldambra:
> [!NOTE]
>

Small, well-scoped change. The centralized gate in `toast.tsx`'s `emit()` is the right call — one choke point instead of sprinkling guards across every feature. No CRITICAL/HIGH findings; one MEDIUM UX papercut plus several LOW/NIT items, none blocking.

### Key findings

- 🟡 **MEDIUM** (convergent: qa-team/frontend + qa-team/generalist-b + qa-team/copy) — The settings-page "Test" harness ("In-app toast" / "Deep-link toast" Send buttons, `NotificationsSettings.tsx:503` & `:512`) silently does nothing when the new toggle is off, since it routes through the now-gated `NotificationBus`. Couldn't post this inline since those lines aren't part of this PR's diff (pre-existing code). Suggest disabling those buttons with a hint, or having the harness bypass the gate.
- 🟢 **LOW** (convergent: paul + xp) — `toast.tsx` (a primitive) now imports the settings feature store — a mild layering inversion within `@posthog/ui`. Works today since `getState()` is read lazily, not a real cycle risk, but worth a thought.
- 🟢 **LOW** (convergent: qa-team/generalist-a + qa-team/generalist-b + paul + xp) — The suppressed-emit early return (`toast.tsx:51`) drew independent attention from all four reviewers: it returns a fabricated/unregistered id, gives no observability into suppression, can strand a stale toast if the setting is toggled mid-upsert, and raises the question of whether suppressing `loading` toasts is deliberate. None of these bite today.
- 🟢 **LOW** (qa-team/generalist-a) — `notifications.ts:49`'s comment ("the in-app toast always shows") is now stale post-PR. Not inline-commentable (file untouched by this PR).
- 🟢 **LOW** (qa-team/frontend) — `useArchiveTask.ts:200`'s "Task archived" toast carries the only Undo action; disabling toasts silently drops undo-archive. Worth a deliberate call (exempt action-bearing toasts, or accept the tradeoff). Not inline-commentable (file untouched by this PR).
- 🟢 **LOW** (qa-team/generalist-b) — Async store hydration means a brand-new session briefly serves the default (`true`) before the persisted value loads. Benign direction (over-shows, never under-shows).
- ⚪ **NIT** (xp) — `toast.tsx:49`'s gate condition could be an extracted `isSuppressed(level)` predicate instead of a comment.

### Convergence

- Layering inversion (toast.tsx importing settingsStore) — paul + xp, independently.
- Suppressed-emit return-value cluster (toast.tsx:51) — qa-team/generalist-a + qa-team/generalist-b + paul + xp, four independent angles on the same line.
- Test-harness silent-no-op — qa-team/frontend + qa-team/generalist-b + qa-team/copy, independently within the qa-team pass.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | APPROVE WITH NITS — one MEDIUM UX papercut (test harness), rest LOW/NIT; security and performance sub-agents found nothing. |
| 👤 paul | "stamp anyway" — clean, well-scoped, loves the parameterized tests; flags the primitive/feature layering and silent-swallow observability as thoughts, not blockers. |
| 📐 xp | Textbook OnceAndOnlyOnce — one choke point instead of ShotgunSurgery across call sites; minor extract-method nit on the gate condition. |
| 🛡 security-audit | 0 findings — purely client-side local preference, no auth/network/data boundary crossed. |

---

packages/ui/src/primitives/toast.tsx
```diff
@@ -45,6 +46,11 @@ function emit(
...
+  // information the user needs regardless of that preference.
+  if (level !== "error" && !useSettingsStore.getState().toastNotifications) {
+    return o.id ?? "";
+  }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `b039f8699` — `emit()` now returns `undefined` instead of `""` when suppressed with no stable id, matching `dismiss()`'s existing `id === undefined` guard. Also updated the one affected call site (`connectivityToast.ts`'s `offlineToastId`) to the `string | undefined` contract and added a regression test.

packages/ui/src/primitives/toast.tsx
pauldambra: > [!NOTE]

Fixed the fabricated/unregistered id part in `b039f8699` — suppressed calls now return `undefined` (or the caller's own stable id) instead of a made-up `""`, so nothing invents an id that was never registered.

The other three points here are open design questions rather than concrete bugs (no observability hook, suppressed-upsert staleness on mid-flight toggle, whether suppressing `loading` is intentional) — leaving those as-is for now since they'd need a product decision rather than a localized fix, and the reviewer themselves flagged them as "worth confirming" rather than defects. Happy to pick any of them up as a follow-up if you'd like the behavior changed.

packages/ui/src/primitives/toast.tsx
pauldambra: > [!NOTE]

Out of scope for this PR — flagged as "not blocking either way" and it's a lazy `getState()` read at call time, not a module-load cycle, so nothing breaks today. Worth reconsidering if `toast.tsx` picks up more feature-store reads in future, but not worth restructuring for a single flag right now.

packages/ui/src/features/settings/settingsStore.ts
pauldambra: > [!NOTE]

Agreed this is informational only, per the review itself — the failure direction only ever over-shows (never wrongly suppresses) and error toasts are exempt from the gate entirely, so there's no user-facing harm during the brief rehydration window. Leaving as-is.

packages/ui/src/primitives/toast.tsx
pauldambra: > [!NOTE]

Agreed the comment leans that way, but you called it NIT/"minor lean, not a must" and noted the *why* half is genuine domain rationale worth keeping regardless — so the net gain of extracting `isSuppressed(level)` here is small. Leaving the single-line condition with its comment as-is.

packages/ui/src/primitives/toast.test.ts
```diff
@@ -87,6 +94,32 @@ describe("toast wrapper", () => {
...
+  });
+
+  it.each([
+    [undefined, undefined],
+    ["stable-id", "stable-id"],
+  ])(
+    "returns %s (not a fabricated id) when suppressed with detail id %s",
+    (id, expected) => {
+      settings.toastNotifications = false;
+      expect(toast.success("Title", id ? { id } : undefined)).toBe(expected);
```
pauldambra, line 119: > [!NOTE]

**[convergent: qa-team/generalist-a + xp]** ⚪ NIT

In the parameterized regression test, the `expected` tuple element always equals `id` (`id ? { id } : undefined` just re-derives what the first tuple element already encodes). Could collapse to a single parameter with `.toBe(id)`, but at two cases this is below the ThreeStrikesAndYouRefactor threshold — take it or leave it, not blocking.

pauldambra (commented):
QA Swarm round 2 review complete (follow-up fix commit). See inline comment.

pauldambra:
> [!NOTE]
>
> Round 2 — reviewing follow-up commit b039f8699 (fixes the round-1 fabricated-id finding)

## Verdict: ✅ APPROVE

All six review passes converged: `emit()`'s `string | undefined` return type change is correct and complete, verified against every call site in the repo that captures a toast return value (`connectivityToast.ts`, `ConfigureAgentsSection.tsx:295`, `useInboxCloudTaskRunner.ts:158` — all safe, two are `toast.loading`/exempt levels that never hit the suppressed branch). Tests pass (13/13 including the new regression case), typecheck clean.

### Key findings

- ⚪ **NIT** (convergent: qa-team/generalist-a + xp) — the new parameterized test's `expected` tuple element always equals `id`; could collapse to one parameter. Optional, posted inline.
- 🟢 **LOW** (qa-team/frontend, re-surfaced) — the pre-existing suppression-before-upsert-lookup edge case (a live stable-id toast created before the setting is toggled off won't receive its upsert/dismiss) was already flagged and resolved as informational in round 1's review — not re-posting as a new thread, just noting it's the same known, deferred item.
- ⚪ **NIT** (qa-team/generalist-b) — `toast.error`'s signature now advertises `string | undefined` even though it can never actually return `undefined` (error is exempt from suppression); a narrower per-level overload would encode the invariant, but for one call site the current shape is a defensible simplicity trade-off.

Nothing blocking. Ready from qa-swarm's side.

---

packages/ui/src/primitives/toast.test.ts
pauldambra: > [!NOTE]

Take-it-or-leave-it per both reviewers, below the ThreeStrikesAndYouRefactor threshold at two cases — leaving as-is.

jonathanlab (approved):

PR #3091, opened 2026-07-02
fix(sessions): fire cloud completion notification once per turn

adboio:
## What

Cloud tasks fired **duplicate completion notifications** — two rapid back-to-back "meep" sounds and two stacked in-app toasts every time a turn finished. The same re-delivery could also duplicate **transcript entries**. This dedups the stream at the source so each entry is delivered exactly once.

## Root cause

The durable cloud SSE stream re-sends the tail by id on reconnect/replay. A re-delivered log entry (e.g. a `_posthog/turn_complete`) was counted as **new** — advancing `totalEntryCount` past the renderer's `processedLineCount` guard — and emitted again, so `notifyPromptComplete` fired a second time (and the entry could reappear in the transcript). The count/offset guard can't catch this because the replay legitimately advances the count.

The notification-bus refactor (#2934) didn't cause the double; it unmasked it — the old focused-elsewhere tier was a native OS notification (which the OS coalesces), the new one stacks toasts and replays a JS sound per call.

## How

Deduplicate re-delivered entries by their **SSE `event.id`** at ingestion, in `CloudTaskService.handleSseEvent`, before anything is counted or emitted:

- Each watcher tracks the ids it has ingested on the current leg (`seenEventIds`). A log entry whose id was already seen is dropped.
- Ids are **leg-scoped**: the set is cleared on a proxy↔django leg switch, and on a watcher re-bootstrap reset (`retry()` / stream self-heal), which re-resolves the leg. In both cases the old id space is unrelated to the new one, so a retained id would false-match — and silently drop — a legitimate new entry; the re-fetched snapshot's existing content-dedup covers the gap instead.
- Entries without an id (legacy servers) fall through unchanged.

Because the stream layer no longer produces re-deliveries, the notification path stays simple — the completion notification just fires on each live turn completion, with no special de-dup bookkeeping.

> Evolution: an earlier revision added a per-turn notification guard (`cloudTurnAwaitingCompletionNotify`) in `SessionService`. Per review (thanks @jonathanmieloo), that treated the symptom; this replaces it with id-based dedup at the source, which also fixes duplicate transcript entries. `SessionService` is back to its original notify logic. A follow-up review pass then caught that the re-bootstrap reset path retained the id set (the inverse hazard: silent drops after a retry), fixed by clearing it in `resetWatcherForRebootstrap`.

## Testing

- New `CloudTaskService` test: a stream that re-sends `id: 1` delivers the entry exactly once (`["first","second"]`, not `["first","first","second"]`) — **verified failing without the dedup**.
- New `CloudTaskService` test: after `retry()` rebuilds the watcher, an entry on the new connection that reuses a previously-seen id is still delivered — **verified failing without the reset-path clear**.
- Renderer notification tests remain (parameterised, snapshot-suppression, payload, permission dedup); the re-delivery case moved upstream to the CloudTaskService suite.
- Full `@posthog/core` suite green (1921 tests), core typecheck clean, biome clean (zero `noRestrictedImports`).

### DIFF
--- a/packages/core/src/cloud-task/cloud-task.test.ts
+++ b/packages/core/src/cloud-task/cloud-task.test.ts
@@ -348,6 +348,79 @@ describe("CloudTaskService", () => {
     expect(messages).toEqual(["first", "second"]);
   });
 
+  it("delivers an id-colliding entry after a retry resets the watcher", async () => {
+    const updates: unknown[] = [];
+    service.on(CloudTaskEvent.Update, (payload) => updates.push(payload));
+
+    // Route by URL: the initial watch and the retry-triggered rebuild each
+    // fetch run detail and history, plus post-bootstrap status verification.
+    mockNetFetch.mockImplementation((input: string | Request) => {
+      const url = typeof input === "string" ? input : input.url;
+      if (url.includes("/session_logs/")) {
+        return Promise.resolve(
+          createJsonResponse([], 200, { "X-Has-More": "false" }),
+        );
+      }
+      return Promise.resolve(
+        createJsonResponse({
+          id: "run-1",
+          status: "in_progress",
+          stage: null,
+          output: null,
+          error_message: null,
+          branch: "main",
+          updated_at: "2026-01-01T00:00:00Z",
+        }),
+      );
+    });
+
+    const consoleEntry = (message: string) =>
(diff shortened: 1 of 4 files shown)


jonathanlab (approved):

adboio:
### Merge activity

* **Jul 2, 5:19 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/3091).

PR #3094, opened 2026-07-02
feat(settings): add random notification sound options

pauldambra:
## Summary

Adds two options to the notification "Sound effect" dropdown in Settings > Notifications:

- **Random (all)** — picks a fresh sound per notification from the built-in sounds plus any installed custom sounds
- **Random (custom)** — picks from the user's custom sounds only; the option is shown once at least one custom sound is installed (same condition as the existing Custom group)

## Details

- New `CompletionSound` values `random-all` / `random-custom` live alongside the `custom:<id>` sentinel rather than in `BuiltInCompletionSound`, which keys the `SOUND_URLS` asset record
- Resolution happens in `resolveSoundUrl`, which is called per notification (and per press of the "Test sound" button), so each firing draws a fresh random pick with no extra plumbing
- `random-custom` with an empty pool resolves to `null` and plays nothing, matching the existing deleted-custom-sound behaviour; removing the *last* custom sound while `random-custom` is active falls back to "None", extending the existing fallback in `removeCustomSound`

## Testing

- Extended `sounds.test.ts`: pool membership for both modes, `null` for `random-custom` with nothing installed, and a `Math.random` spy at both extremes proving custom sounds are reachable in the `random-all` pool
- Extended the parameterised `removeCustomSound` test in `settingsStore.test.ts` with the `random-custom` fallback, plus a case verifying it stays active while other custom sounds remain

---

### DIFF
--- a/packages/ui/src/features/settings/sections/NotificationsSettings.tsx
+++ b/packages/ui/src/features/settings/sections/NotificationsSettings.tsx
@@ -242,6 +242,10 @@ export function NotificationsSettings() {
             <Select.Trigger className="min-w-[100px]" />
             <Select.Content>
               <Select.Item value="none">None</Select.Item>
+              <Select.Item value="random-all">Random (all)</Select.Item>
+              {customSounds.length > 0 && (
+                <Select.Item value="random-custom">Random (custom)</Select.Item>
+              )}
               <Select.Item value="guitar">Guitar solo</Select.Item>
               <Select.Item value="danilo">I'm ready</Select.Item>
               <Select.Item value="revi">Cute noise</Select.Item>
--- a/packages/ui/src/features/settings/settingsStore.test.ts
+++ b/packages/ui/src/features/settings/settingsStore.test.ts
@@ -257,8 +257,6 @@ describe("feature settingsStore custom sounds", () => {
         JSON.stringify({ state: persistedState, version: 0 }),
       );
 
-      useSettingsStore.setState({ completionSound: "none", customSounds: [] });
-
       await useSettingsStore.persist.rehydrate();
 
       expect(useSettingsStore.getState().completionSound).toBe(
--- a/packages/ui/src/features/settings/settingsStore.test.ts
+++ b/packages/ui/src/features/settings/settingsStore.test.ts
@@ -230,6 +230,42 @@ describe("feature settingsStore custom sounds", () => {
     const persisted = JSON.parse(lastCall[1]);
     expect(persisted.state.customSounds).toEqual([sound]);
   });
+
+  it.each([
+    {
+      label: "random-custom with empty customSounds array",
+      persistedState: { completionSound: "random-custom", customSounds: [] },
+      expectedCompletionSound: "none",
+    },
+    {
+      label: "random-custom with absent customSounds key",
+      persistedState: { completionSound: "random-custom" },
+      expectedCompletionSound: "none",
+    },
+    {
+      label: "random-custom with non-empty library",
+      persistedState: {
+        completionSound: "random-custom",
+        customSounds: [sound],
+      },
+      expectedCompletionSound: "random-custom",
+    },
+  ])(
+    "rehydrate merge normalizes $label",
+    async ({ persistedState, expectedCompletionSound }) => {
+      getItem.mockResolvedValue(
+        JSON.stringify({ state: persistedState, version: 0 }),
+      );
+
+      useSettingsStore.setState({ completionSound: "none", customSounds: [] });
+
+      await useSettingsStore.persist.rehydrate();
+
+      expect(useSettingsStore.getState().completionSound).toBe(
+        expectedCompletionSound,
+      );
+    },
+  );
 });
 
 describe("feature settingsStore terminal font", () => {
--- a/packages/ui/src/features/settings/settingsStore.test.ts
+++ b/packages/ui/src/features/settings/settingsStore.test.ts
@@ -195,6 +195,11 @@ describe("feature settingsStore custom sounds", () => {
       activeSound: "meep" as CompletionSound,
       expectedSound: "meep" as CompletionSound,
     },
+    {
+      label: "last sound feeding random-custom",
+      activeSound: "random-custom" as CompletionSound,
+      expectedSound: "none" as CompletionSound,
+    },
   ])(
     "removing the $label leaves completionSound as $expectedSound",
     ({ activeSound, expectedSound }) => {
@@ -205,6 +210,14 @@ describe("feature settingsStore custom sounds", () => {
       expect(useSettingsStore.getState().completionSound).toBe(expectedSound);
     },
   );
+
+  it("keeps random-custom active while other custom sounds remain", () => {
+    useSettingsStore.getState().addCustomSound(sound);
+    useSettingsStore.getState().addCustomSound({ ...sound, id: "def" });
+    useSettingsStore.getState().setCompletionSound("random-custom");
+    useSettingsStore.getState().removeCustomSound("abc");
+    expect(useSettingsStore.getState().completionSound).toBe("random-custom");
+  });
 
   it("persists custom sounds", async () => {
     useSettingsStore.getState().addCustomSound(sound);
--- a/packages/ui/src/features/settings/settingsStore.ts
+++ b/packages/ui/src/features/settings/settingsStore.ts
@@ -45,7 +45,11 @@ export type BuiltInCompletionSound =
   | "icq";
 
 // A user-installed sound is selected by referencing its id as `custom:<id>`.
-export type CompletionSound = BuiltInCompletionSound | `custom:${string}`;
+export type CompletionSound =
+  | BuiltInCompletionSound
+  | "random-all"
+  | "random-custom"
+  | `custom:${string}`;
 
 // A notification sound the user recorded or imported. The clip is stored inline
 // as a base64 data URL so it persists with the rest of the settings (no host
@@ -276,14 +280,21 @@ export const useSettingsStore = create<SettingsStore>()(
       addCustomSound: (sound) =>
         set((state) => ({ customSounds: [...state.customSounds, sound] })),
       removeCustomSound: (id) =>
-        set((state) => ({
-          customSounds: state.customSounds.filter((s) => s.id !== id),
-          // If the deleted sound was the active one, fall back to silence.
-          completionSound:
-            state.completionSound === `custom:${id}`
+        set((state) => {
+          const customSounds = state.customSounds.filter((s) => s.id !== id);
+          // If the deleted sound was the active one, or it was the last sound
+          // feeding "random-custom", fall back to silence.
+          const soundNowUnplayable =
+            state.completionSound === `custom:${id}` ||
+            (state.completionSound === "random-custom" &&
+              customSounds.length === 0);
+          return {
+            customSounds,
+            completionSound: soundNowUnplayable
               ? "none"
               : state.completionSound,
-        })),
+          };
+        }),
       renameCustomSound: (id, name) =>
         set((state) => ({
           customSounds: state.customSounds.map((s) =>
--- a/packages/ui/src/features/settings/settingsStore.ts
+++ b/packages/ui/src/features/settings/settingsStore.ts
@@ -469,6 +469,12 @@ export const useSettingsStore = create<SettingsStore>()(
         if ((merged.autoConvertLongText as string) === "500") {
           (merged as Record<string, unknown>).autoConvertLongText = "1000";
         }
+        if (
+          merged.completionSound === "random-custom" &&
+          (!merged.customSounds || merged.customSounds.length === 0)
+        ) {
+          (merged as Record<string, unknown>).completionSound = "none";
+        }
         return merged;
       },
     },
--- a/packages/ui/src/features/settings/settingsStore.ts
+++ b/packages/ui/src/features/settings/settingsStore.ts
@@ -282,8 +282,6 @@ export const useSettingsStore = create<SettingsStore>()(
       removeCustomSound: (id) =>
         set((state) => {
           const customSounds = state.customSounds.filter((s) => s.id !== id);
-          // If the deleted sound was the active one, or it was the last sound
-          // feeding "random-custom", fall back to silence.
           const soundNowUnplayable =
             state.completionSound === `custom:${id}` ||
             (state.completionSound === "random-custom" &&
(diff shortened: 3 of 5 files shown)


packages/ui/src/features/settings/sections/NotificationsSettings.tsx
```diff
@@ -242,6 +242,10 @@ export function NotificationsSettings() {
             <Select.Trigger className="min-w-[100px]" />
             <Select.Content>
               <Select.Item value="none">None</Select.Item>
+              <Select.Item value="random-all">Random (all)</Select.Item>
+              {customSounds.length > 0 && (
```
pauldambra, line 246: > [!NOTE]

**[convergent: qa-team/frontend + qa-team/generalist-b + paul]** 🟢 LOW

The "Random (custom)" `Select.Item` is only rendered when `customSounds.length > 0`, but the store can still hold `completionSound: "random-custom"` with an empty library outside the guarded `removeCustomSound` path (partial rehydrate, app-version downgrade/upgrade window, or any future code that clears `customSounds` directly). In that state the Radix Select trigger renders blank, and the volume / scale rows stay visible (they gate on `!== "none"`) even though nothing will ever play. `resolveSoundUrl` degrades to silence so nothing breaks, and it's the same shape as an orphaned `custom:<id>` today — consider normalizing an unplayable `completionSound` back to `"none"` on rehydrate (persist `merge`) so the trigger never shows a ghost value.

packages/ui/src/utils/sounds.ts
```diff
@@ -68,15 +68,27 @@ export function playbackRateForTaskDuration(durationMs: number): number {
...
+// unknown built-in, a `custom:` id no longer installed (e.g. the active sound
+// was deleted), or `random-custom` with no sounds installed — callers then
+// play nothing.
 export function resolveSoundUrl(
   sound: CompletionSound,
   customSounds: CustomSound[],
 ): string | null {
   if (sound === "none") return null;
+  const customUrls = customSounds.map((s) => s.dataUrl);
+  if (sound === "random-all") {
```
pauldambra, line 88: > [!NOTE]

**[qa-team/generalist-a]** 🟢 LOW

`resolveSoundUrl` is now nondeterministic for the `random-*` modes, and `NotificationBus.notify` (packages/ui/src/features/notifications/notifications.ts) resolves twice per native notification — once inside `playCompletionSound` and once for `willPlaySound`. Behavior is correct today because only null-ness is consumed and that is deterministic for both modes, but the checked URL and the played URL are now different random rolls; a future change that uses the resolved URL for anything beyond truthiness would silently diverge from what actually plays. Resolving once and threading the URL through would remove the trap.

packages/ui/src/utils/sounds.ts
```diff
@@ -68,15 +68,27 @@ export function playbackRateForTaskDuration(durationMs: number): number {
...
+// random pick per call for the `random-*` modes. Returns null for `none`, an
+// unknown built-in, a `custom:` id no longer installed (e.g. the active sound
+// was deleted), or `random-custom` with no sounds installed — callers then
+// play nothing.
 export function resolveSoundUrl(
   sound: CompletionSound,
   customSounds: CustomSound[],
 ): string | null {
   if (sound === "none") return null;
+  const customUrls = customSounds.map((s) => s.dataUrl);
```
pauldambra, line 87: > [!NOTE]

**[convergent: qa-team/performance + xp]** ⚪ NIT

`customUrls` is computed on every call, including the `none`-adjacent built-in and `custom:` paths that never use it, and `random-all` allocates a fresh combined array per call. It only copies string references once per notification, so impact is negligible — moving the `.map` inside the two random branches would keep the common path allocation-free and states the intent more precisely. Not worth changing on its own.

packages/ui/src/features/settings/settingsStore.ts
```diff
@@ -276,14 +280,21 @@ export const useSettingsStore = create<SettingsStore>()(
...
+          const customSounds = state.customSounds.filter((s) => s.id !== id);
+          // If the deleted sound was the active one, or it was the last sound
+          // feeding "random-custom", fall back to silence.
+          const soundNowUnplayable =
+            state.completionSound === `custom:${id}` ||
+            (state.completionSound === "random-custom" &&
+              customSounds.length === 0);
+          return {
+            customSounds,
+            completionSound: soundNowUnplayable
```
pauldambra, line 291: > [!NOTE]

**[qa-team/frontend]** ⚪ NIT

Removing the last custom sound while "Random (custom)" is active silently flips the completion sound to "None" with no user feedback. This mirrors the pre-existing `custom:<id>` fallback so it's consistent — but a brief toast ("Completion sound reset to None") would make the setting change less surprising for both cases.

packages/ui/src/features/settings/sections/NotificationsSettings.tsx
```diff
@@ -242,6 +242,10 @@ export function NotificationsSettings() {
...
+                <Select.Item value="random-custom">Random (custom)</Select.Item>
```
pauldambra, line 247: > [!NOTE]

**[xp]** ⚪ NIT

`customSounds.length > 0` now guards both the "Random (custom)" item and the existing Custom group in this component — the same domain rule ("there are custom sounds") expressed twice in the JSX. A small local `hasCustomSounds` const would dedupe the two uses and read a touch better; not worth a shared abstraction.

pauldambra (commented):

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE

Small, well-tested feature with zero critical/high/medium findings across all four review perspectives. Everything flagged is a low-severity polish item or nit; nothing blocks merge.

### Key findings

**🟢 Low**
- `random-custom` can persist with an empty custom-sound library outside the guarded `removeCustomSound` path (rehydrate / downgrade window), leaving a blank Select trigger with orphaned volume/scale rows — normalize on rehydrate (convergent: 3 reviewers)
- `resolveSoundUrl` is now nondeterministic and `NotificationBus.notify` resolves twice per native notification; only null-ness is consumed today so behaviour is correct, but the played and checked URLs are different rolls — a latent trap

**⚪ Nits**
- `customUrls` is mapped on every call including paths that never use it (negligible)
- Silent fallback to "None" when the last custom sound is removed could use a toast
- `customSounds.length > 0` guard now appears twice in the JSX — small local `hasCustomSounds` const

### Convergence

- The orphaned `random-custom` + empty-library state was flagged independently by qa-team/frontend, qa-team/generalist-b, and paul-reviewer — highest confidence finding
- The per-call `customUrls` allocation was flagged independently by qa-team/performance and xp-reviewer

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Overall LOW: correct branch ordering, safe empty pools, solid tests; one convergent low-severity ghost-state gap |
| 👤 paul | "nice little feature, and honestly the tests carry it... perfetto. ship as you see fit" |
| 📐 xp | Clean work at "make it right"; strong parameterized tests, only minor duplication and incidental allocation |
| 🛡 security-audit | No exploitable vulnerabilities: pool is build-time assets + user's own local data URLs into a media-fetch sink; no trust boundary crossed |

---

packages/ui/src/features/settings/sections/NotificationsSettings.tsx
pauldambra: > [!NOTE]

Fixed in 81b8e580a — added a guard in the existing `merge` function: if `completionSound === "random-custom"` and `customSounds` is empty on rehydrate, it falls back to `"none"`, making the bad state unrepresentable at the persist boundary.

packages/ui/src/utils/sounds.ts
pauldambra: > [!NOTE]

Out of scope — the double-resolve is a pre-existing pattern and behavior is correct today. A follow-up refactor to resolve once and thread the URL through is worth doing but not in this PR.

packages/ui/src/utils/sounds.ts
pauldambra: > [!NOTE]

Intentional — allocation is negligible (string references only) and the reviewer noted it's not worth changing on its own.

packages/ui/src/features/settings/settingsStore.ts
pauldambra: > [!NOTE]

Out of scope — a toast on implicit sound reset is a UX enhancement but adds noise for this PR. Consistent with the existing custom:<id> fallback behavior. Could be a follow-up.

packages/ui/src/features/settings/sections/NotificationsSettings.tsx
pauldambra: > [!NOTE]

Intentional — the reviewer explicitly noted this is not worth a shared abstraction. The two uses are close together and readable as-is.

packages/ui/src/features/settings/settingsStore.ts
```diff
@@ -458,6 +467,12 @@ export const useSettingsStore = create<SettingsStore>()(
         if ((merged.autoConvertLongText as string) === "500") {
           (merged as Record<string, unknown>).autoConvertLongText = "1000";
         }
+        if (
```
pauldambra, line 470: > [!NOTE]

**[convergent: qa-team/generalist-a + qa-team/generalist-b + paul + xp]** 🟢 LOW

The new `merge()` normalization (orphaned `random-custom` -> `"none"` on rehydrate) is the actual fix this delta ships, yet it has no test. The runtime `removeCustomSound` path is well covered by the parameterized cases, but this is a different code path. The test file already has the exact rehydrate pattern needed (`getItem.mockResolvedValue(...)` + `useSettingsStore.persist.rehydrate()`): a parameterized case covering `{"completionSound":"random-custom","customSounds":[]}` -> `"none"`, `customSounds` key absent -> `"none"`, and non-empty library -> `"random-custom"` preserved would be a few lines and pins the exact downgrade/partial-rehydrate regression this commit exists to prevent.

packages/ui/src/features/settings/settingsStore.ts
```diff
@@ -458,6 +467,12 @@ export const useSettingsStore = create<SettingsStore>()(
...
+          merged.completionSound === "random-custom" &&
+          (!merged.customSounds || merged.customSounds.length === 0)
```
pauldambra, line 472: > [!NOTE]

**[convergent: paul + qa-team/generalist-b]** 🟢 LOW

silly question... the original thread also mentioned orphaned `custom:<id>` values showing the same ghost state in the Select. This merge block only normalizes `random-custom`, not a persisted `custom:gone` whose id is no longer in the rehydrated `customSounds` — both are guarded identically by `removeCustomSound` at runtime, so both only leak in via the same downgrade/manual-edit paths. It's pre-existing behaviour and the delta is a strict improvement, so not blocking — but if you're already in here it'd be one more condition to close it symmetrically (`completionSound.startsWith("custom:") && no matching id in customSounds` -> `"none"`). Curious whether it was left deliberately.

pauldambra (commented):
QA Swarm follow-up review of the new commits (81b8e58, fca242c). See inline comments.

pauldambra:
> [!NOTE]
>
> Follow-up review of commits 81b8e58 + fca242c (merge() normalization fix, comment removal)

## Verdict: ✅ APPROVE

The delta is a correct, well-targeted fix for the prior convergent finding — it handles missing-key, null, and empty-array shapes, correctly leaves `random-all` alone, and security review confirms it strictly narrows reachable state. Two low-severity follow-ups flagged, nothing blocking.

### Key findings

**🟢 Low**
- The new `merge()` normalization has no test, despite the test file having the exact rehydrate pattern needed (convergent: 4 reviewers)
- The fix closes the `random-custom` ghost but not the equally-reachable orphaned `custom:<id>` ghost (pre-existing, convergent: 2 reviewers)

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | LOW — fix verified correct across all persisted-state permutations; missing test is the one gap |
| 👤 paul | "nice fix... no correctness bugs in the delta. ship as you see fit" |
| 📐 xp | Clean, minimal, in the right place; two-place guard logic is acceptable divergence, not a OnceAndOnlyOnce violation |
| 🛡 security-audit | Zero findings — writes a hardcoded literal, strictly reduces reachable state |

---

packages/ui/src/features/settings/settingsStore.ts
pauldambra: > [!NOTE]

Added a parameterized rehydrate test covering all three cases in d9f01558d: empty `customSounds` array -> `"none"`, absent `customSounds` key -> `"none"`, non-empty library -> `"random-custom"` preserved. All 41 tests pass.

packages/ui/src/features/settings/settingsStore.ts
pauldambra: > [!NOTE]

Left deliberately out of scope for this PR. The `custom:<id>` orphan case is pre-existing behaviour that predates this change — adding it here would widen the scope beyond the fix this PR exists to ship (`random-custom` with an empty library). A follow-up task or PR can add the symmetric condition once the current change is merged.

packages/ui/src/features/settings/settingsStore.test.ts
```diff
@@ -217,6 +230,40 @@ describe("feature settingsStore custom sounds", () => {
...
     expect(persisted.state.customSounds).toEqual([sound]);
   });
+
+  it.each([
+    {
+      label: "random-custom with empty customSounds array",
+      persistedState: { completionSound: "random-custom", customSounds: [] },
+      expectedCompletionSound: "none",
+    },
+    {
```
pauldambra, line 240: > [!NOTE]

**[qa-team]** ⚪ NIT

Cases 1 and 2 of the new `rehydrate merge normalizes` `it.each` expect `"none"`, which is exactly the `completionSound` the describe's `beforeEach` seeds — each case individually would pass vacuously if `persist.rehydrate()` silently no-oped (storage key rename, malformed envelope). The suite as a whole is still sound: mutation-testing confirmed both cases fail when the merge() normalization block is deleted, and case 3 (`"random-custom"` != seed) would catch a fully broken rehydrate. Seeding `completionSound` to a non-`"none"` sentinel before rehydrate (or also asserting `customSounds` post-rehydrate) would make each case self-sufficient. Optional hardening, not a gap.

pauldambra (commented):
QA Swarm follow-up review of the rehydrate-test commits (d9f0155, 729b5ea). One optional nit; verdict unchanged (APPROVE).

packages/ui/src/features/settings/settingsStore.test.ts
pauldambra: > [!NOTE]

NIT acknowledged. The suite is sound — mutation testing confirmed both cases fail when the normalization block is deleted, and case 3 with a distinct sentinel provides additional coverage. The optional per-case sentinel hardening is a nice-to-have but not required for correctness. Resolving.

PR #3104, opened 2026-07-02
fix(task-input): serve repo/branch selector from a persisted cache

fercgomes:
Closes [GROW-73](https://linear.app/posthog-team-growth/issue/GROW-73/posthog-code-repobranch-selector-takes-too-long-to-load)

## Problem

The task composer's repo/branch selector shows loading states for too long, especially on first open:

- **Branches had no cold-start cache.** The repo picker already persists `cachedCloudRepositoryMap`, but the branch list paid the full PostHog → GitHub round trip on every app launch — the branch button sat on "Loading..." until it resolved, and the query couldn't even start until the repo list had loaded.
- **Opening the repo picker always flashed "Loading repositories..."** because the picker's server-search query starts fresh on open, even though the full repo list was already in memory.

## Changes

**Persisted branch cache** (the main fix). `useUserGithubBranches` now keeps the last-known branch list + default branch per repo in the settings store (`cachedCloudBranchMap`, next to the existing repo cache). On cold start or repo switch, the selector renders the cached list and auto-selects the cached default branch instantly while the live query revalidates behind the refresh spinner. Fresh results are written back after each successful fetch — unsearched first page only (50 branches, all the combobox renders anyway), LRU-capped at 20 repos, cleared when a repo cleanly reports no branches or integrations are removed. Most branches being worked on never change, so serving the stale list is almost always right.

**Repo picker open fallback.** While the picker's search query loads its first page with no search typed, it falls back to the already-loaded full repository list instead of an empty loading state. Also adds keep-previous-data (`placeholderData`) to the user repo search queries so typing doesn't blank the list (the org-level hook already did this).

Cache decisions are pure functions in `@posthog/core` (`resolveEffectiveBranches`, `resolveBranchCacheUpdate`, `resolveRepositoryPickerList`) with the hooks as thin glue, mirroring the existing `useUserRepositoryIntegration` pattern.

## Notes

- A branch deleted remotely can briefly appear from cache until revalidation completes — the same window the existing 5-minute in-memory `staleTime` already allowed, just extended across restarts.
- 32 new unit tests for the cache resolvers (skip/write/reorder/evict/clear, cache-vs-live precedence, search + error handling).
- Full suites pass: core 1943 tests, UI 1205 tests, `pnpm typecheck` across all packages, Biome clean.

### DIFF
--- a/packages/core/src/integrations/branches.test.ts
+++ b/packages/core/src/integrations/branches.test.ts
@@ -2,10 +2,15 @@ import { describe, expect, it } from "vitest";
 import {
   BRANCHES_FIRST_PAGE_SIZE,
   BRANCHES_PAGE_SIZE,
+  type BranchCacheUpdateInputs,
   branchPageSizeForOffset,
+  type CachedCloudBranchMap,
   computeNextBranchOffset,
   flattenBranchPages,
   type GithubBranchesPage,
+  MAX_CACHED_BRANCH_REPOS,
+  resolveBranchCacheUpdate,
+  resolveEffectiveBranches,
 } from "./branches";
 
 const page = (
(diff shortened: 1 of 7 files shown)


charlesvien:

PR #3136, opened 2026-07-03
fix(sessions): persist run config on start-failure error session

robbie-c:
## Problem

Follow-up to #3127. That PR preserved the user's selected model across retries when a fully-created session later errored — but not when the initial session *start* times out. In that case the error session built in the connect catch-path carried only the initial prompt, so the auto-retry recreated the session without the model/adapter/mode/effort and fell back to the default model. This is the exact "Failed to create session" timeout users were hitting, so the selected model (e.g. Fable) still silently reverted to the default.

## Changes

- Add a `stampRunConfig` helper that persists the prompt, model, adapter, execution mode, and reasoning level onto a session in one place.
- Apply it at both error-path sites (start-failure catch + no-auth), so a retry after a start timeout keeps the user's original choices.

## How did you test this?

- Added a test driving a session-start failure and asserting the stored error session carries the full run config; it passes.
- `pnpm --filter @posthog/core typecheck` and the `packages/core/src/sessions` vitest suite pass.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1783085269646269?thread_ts=1783085269.646269&cid=C09G8Q32R6F)*

### DIFF
--- a/packages/core/src/sessions/sessionService.ts
+++ b/packages/core/src/sessions/sessionService.ts
@@ -652,6 +652,16 @@ export class SessionService {
     return connectPromise;
   }
 
+  private stampRunConfig(session: AgentSession, params: ConnectParams): void {
+    session.adapter = params.adapter;
+    session.model = params.model;
+    session.executionMode = params.executionMode;
+    session.reasoningLevel = params.reasoningLevel;
+    if (params.initialPrompt?.length) {
+      session.initialPrompt = params.initialPrompt;
+    }
+  }
+
   private async doConnect(params: ConnectParams): Promise<void> {
     const {
       task,
@@ -693,9 +703,7 @@ export class SessionService {
         session.status = "error";
         session.errorMessage =
           "Authentication required. Please sign in to continue.";
-        if (initialPrompt?.length) {
-          session.initialPrompt = initialPrompt;
-        }
+        this.stampRunConfig(session, params);
         this.d.store.setSession(session);
         return;
       }
@@ -784,9 +792,7 @@ export class SessionService {
 
       const taskRunId = latestRun?.id ?? `error-${taskId}`;
       const session = createBaseSession(taskRunId, taskId, taskTitle);
-      if (initialPrompt?.length) {
-        session.initialPrompt = initialPrompt;
-      }
+      this.stampRunConfig(session, params);
       if (latestRun?.log_url) {
         try {
           const { rawEntries } = await this.fetchSessionLogs(
--- a/packages/core/src/sessions/sessionServiceRetryConfig.test.ts
+++ b/packages/core/src/sessions/sessionServiceRetryConfig.test.ts
@@ -165,3 +165,60 @@ describe("SessionService.connectToTask start failure", () => {
     ]);
   });
 });
+
+describe("SessionService.connectToTask missing auth", () => {
+  it("persists the run configuration on the auth-required error session", async () => {
+    const setSession = vi.fn();
+    const deps = {
+      store: {
+        getSessionByTaskId: () => undefined,
+        getSessions: () => ({}),
+        setSession,
+      },
+      log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
+      getIsOnline: () => true,
+      trpc: {
+        agent: {
+          onSessionIdleKilled: {
+            subscribe: () => ({ unsubscribe: vi.fn() }),
+          },
+        },
+      },
+    } as unknown as SessionServiceDeps;
+
+    const service = new SessionService(deps);
+    // No credentials → the no-auth branch stores an error session and returns
+    // before ever starting a run.
+    vi.spyOn(
+      service as unknown as {
+        getAuthCredentialsStatus: () => Promise<unknown>;
+      },
+      "getAuthCredentialsStatus",
+    ).mockResolvedValue({ kind: "missing" });
+
+    await service.connectToTask({
+      task: {
+        id: "task-1",
+        title: "Test task",
+        description: "Ship the fix",
+        latest_run: null,
+      } as unknown as Task,
+      repoPath: "/repo",
+      initialPrompt: [{ type: "text", text: "Ship the fix" }],
+      executionMode: "auto",
+      adapter: "claude",
+      model: "claude-fable-5",
+      reasoningLevel: "high",
+    });
+
+    const stored = setSession.mock.calls.at(-1)?.[0] as AgentSession;
+    expect(stored.status).toBe("error");
+    expect(stored.model).toBe("claude-fable-5");
+    expect(stored.adapter).toBe("claude");
+    expect(stored.executionMode).toBe("auto");
+    expect(stored.reasoningLevel).toBe("high");
+    expect(stored.initialPrompt).toEqual([
+      { type: "text", text: "Ship the fix" },
+    ]);
+  });
+});
--- a/packages/core/src/sessions/sessionServiceRetryConfig.test.ts
+++ b/packages/core/src/sessions/sessionServiceRetryConfig.test.ts
@@ -1,4 +1,5 @@
 import type { AgentSession } from "@posthog/shared";
+import type { Task } from "@posthog/shared/domain-types";
 import { describe, expect, it, vi } from "vitest";
 import { SessionService, type SessionServiceDeps } from "./sessionService";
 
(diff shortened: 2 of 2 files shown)


robbie-c:
> [!NOTE]
> 🤖 Automated comment by **QA Swarm** (via pr-shepherd) — not written by a human
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — not vendored in this repo)

## Verdict: ✅ APPROVE WITH NITS

Small, well-scoped change; reviewers converged on approve. No correctness or security findings. One convergent in-scope gap was auto-fixed in the loop; one pre-existing coupling item is deferred as a follow-up.

### Auto-fixed in this loop
- **Test gap (paul + xp, convergent):** `stampRunConfig` now also stamps the full run config at the auth-missing error site, but only the start-timeout path was tested → added a `connectToTask missing auth` test (`0187eaa`).
- **Simplify:** deduped the two error-path tests behind a shared `CONNECT_PARAMS` + `assertRunConfigPersisted` helper (`bfd25bb`).
- **Robustness (paul, round 2):** pinned "exactly one error session stored" so the `.at(-1)` assertion can't mask a future regression (`8a32d3f`).

### Deferred — follow-up, out of scope here
- **Pre-existing positional-arg coupling (paul + xp):** `createNewLocalSession` / `clearSessionError` pass these same five run-config values as a long positional list and hand-stamp them a *third* time, not routed through `stampRunConfig`. Giving that method a run-config object and reusing `stampRunConfig` would finish OnceAndOnlyOnce and shorten a 9-arg signature. Not blocking this PR.

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 👤 paul | approve — correct fix, coherent end to end; nits on positional-arg fragility + the (now-added) auth-site test |
| 📐 xp | approve — clean helper + real end-to-end test; the third hand-stamped copy is the remaining OnceAndOnlyOnce gap (out of scope) |
| 🛡 security-audit | no findings — client-side session-config plumbing, no user-controlled source→sink |

---
*Automated by QA Swarm / pr-shepherd — not a human review*

packages/core/src/sessions/sessionServiceRetryConfig.test.ts
```diff
@@ -90,3 +91,77 @@ describe("SessionService.clearSessionError retry config", () => {
...
+    expect(stored.status).toBe("error");
+    expect(stored.model).toBe("claude-fable-5");
+    expect(stored.adapter).toBe("claude");
+    expect(stored.executionMode).toBe("auto");
+    expect(stored.reasoningLevel).toBe("high");
+    expect(stored.initialPrompt).toEqual([
+      { type: "text", text: "Ship the fix" },
+    ]);
+  });
+});
```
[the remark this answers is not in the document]
robbie-c: Already covered — `sessionServiceRetryConfig.test.ts` has `describe("SessionService.connectToTask missing auth")` → `it("persists the run configuration on the auth-required error session")` (added in 0187eaa, currently lines 179–207). It drives the no-auth branch via `getAuthCredentialsStatus` → `{ kind: "missing" }` and asserts the stored error session retains `model`, `adapter`, `executionMode`, `reasoningLevel`, and `initialPrompt` — exactly the coverage requested. Resolving.

packages/core/src/sessions/sessionService.ts
```diff
@@ -652,6 +652,16 @@ export class SessionService {
     return connectPromise;
   }
 
+  private stampRunConfig(session: AgentSession, params: ConnectParams): void {
+    session.adapter = params.adapter;
+    session.model = params.model;
+    session.executionMode = params.executionMode;
+    session.reasoningLevel = params.reasoningLevel;
+    if (params.initialPrompt?.length) {
```
pauldambra, line 660: > [!NOTE]

**[convergent: paul + xp]** ⚪ NIT

paul: nit picking... four fields assign raw (`session.model = params.model` etc, happily writing `undefined`) but `initialPrompt` is guarded behind `if (params.initialPrompt?.length)`. it's correct — both call sites pass fresh `createBaseSession` objects so there's nothing to clobber, and it mirrors `createNewLocalSession` exactly (nice) — just that the mixed shape made me pause to work out whether the asymmetry was load-bearing. a one-liner comment would save the next reader that pause. very not blocking.

xp: The guard faithfully mirrors the old code and is defensive if `stampRunConfig` is ever reused on a non-fresh session (don't stomp an existing prompt with an empty one), so leaving it is right — but a one-line "guarded so a later non-fresh caller can't blank an existing prompt" comment would stop the next reader wondering why one field is special.

packages/core/src/sessions/sessionService.ts
```diff
@@ -652,6 +652,16 @@ export class SessionService {
     return connectPromise;
   }
 
+  private stampRunConfig(session: AgentSession, params: ConnectParams): void {
```
pauldambra, line 655: > [!NOTE]

**[xp]** 🟢 LOW

Nice extraction — `stampRunConfig` gets the run-config bundle to OnceAndOnlyOnce across both error-path sites, and since both operate on a fresh `createBaseSession(...)` the unconditional writes can't clobber real user values. Worth calling out that this is now the *third* place this same 5-field clump is written: the happy path in `createNewLocalSession` assigns adapter/model/executionMode/reasoningLevel/initialPrompt too, and `clearSessionError` reads the very same bundle back for the retry. That's a Data Clump quietly asking to become a small `RunConfig` value type. Not pushing to force the happy path through `stampRunConfig` now — it assigns from destructured locals and interleaves other state, so routing it through would need a flag or reshuffle (the wrong-abstraction trap). Flagging the clump for when it strikes a fourth time.

packages/core/src/sessions/sessionServiceRetryConfig.test.ts
```diff
@@ -90,3 +95,123 @@ describe("SessionService.clearSessionError retry config", () => {
...
+    const deps = {
+      store: {
+        getSessionByTaskId: () => undefined,
+        getSessions: () => ({}),
+        setSession,
+      },
+      log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
+      settings: { customInstructions: "" },
+      DEFAULT_GATEWAY_MODEL: "claude-opus-4-8",
+      // Online for the create-branch check, offline in the catch so the
```
pauldambra, line 142: > [!NOTE]

paul: this `getIsOnline` mock is order-coupled — `mockReturnValueOnce(true).mockReturnValue(false)` only works because the create-new path hits `getIsOnline()` once at the offline guard (true, so we proceed) and then again in the catch for `shouldAutoRetry` (false, so we stop before the retry loop). i traced it and it's exactly right today... but positional mocks always make me a little twitchy — they silently rot if someone adds another `getIsOnline()` check upstream. not blocking, the comment carries it.

xp: Agreed on the mechanism, opposite emphasis: the `expect(setSession.mock.calls.length).toBe(1)` guard is exactly the backstop for this — if the online/offline sequencing ever drifts and the auto-retry loop spins, the test fails loudly instead of `.at(-1)` silently reading an intermediate session. No change wanted.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete (round for 46aa502). See inline comments and the summary comment.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — reviewer files not installed in this environment)

## Verdict: ✅ APPROVE

No security findings and no correctness issues; all findings are LOW/NIT observations explicitly marked not-blocking by their reviewers. The fix is tight, both stamp sites operate on fresh sessions, and the stamped fields are exactly what the retry path reads back.

### Key findings

- 🟢 LOW `sessionService.ts:655` (xp) — the 5-field run-config bundle is now written in three places (stampRunConfig, createNewLocalSession, read back in clearSessionError): a latent Data Clump that wants a `RunConfig` value type *eventually*; forcing it now would risk the wrong abstraction.
- 🟢 LOW `sessionServiceRetryConfig.test.ts:142` (paul + xp) — the two-phase `getIsOnline` mock is order-coupled; the `calls.length === 1` assertion is the backstop that makes it safe. No change wanted.
- ⚪ NIT `sessionService.ts:660` (paul + xp, convergent) — the guarded `initialPrompt` vs raw-assigned other fields asymmetry could use a one-line comment for the next reader.

### Convergence

- The `initialPrompt` guard asymmetry was flagged independently by paul and xp (higher confidence it will trip the next reader).
- The positional `getIsOnline` mock drew both reviewers: paul flags the fragility, xp endorses the existing `calls.length` backstop as the right mitigation.

### Reviewer divergence (not posted inline)

paul floated merging the two new tests into an `it.each`; xp explicitly endorsed keeping them separate, citing the repo convention (parameterise when only inputs/expectations differ; keep separate tests when setup and intent differ — they do here). Left as-is.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | skipped — reviewer skill files not available in this environment |
| 👤 paul | tidy little fix; only real flag is the order-coupled mock, already commented and backstopped; "ship as you see fit" |
| 📐 xp | well-scoped, verified end-to-end that the stamped config is exactly what retry reads back; data clump is a "later", not a "now"; "Ship it" |
| 🛡 security-audit | zero findings — client-side trust domain, no persistence beyond in-memory store, no new logging of prompt content, fixed-key assignment on fresh objects |

---

packages/core/src/sessions/sessionService.ts
pauldambra: > [!NOTE]

Added the one-liner comment "Guarded so a later non-fresh caller can't blank an existing prompt with an empty one." above the guard — fix applied, landing in the next shepherd commit.

packages/core/src/sessions/sessionService.ts
pauldambra: > [!NOTE]

Noted — the Data Clump flag is a useful marker for when a fourth site emerges. No change in this PR per reviewer guidance.

packages/core/src/sessions/sessionServiceRetryConfig.test.ts
pauldambra: > [!NOTE]

No change wanted per reviewer — the existing  guard provides the loud backstop if the sequencing ever drifts.

pauldambra:
> [!NOTE]
>
> Round 2 (incremental, 46aa502..7411069): paul-reviewer, xp-reviewer, security-audit

## Verdict: ✅ APPROVE

Zero findings from all three reviewers. The delta — the requested `initialPrompt`-guard comment and the simplify pass on the assertion helper — does exactly what round 1 asked. xp verified the new comment states a real, file-wide invariant (the same guard-before-assign pattern protects `initialPrompt` at every write site in the file), and both paul and xp confirmed `toHaveBeenCalledTimes(1)` + `calls[0][0]` is equivalent-and-cleaner. security-audit: no new data flows or trust-boundary changes; round-1 clean verdict stands.

One optional-polish observation (not a finding, no change requested): the pre-existing comment above the assertion still says "last-call assertions" while the code now reads `calls[0][0]` — with exactly one call pinned, first == last, so the reasoning stays self-consistent.

---

PR #3141, opened 2026-07-03
perf(worktree): stop worktree session creation blocking on large-repo scans and deletes

pauldambra:
## Problem

Starting a new worktree session in a large repo (posthog/posthog) gets slower and slower over a week of use. Three compounding causes, all measured on a real posthog checkout:

1. **~20s of ls-files walks per create.** `processWorktreeLink`/`processWorktreeInclude` each run `git ls-files --ignored --others --directory --exclude-from=<file>`. Because `--exclude-from` replaces the standard excludes, git cannot collapse `node_modules/` and recurses the entire ignored tree looking for pattern matches (9.6s + 9.8s, warm cache, growing with ignored-file count). The walk also matched strays like `node_modules/.pnpm/psl@1.9.0/node_modules/psl/.env`, and copying that pre-created `node_modules/.pnpm/` in the fresh worktree — silently defeating posthog's own `.husky/post-checkout` bootstrap check (`[ ! -d node_modules/.pnpm ]`).
2. **Creation blocks on suspending the least-recent task once at the worktree cap.** `doCreateWorkspace` awaited `suspendLeastRecentIfOverLimit()` before creating: a full checkpoint capture plus deletion of a ~6GB, node_modules-laden worktree, inline, before `git worktree add` even starts. Under the cap (early in the week) this branch is skipped entirely, which is why creation "became" slow after days of use.
3. **Worktree deletes rm multi-gigabyte trees inside the repo write lock.** Every suspend, archive, and hourly inactivity sweep serialized with worktree creation on the per-repo write lock while unlinking ~1M files.

## Changes

- `getIgnoredPathsFromExcludeFile` now lists candidates with `--exclude-standard --exclude-from=<file>` (so git collapses standard-ignored trees instead of recursing them) and re-applies the exclude file's own patterns in-process via a new gitignore-subset matcher (`exclude-patterns.ts`: comments, negation, dir-only, anchoring, `*`/`?`/`**`/`[...]`). Matches buried inside standard-ignored trees no longer surface, fixing the node_modules bootstrap bug. Repos without `.worktreelink`/`.worktreeinclude` now skip the git calls entirely.
- The worktree cap is enforced *after* workspace creation as a fire-and-forget, instead of blocking creation. `suspendLeastRecentIfOverLimit` now suspends exactly the excess over the cap, oldest first (previously it fired when merely *at* the cap, pre-create). Net behavior change: the active count can briefly reach `maxActiveWorktrees + 1` while the background suspend runs.
- `deleteWorktree` renames the worktree into `<worktreeBase>/.trash/` (instant, same volume), runs `git worktree prune` under the write lock, and does the recursive removal in the background. Falls back to the old `git worktree remove --force` path when the rename cannot work (cross-volume, already gone). Boot sweeps `.trash/` for leftovers from interrupted deletes, alongside the existing orphan cleanup.

Measured on posthog/posthog: the `.worktreeinclude` scan went from ~10s to ~140ms with identical output minus the node_modules stray; worktree deletion no longer holds the write lock for the duration of a multi-gigabyte rm.

## How did you test this?

- 36 new parameterized matcher tests (`exclude-patterns.test.ts`).
- New real-git integration test: `.envrc` linked, `.env` copied, and `node_modules/` never created in a fresh worktree (regression for the bootstrap bug).
- New suspension tests: at-cap does nothing, multi-excess suspends oldest first.
- New `sweepTrash` test; existing worktree lifecycle (add/remove/prune) tests pass unchanged against the trash-based delete.
- Full suites green: `@posthog/git` 306 tests, `@posthog/workspace-server` 612 tests. Typecheck and Biome clean.
- Verified the new scan end-to-end against the real posthog repo via the built package: `.worktreelink -> []` (comments only), `.worktreeinclude -> [".env"]` in ~140ms.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/git/src/exclude-patterns.ts
+++ b/packages/git/src/exclude-patterns.ts
@@ -38,7 +38,15 @@ export function parseExcludePatterns(content: string): ExcludePattern[] {
       pattern = pattern.slice(1);
     }
 
-    patterns.push({ negated, dirOnly, regex: globToRegExp(pattern, anchored) });
+    // A single malformed pattern must not drop the whole exclude file: skip the
+    // offending line rather than letting a RegExp throw propagate out.
+    let regex: RegExp;
+    try {
+      regex = globToRegExp(pattern, anchored);
+    } catch {
+      continue;
+    }
+    patterns.push({ negated, dirOnly, regex });
   }
 
   return patterns;
@@ -86,7 +94,9 @@ function patternMatches(
 }
 
 function trimUnescapedTrailingSpaces(line: string): string {
-  return line.replace(/(?<!\\) +$/, "");
+  // Drop a trailing CR first so CRLF-terminated exclude files don't bake a \r
+  // into every pattern (which would make the compiled regex match nothing).
+  return line.replace(/\r$/, "").replace(/(?<!\\) +$/, "");
 }
 
 function globToRegExp(pattern: string, anchored: boolean): RegExp {
@@ -98,8 +108,19 @@ function globToRegExp(pattern: string, anchored: boolean): RegExp {
     if (char === "*") {
       if (pattern[i + 1] === "*") {
         if (pattern[i + 2] === "/") {
+          // Collapse a run of consecutive `**/` into one `(?:.*/)?`. They are
+          // semantically equivalent, and emitting one group per segment would
+          // stack overlapping backtracking `.*` groups — catastrophic on a
+          // slash-heavy path that fails the final literal (ReDoS).
           source += "(?:.*/)?";
           i += 3;
+          while (
+            pattern[i] === "*" &&
+            pattern[i + 1] === "*" &&
+            pattern[i + 2] === "/"
+          ) {
+            i += 3;
+          }
         } else {
           source += ".*";
           i += 2;
--- a/packages/git/src/exclude-patterns.ts
+++ b/packages/git/src/exclude-patterns.ts
@@ -1,0 +1,141 @@
+export interface ExcludePattern {
+  negated: boolean;
+  dirOnly: boolean;
+  regex: RegExp;
+}
+
+/**
+ * Parses gitignore-style pattern lines (comments, negation, dir-only trailing
+ * slash, root anchoring, `*`/`?`/`**`/`[...]` globs). Used to re-apply an
+ * exclude file's patterns in-process against paths git has already listed, so
+ * callers can avoid asking git to walk huge ignored trees.
+ */
+export function parseExcludePatterns(content: string): ExcludePattern[] {
+  const patterns: ExcludePattern[] = [];
+
+  for (const rawLine of content.split("\n")) {
+    const line = trimUnescapedTrailingSpaces(rawLine);
+    if (!line || line.startsWith("#")) continue;
+
+    let pattern = line;
+    let negated = false;
+    if (pattern.startsWith("!")) {
+      negated = true;
+      pattern = pattern.slice(1);
+    } else if (pattern.startsWith("\\!") || pattern.startsWith("\\#")) {
+      pattern = pattern.slice(1);
+    }
+
+    let dirOnly = false;
+    if (pattern.endsWith("/")) {
+      dirOnly = true;
+      pattern = pattern.slice(0, -1);
+    }
+    if (!pattern) continue;
+
+    const anchored = pattern.includes("/");
+    if (pattern.startsWith("/")) {
+      pattern = pattern.slice(1);
+    }
+
+    patterns.push({ negated, dirOnly, regex: globToRegExp(pattern, anchored) });
+  }
+
+  return patterns;
+}
+
+/**
+ * Whether a path matches the pattern list, last match wins (gitignore
+ * semantics). `entry` may carry a trailing slash to mark a directory, as in
+ * `git ls-files --directory` output. A pattern matching a parent directory
+ * matches everything beneath it.
+ */
+export function matchesExcludePatterns(
+  entry: string,
+  patterns: ExcludePattern[],
+): boolean {
+  const isDir = entry.endsWith("/");
+  const entryPath = isDir ? entry.slice(0, -1) : entry;
+
+  let matched = false;
+  for (const pattern of patterns) {
+    if (patternMatches(pattern, entryPath, isDir)) {
+      matched = !pattern.negated;
+    }
+  }
+  return matched;
+}
+
+function patternMatches(
+  pattern: ExcludePattern,
+  entryPath: string,
+  isDir: boolean,
+): boolean {
+  if ((isDir || !pattern.dirOnly) && pattern.regex.test(entryPath)) {
+    return true;
+  }
+
+  let separatorIndex = entryPath.indexOf("/");
+  while (separatorIndex !== -1) {
+    if (pattern.regex.test(entryPath.slice(0, separatorIndex))) {
+      return true;
+    }
+    separatorIndex = entryPath.indexOf("/", separatorIndex + 1);
+  }
+  return false;
+}
+
+function trimUnescapedTrailingSpaces(line: string): string {
+  return line.replace(/(?<!\\) +$/, "");
+}
+
+function globToRegExp(pattern: string, anchored: boolean): RegExp {
+  let source = anchored ? "^" : "^(?:.*/)?";
+  let i = 0;
+
+  while (i < pattern.length) {
+    const char = pattern[i];
+    if (char === "*") {
+      if (pattern[i + 1] === "*") {
+        if (pattern[i + 2] === "/") {
+          source += "(?:.*/)?";
+          i += 3;
+        } else {
+          source += ".*";
+          i += 2;
+        }
+      } else {
+        source += "[^/]*";
+        i += 1;
+      }
+    } else if (char === "?") {
+      source += "[^/]";
+      i += 1;
+    } else if (char === "[") {
+      const classEnd = pattern.indexOf("]", i + 2);
+      if (classEnd === -1) {
+        source += "\\[";
+        i += 1;
+      } else {
+        let charClass = pattern.slice(i + 1, classEnd);
+        if (charClass.startsWith("!")) {
+          charClass = `^${charClass.slice(1)}`;
+        }
+        source += `[${charClass}]`;
+        i = classEnd + 1;
+      }
+    } else if (char === "\\" && i + 1 < pattern.length) {
+      source += escapeRegExp(pattern[i + 1]);
+      i += 2;
+    } else {
+      source += escapeRegExp(char);
+      i += 1;
+    }
+  }
+
+  return new RegExp(`${source}$`);
+}
+
+function escapeRegExp(char: string): string {
+  return /[.*+?^${}()|[\]\\/]/.test(char) ? `\\${char}` : char;
+}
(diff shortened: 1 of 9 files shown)


pauldambra:
* **#3141** <a href="https://app.graphite.com/github/pr/PostHog/code/3141?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/3141?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/git/src/exclude-patterns.ts
```diff
@@ -0,0 +1,141 @@
...
+function globToRegExp(pattern: string, anchored: boolean): RegExp {
+  let source = anchored ? "^" : "^(?:.*/)?";
+  let i = 0;
+
+  while (i < pattern.length) {
+    const char = pattern[i];
+    if (char === "*") {
+      if (pattern[i + 1] === "*") {
+        if (pattern[i + 2] === "/") {
+          source += "(?:.*/)?";
```
pauldambra, line 115: > [!NOTE]

**[convergent: security-audit/redos + qa-team/data-integrity]** 🟠 HIGH

ReDoS in the glob→regex conversion. Consecutive `**/` segments each compile to an overlapping `(?:.*/)?` group, so a pattern like `**/**/…/NOMATCH` backtracks catastrophically against slash-heavy candidate paths. Both the `.worktreelink`/`.worktreeinclude` contents and the directory depth are repo-controlled, and `matchesExcludePatterns` runs synchronously on the single-threaded workspace-server — so a malicious/contributor repo can freeze worktree creation (and every git/fs op the server handles).

Reproduced against the built matcher: pattern `("**"×20)/NOMATCH` — depth 12 = 301ms, depth 14 = 2.7s, depth 16 = 22.8s, growing exponentially.

Fix: collapse consecutive `**/` runs at parse time (a no-op for gitignore semantics), and/or bound pattern complexity. Add a regression test asserting bounded match time.

packages/git/src/exclude-patterns.ts
```diff
@@ -0,0 +1,141 @@
...
+  while (separatorIndex !== -1) {
+    if (pattern.regex.test(entryPath.slice(0, separatorIndex))) {
+      return true;
+    }
+    separatorIndex = entryPath.indexOf("/", separatorIndex + 1);
+  }
+  return false;
+}
+
+function trimUnescapedTrailingSpaces(line: string): string {
```
pauldambra, line 96: > [!NOTE]

**[qa-team/data-integrity]** 🟠 HIGH

CRLF exclude files silently match nothing. `trimUnescapedTrailingSpaces` strips trailing spaces but not `\r`, so a `.worktreelink`/`.worktreeinclude` saved with Windows line endings compiles each pattern with a trailing `\r` (`^(?:.*/)?\.env\r$`) and never matches — every link/include silently no-ops on such a repo.

Verified: `parseExcludePatterns(".env\r\n")` then matching `.env` returns false.

Fix: strip a trailing `\r` (normalize CRLF) in the trim step; add a CRLF case to the parameterized tests.

packages/git/src/worktree.ts
```diff
@@ -670,6 +746,7 @@ function getIgnoredPathsFromExcludeFile(
         "--ignored",
         "--others",
         "--directory",
+        "--exclude-standard",
```
pauldambra, line 819: > [!NOTE]

**[convergent: paul + xp + qa-team/compatibility]** 🟡 MEDIUM

Switching to `--exclude-standard` means an exclude-file entry that lives *inside* a standard-ignored tree (e.g. `.worktreeinclude` lists `dist/keep.env` while `dist/` is gitignored) is no longer a candidate git lists, so it silently stops being copied. For the stray `node_modules/.pnpm/**/.env` this is the intended fix, but the behavior change is invisible. Add a test naming the "path inside a standard-ignored dir" case as *intentionally dropped* so a future maintainer doesn't "fix" it back into the slow full walk — plus a line in the PR body.

packages/workspace-server/src/services/workspace/workspace.ts
```diff
@@ -796,6 +794,13 @@ export class WorkspaceService extends TypedEventEmitter<WorkspaceServiceEvents>
       path: worktree.worktreePath,
     });
 
+    // Enforce the worktree cap after the new workspace exists rather than
+    // before: suspending the least-recent task checkpoints and deletes a
+    // potentially multi-gigabyte worktree, which must not block this creation.
+    this.suspensionService.suspendLeastRecentIfOverLimit().catch((error) => {
```
pauldambra, line 800: > [!NOTE]

Moving `suspendLeastRecentIfOverLimit()` off the create path is the right latency call, but it's now fire-and-forget and recomputes `active` each call. Two near-simultaneous creates can both read the same active list and pick the same oldest task, firing `autoSuspend` on it twice. Please confirm a duplicate suspend of one taskId is a safe no-op (`executeSuspend` guards on an existing suspension row, so it likely is) — otherwise the background sweep could checkpoint/delete a worktree already being suspended.

pauldambra (commented):
QA Swarm review — 4 perspectives (qa-team, paul-reviewer, xp-reviewer, security-audit). Two findings independently confirmed by re-running the built matcher (ReDoS + CRLF). See inline comments + summary.

pauldambra:
> [!NOTE]
>

## Verdict: ⚠️ REQUEST CHANGES <sub>(round 1 @ 2d9cf78)</sub>

Well-diagnosed, well-tested performance PR — but the newly added `exclude-patterns.ts` matcher has two independently-confirmed correctness/availability bugs that should be fixed before merge. Both were reproduced by re-running the built matcher, not just reasoned about.

### Key findings

**🟠 HIGH**
- **ReDoS in glob→regex** (`exclude-patterns.ts`) — consecutive `**/` compile to overlapping `(?:.*/)?` groups; `("**"×20)/NOMATCH` blows up exponentially (depth 16 = 22.8s) against repo-controlled patterns + paths, freezing the single-threaded workspace-server during worktree creation. Flagged by security-audit, confirmed by qa-team and again here.
- **CRLF exclude files match nothing** (`exclude-patterns.ts`) — trailing `\r` is not stripped, so a Windows-line-ending `.worktreelink`/`.worktreeinclude` bakes `\r` into every pattern and silently links/copies nothing. Verified empirically.

**🟡 MEDIUM**
- **Silent include regression for paths inside standard-ignored dirs** — a deliberate consequence of `--exclude-standard` (and the right fix for the node_modules stray), but invisible; wants a naming test + a PR-body note. Flagged independently by paul, xp, and qa-team/compatibility.

**🟢 LOW / ⚪ NIT**
- Fire-and-forget suspend can double-target the same oldest task under concurrent creates (paul + xp) — confirm `executeSuspend`'s existing-row guard makes it a no-op.
- `moveToTrash` swallows all errors into `null` with no log (xp) — a persistent rename failure silently degrades every delete to the slow in-lock path.
- Defensive: wrap per-line `globToRegExp` so one malformed pattern can't drop the whole exclude file (qa-team/data-integrity).
- Shared `.trash` swept once per folder at boot (paul); trash-name `Date.now()` collision + layout coupling (xp/paul); stale `active.length` in the suspend loop log (xp); test label "star glob does not cross directories" asserts a match (paul).
- `core.ignorecase` divergence vs git on macOS/Windows (qa-team/compatibility).

### Convergence

The **collapsed-dir include regression** was raised independently by paul, xp, and qa-team — highest-confidence non-blocking item. The **ReDoS** was raised by security-audit and corroborated by qa-team's own reproduction. Convergent findings carry higher confidence.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | HIGH — empirically verified CRLF total-failure and the collapsed-dir include regression; also flagged `core.ignorecase` divergence and trash collision. |
| 👤 paul | Approve with comments; the one answer he wants before shipping is a fallback/kill-switch for the hand-rolled matcher (product-wide blast radius). |
| 📐 xp | Strong make-it-fast pass; one substantive concern — fire-and-forget suspend drops the old serialization; rest LOW/NIT polish. |
| 🛡 security-audit | One real bug: ReDoS in the glob→regex conversion (Medium, availability, attacker-influenced repo). Rest of the diff is path-safe and argv-safe. |

---

packages/git/src/worktree.ts
```diff
@@ -552,6 +557,16 @@ export class WorktreeManager {
       }
     }
 
+    const trashedPath = await this.moveToTrash(resolvedWorktreePath);
+    if (trashedPath) {
+      await manager.executeWrite(this.mainRepoPath, (git) =>
```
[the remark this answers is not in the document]
pauldambra: Fixed in d47507ea — the rename-to-trash and `git worktree prune` now run together inside `executeWrite`, and the rename is rolled back if prune throws, so a prune failure leaves the worktree consistent (dir present, still registered) and falls through to the in-place remove. Only the slow recursive delete stays outside the lock.

packages/workspace-server/src/services/suspension/suspension.ts
```diff
@@ -149,23 +149,30 @@ export class SuspensionService extends TypedEventEmitter<SuspensionServiceEvents
...
-    this.log.info(
-      `Auto-suspending task ${oldest.taskId} (max: ${maxActive}, active: ${active.length})`,
-    );
-    await this.autoSuspend(oldest.taskId, "max_worktrees");
+    for (const workspace of oldestFirst.slice(0, excess)) {
+      this.log.info(
+        `Auto-suspending task ${workspace.taskId} (max: ${maxActive}, active: ${active.length})`,
+      );
+      await this.autoSuspend(workspace.taskId, "max_worktrees");
+    }
```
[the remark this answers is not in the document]
pauldambra: Already addressed in 629891d9 — the loop now logs the over-cap count once up front (`Auto-suspending N task(s) over the worktree cap`) rather than a per-iteration count, so there's no stale/repeated `active: N`.

packages/git/src/exclude-patterns.ts
pauldambra: Fixed in 629891d9 — consecutive `**/` runs are collapsed into one `(?:.*/)?` at compile time, so the pathological pattern now resolves in ~0ms (was 22.8s at depth 16). Regression test added asserting bounded match time; verified robust against separated `**/` shapes too.

packages/git/src/exclude-patterns.ts
pauldambra: Fixed in 629891d9 — a trailing `\r` is stripped before pattern compilation, so CRLF exclude files match correctly. CRLF case added to the parameterized tests.

packages/git/src/worktree.ts
pauldambra: Intended behavior (this is the node_modules bootstrap fix). Resolving as by-design; the PR body documents that entries inside standard-ignored trees are deliberately no longer copied. The real-git test asserts `node_modules/` is never created in a fresh worktree.

packages/workspace-server/src/services/workspace/workspace.ts
pauldambra: Fixed in 629891d9 — `suspendLeastRecentIfOverLimit` now serializes sweeps via a chained promise, so concurrent creates can't both pick the same oldest task. Each sweep re-reads the active set after the previous one commits.

charlesvien (approved):

PR #3144, opened 2026-07-04
fix(ui): stop deleted archived tasks from reappearing in sidebar

igennova:
## Summary

- Invalidate the workspaces cache when an archived task is deleted, so the sidebar no longer shows the task after delete.
- Consolidate restore/delete post-action refresh into one helper that invalidates workspaces, archive, and tasks together.

Fixes #3143 

## Problem

Deleting a task from the Archived tasks view removes its local `archives` and `workspaces` rows, but the UI only refreshed archive and tasks caches. Sidebar visibility depends on both `archivedTaskIds` and `workspaces`, so after delete the app briefly had:

- not archived (refreshed)
- still has workspace (stale)

That made the task reappear in the sidebar until workspaces refetched for another reason.

Restore already invalidated workspaces via `invalidateOnRestore`; delete used a separate helper that did not.

## Solution

Include `WORKSPACE_QUERY_KEY` invalidation in `invalidateArchiveQueries`, which runs after archived-task delete (and is shared with restore).

## Video

https://github.com/user-attachments/assets/9a39b629-7280-4f24-8898-2b9f445dcdd9

### DIFF
--- a/packages/ui/src/features/archive/useUnarchiveTask.ts
+++ b/packages/ui/src/features/archive/useUnarchiveTask.ts
@@ -34,15 +34,16 @@ export function useUnarchiveTask(): UseUnarchiveTask {
 
   const invalidateArchiveQueries = useCallback(async () => {
     await Promise.all([
+      queryClient.invalidateQueries({ queryKey: WORKSPACE_QUERY_KEY }),
       queryClient.invalidateQueries(trpc.archive.pathFilter()),
       queryClient.refetchQueries({ queryKey: ["tasks"] }),
     ]);
   }, [queryClient, trpc]);
 
-  const invalidateOnRestore = useCallback(async () => {
-    await queryClient.invalidateQueries({ queryKey: WORKSPACE_QUERY_KEY });
-    await invalidateArchiveQueries();
-  }, [queryClient, invalidateArchiveQueries]);
+  const invalidateOnRestore = useCallback(
+    async () => invalidateArchiveQueries(),
+    [invalidateArchiveQueries],
+  );
 
   const restore = useCallback(
     async (
--- a/packages/ui/src/features/archive/useUnarchiveTask.ts
+++ b/packages/ui/src/features/archive/useUnarchiveTask.ts
@@ -40,11 +40,6 @@ export function useUnarchiveTask(): UseUnarchiveTask {
     ]);
   }, [queryClient, trpc]);
 
-  const invalidateOnRestore = useCallback(
-    async () => invalidateArchiveQueries(),
-    [invalidateArchiveQueries],
-  );
-
   const restore = useCallback(
     async (
       taskId: string,
@@ -53,11 +48,11 @@ export function useUnarchiveTask(): UseUnarchiveTask {
     ) => {
       const outcome = await controller.restore(taskId, hasTask, options);
       if (outcome.kind === "restored") {
-        await invalidateOnRestore();
+        await invalidateArchiveQueries();
       }
       return outcome;
     },
-    [controller, invalidateOnRestore],
+    [controller, invalidateArchiveQueries],
   );
 
   const remove = useCallback(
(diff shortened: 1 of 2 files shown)


packages/ui/src/features/archive/useUnarchiveTask.ts
```diff
@@ -34,16 +34,12 @@ export function useUnarchiveTask(): UseUnarchiveTask {
 
   const invalidateArchiveQueries = useCallback(async () => {
```
pauldambra: > [!NOTE]

**[convergent: qa-team + paul + xp]** 🟢 LOW

`invalidateArchiveQueries` now also invalidates `WORKSPACE_QUERY_KEY` and refetches `["tasks"]`, not just archive queries — the name undersells its scope. A future reader deciding whether to reuse or extend it will assume it's archive-only. Consider a name like `invalidateAfterArchiveMutation` or `invalidateTaskListCaches`. Not blocking.

packages/ui/src/features/archive/useUnarchiveTask.ts
```diff
@@ -34,16 +34,12 @@ export function useUnarchiveTask(): UseUnarchiveTask {
...
     await Promise.all([
+      queryClient.invalidateQueries({ queryKey: WORKSPACE_QUERY_KEY }),
```
pauldambra, line 37: > [!NOTE]

**[convergent: qa-team/reliability + paul]** 🟡 MEDIUM

The bug this PR fixes existed because the restore and delete paths' invalidation sets silently drifted apart (delete never invalidated `WORKSPACE_QUERY_KEY`, so the sidebar kept showing deleted archived tasks). Nothing added here prevents that drift from recurring. Worth a small regression test asserting the delete/remove outcome invalidates `WORKSPACE_QUERY_KEY` — `workspaceQueryKey.test.ts` and `useWorkspaceMutations.test.tsx` already show the pattern (real `QueryClient` + tRPC options proxy, assert partial-key match / invalidation spy).

packages/ui/src/features/archive/useUnarchiveTask.ts
```diff
@@ -78,7 +74,7 @@ export function useUnarchiveTask(): UseUnarchiveTask {
         hasTask,
       );
       if (outcome.kind === "restore" && outcome.outcome.kind === "restored") {
```
pauldambra, line 76: > [!NOTE]

Both branches of this outcome check now call the identical `invalidateArchiveQueries`; the `if`/`else if` could collapse into a single predicate (`(restore && restored) || (delete && deleted)`). Fine to keep as-is if you prefer the explicit outcome-kind documentation.

pauldambra (commented):

pauldambra:
> [!NOTE]
>

## Verdict: 💬 APPROVE WITH NITS <sub>(round 1 @ 6291f0a)</sub>

The fix is correct and well-targeted: the delete path never invalidated the sidebar's workspace query, so deleted archived tasks kept reappearing. Folding that invalidation into the shared `invalidateArchiveQueries` (and removing the now-redundant `invalidateOnRestore` wrapper) fixes it cleanly for both restore and delete. A couple of minor, non-blocking suggestions below.

### Key findings

- 🟡 MEDIUM — the restore/delete invalidation sets drifted apart once before (that's the bug this PR fixes); no regression test pins it against drifting apart again
- 🟢 LOW — `invalidateArchiveQueries` now also touches the workspace cache and refetches tasks; the name undersells its scope
- ⚪ NIT — `runContextMenuAction`'s two identical-outcome branches could collapse into one predicate

### Convergence

- **Naming** (`invalidateArchiveQueries` scope) — independently flagged by qa-team, paul, and xp
- **Missing regression test** — independently flagged by qa-team and paul

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Correct fix, verified against PR head; suggests a regression test and a scope-honest rename |
| 👤 paul | The invalidation fix is the right call; flags the same naming and test-coverage points |
| 📐 xp | Good OnceAndOnlyOnce move removing the wrapper; same naming nit |
| 🛡 security-audit | Pure frontend cache-invalidation change, no user-controlled data flow — zero findings |

---

igennova:
@pauldambra Ready for review
Thanks

charlesvien (approved):

Issue #1076, opened 2026-02-27
Try out `rtk-ai/rtk` to reduce LLM token consumption


jonathanlab:
https://github.com/rtk-ai/rtk

charlesvien:
It's not bad, I submitted a PR to add graphite support this weekend for some fun in Rust https://github.com/rtk-ai/rtk/pull/290

PR #3145, opened 2026-07-04
feat(agent): bundle rtk with the agent package

pauldambra:
## Problem

Follow-up to #3096, which added RTK command-output compression to the agent but only activated when `rtk` happened to be on `PATH`. That leaves cloud runs (and any machine without rtk installed) getting no compression, and different environments potentially running different rtk versions. This packages a pinned rtk with the agent so the token savings apply consistently everywhere — including cloud runs — as discussed in the thread on #3096.

Refs #1076

## Changes

- **`packages/agent/build/rtk-binary.mjs`** — pins the rtk version and downloads the matching GitHub release for the build target, caching it under `node_modules/.cache`. The agent's `tsup` build copies it into `dist/rtk/`. Best-effort: an offline build warns and skips rather than failing, and the runtime then falls back to `PATH`.
- **Cloud runs** — the `agent-server` bin defaults `POSTHOG_RTK` to the bundled binary (its `dist/rtk/` sibling) when the env var is unset.
- **Desktop** — a vite plugin stages `dist/rtk/` into `.vite/build/rtk/` (asar-unpacked), and the host points `POSTHOG_RTK` at it, mirroring how the Claude CLI binary is bundled. Single source of truth keeps the version consistent across hosts.

The runtime resolver added in #3096 is unchanged: it still honors an explicit `POSTHOG_RTK` path or a `0`/`false` opt-out, and no bundle means it falls back to `PATH` (zero behavior change).

## How did you test this?

- `pnpm --filter @posthog/agent build` — confirmed it downloads and vendors `dist/rtk/rtk` (`rtk 0.43.0`), executable, at the path the cloud bin resolves (`dist/server/../rtk/rtk`).
- Verified the target mapping resolves the right release triple for linux/darwin/win x64+arm64 (win-arm64 has no asset and skips cleanly).
- `pnpm --filter @posthog/agent typecheck`, `@posthog/workspace-server typecheck`, `code typecheck` — all clean.
- `pnpm --filter @posthog/agent test rtk` (38 pass) and the `auth-adapter` suite, including a new test covering the `POSTHOG_RTK` wiring (7 pass).
- Biome clean on all changed files; host-boundary check reports no new violations.

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1783166413913899?thread_ts=1783166413.913899&cid=C09G8Q32R6F)*


packages/workspace-server/src/services/agent/auth-adapter.ts
```diff
@@ -153,6 +156,9 @@ export class AgentAuthAdapter {
 
     process.env.LLM_GATEWAY_URL = proxyUrl;
     process.env.CLAUDE_CODE_EXECUTABLE = claudeCliPath;
+    // Only set when the bundle is present; leaving it unset lets the agent's
+    // resolver fall back to an rtk on PATH.
+    if (rtkPath) process.env.POSTHOG_RTK = rtkPath;
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]
> 🤖 Automated triage by **PR Shepherd** — not written by a human

Good catch — fixed in c016f358. `configureProcessEnv` now guards with `if (rtkPath && !process.env.POSTHOG_RTK)`, so an explicit operator override (custom path, or `0`/`false` opt-out) wins over the bundled default, matching the cloud `bin.ts` behavior. Added a test covering the `POSTHOG_RTK=0` opt-out case.

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,191 @@
...
+export function rtkReleaseTarget(
+  platform = targetPlatform(),
+  arch = targetArch(),
+) {
+  return RTK_TARGETS[platform]?.[arch];
+}
+
+function rtkAssetUrl(target) {
+  const ext = target.includes("windows") ? "zip" : "tar.gz";
+  return `https://github.com/rtk-ai/rtk/releases/download/v${RTK_VERSION}/rtk-${target}.${ext}`;
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]
> 🤖 Automated triage by **PR Shepherd** — not written by a human

Verified as a false positive: the target string `x86_64-pc-windows-msvc` **does** contain the substring `windows` (`pc-windows-msvc`), so `target.includes("windows")` returns `true` and the Windows asset is correctly handled as `.zip` in both the download URL and the extractor. Confirmed: `node -e 'console.log("x86_64-pc-windows-msvc".includes("windows"))'` → `true`. No change needed.

pauldambra:
> [!NOTE]
> 🤖 Automated comment by **QA Swarm** (via PR Shepherd) — not written by a human
>

## Verdict: 💬 APPROVE WITH NITS <sub>(final @ bc53fab)</sub>

Nine review rounds across the PR's evolution (rtk bundling → gauge telemetry → desktop reporter → env scrub → runnability probe); every finding is fixed and verified except the one cross-repo item below. CI green (20 pass / 0 fail as of b9c14cc; latest push re-running). Awaiting human sign-off (requested: @joshsny) — stamphog's gates require it for the auth-adapter touch and third-party binary pipeline regardless of review state.

### Open item (cross-repo, unresolvable here)

- 🟡 **Cloud consumer for `_posthog/rtk_savings`** — verified against posthog master: the tasks backend stores the notification but never captures it as an analytics event. Follow-up in posthog: map it to the `"RTK savings gauge"` event (`RtkSavingsGaugeProperties` in @posthog/shared), consumed as a per-counter diff, never a sum.

### Highlights of what the loop fixed

- 🟠 EXDEV cross-device rename introduced by an earlier fix — extraction now lives wholly in a per-process temp dir beside the cache.
- 🟡 `counter_id` re-keyed run→task after spotting warm snapshot-resumes preserve the rtk tally across run_ids.
- 🟡 Present-but-unrunnable binary hazard — the rewrite hook now probes `rtk --version` once per session and downgrades to no-rewrite on failure.
- 🔒 `rtk gain` env scrubbed to a platform-essentials allowlist (no more tokens/keys reaching the third-party binary); telemetry reshaped to gauge semantics (cumulative counter reads, diff-not-sum contract); desktop reports daily via the typed analytics event map.

### Supply-chain posture (discussed, tracked)

Hash-pinned download closes post-pin tampering; agreed follow-up is mirroring rtk release artifacts into PostHog-controlled storage so builds stop fetching third-party infra (per-platform npm packages as the longer-term option).

<details>
<summary>Previous rounds (8)</summary>

r1 @ 2168f7e — nits on retry shape/teardown (fixed). r2 @ 4e5352d — 2 MED + 5 LOW + 3 NIT; 5 fixed, 3 deferred. r3 @ 20cb39c — convergent HIGH (EXDEV) in r2's own fix; fixed. r4 @ 258e019 — polish only; fixed. r5 @ 4763eb5 — comment accuracy; fixed. r6 @ 5fe748e — MED counter_id lifetime + recipe gaps; fixed. r7 @ 7b77dfd — desktop reporter delta: 2 LOW + 3 NIT (bindings map, sync machineId, init catch); fixed. r8 @ b9c14cc — env scrub + polish delta: clean.

</details>

---

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,150 @@
...
+  }
+}
+
+async function extractArchive(archivePath, destDir, target) {
+  if (target.includes("windows")) {
+    const entries = unzipSync(readFileSync(archivePath));
+    const bin = rtkBinName("win32");
+    const data = entries[bin] ?? entries[`rtk-${target}/${bin}`];
+    if (!data) throw new Error(`rtk binary not found in ${archivePath}`);
+    writeFileSync(join(destDir, bin), data);
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]
> 🤖 Automated triage by **PR Shepherd** — not written by a human

This is already addressed (the comment predates the fix). `rtk-binary.mjs` now pins the SHA-256 of each release archive in `RTK_SHA256` and calls `verifyChecksum()` on the downloaded archive **before** extraction, deleting the archive and failing the build on mismatch — so a swapped or truncated asset never gets cached or bundled. See commit `0146ce3`.

packages/agent/src/server/bin.ts
```diff
@@ -1,10 +1,28 @@
...
+ * on rtk being on the sandbox PATH. Only default when POSTHOG_RTK is unset —
+ * any explicit value wins (a path override, or `0`/`false` to opt out, both
+ * interpreted downstream by resolveRtkPrefix). Stay silent when the bundle is
+ * absent so the runtime resolver falls back to PATH.
+ */
+function applyBundledRtkDefault(): void {
+  if (process.env.POSTHOG_RTK) return;
+  const binName = process.platform === "win32" ? "rtk.exe" : "rtk";
+  const bundled = resolve(__dirname, "..", "rtk", binName);
+  if (existsSync(bundled)) process.env.POSTHOG_RTK = bundled;
```
pauldambra, line 23: > [!NOTE]

**[qa-team/compatibility]** 🟡 MEDIUM

`applyBundledRtkDefault` (and the desktop equivalent in `auth-adapter.ts`) adopts the bundled rtk purely on `existsSync`, with no check that the binary can actually execute on this host. A present-but-unrunnable binary — e.g. the aarch64-linux-gnu build on a musl-based arm64 image, a glibc mismatch, or a mac binary whose ad-hoc codesign step failed (`copyRtkBinary` warns and continues on signing failure) — passes `resolveRtkPrefix`'s `statSync` check, and the PreToolUse hook then rewrites every eligible `grep`/`find`/`ls`/`git` command to `rtk ...`, which fails with exec-format/loader errors and degrades the whole run. That is strictly worse than having no rtk (where the resolver falls back to PATH or disables). Suggestion: probe once with a cheap `rtk --version` (execFile, short timeout) before setting POSTHOG_RTK or enabling the rewrite hook.

packages/agent/src/acp-extensions.ts
```diff
@@ -78,6 +78,9 @@ export const POSTHOG_NOTIFICATIONS = {
 
   /** Permission request resolved, persisted so a reconnecting client can tell it is no longer pending */
   PERMISSION_RESOLVED: "_posthog/permission_resolved",
+
+  /** RTK output-compression token savings tallied at the end of a run */
+  RTK_SAVINGS: "_posthog/rtk_savings",
```
pauldambra, line 83: > [!NOTE]

**[qa-team/data-integrity]** 🟡 MEDIUM

The new `_posthog/rtk_savings` notification is enqueued into the task-run event stream, but nothing in this repo shows the cloud-side ingestion consumer handling that method. If the consumer drops or rejects unknown notification methods, the telemetry — the entire point of the emit commit — is silently discarded with no alert. Verify the PostHog-side event-stream consumer maps `_posthog/rtk_savings` params into event properties, and consider a smoke check (query for the event after the first cloud deploy).

packages/agent/src/server/agent-server.ts
```diff
@@ -3373,6 +3381,49 @@ ${signedCommitInstructions}
...
+   * Emit RTK's output-compression token savings for this run as a telemetry
+   * event, so PostHog can report how much context RTK saved. Best-effort and
+   * cloud-only (no-op when the event stream isn't configured). Fires at most
+   * once per run: it's called from both terminal seams (the error path stops
+   * the stream in signalTaskComplete, before cleanupSession runs) and must land
+   * before the stream is stopped, since enqueue is a no-op once stopped.
+   */
+  private async emitRtkSavings(): Promise<void> {
+    if (!this.eventStreamSender || this.rtkSavingsEmitted) return;
+    this.rtkSavingsEmitted = true;
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + paul + xp]** 🟢 LOW

three reviewers converged here: `rtkSavingsEmitted` is set to true *before* the await, so a transient `rtk gain` failure at the first terminal seam permanently forfeits the telemetry even though the second seam still has the stream open. if that's the intended at-most-one-attempt tradeoff (avoids retry storms on the terminal path — probably right), the flag name undersells it: it really means "attempted", not "emitted". either rename to `rtkSavingsAttempted` or add a one-liner saying a failed attempt still counts, so a future reader doesn't move the flag below the try expecting a retry. not blocking.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete (round for commit 4e5352d6f). See inline comments; summary comment updated.

packages/workspace-server/src/services/agent/agent.ts
```diff
@@ -413,6 +413,17 @@ export class AgentService extends TypedEventEmitter<AgentServiceEvents> {
...
   }
 
+  /**
+   * Bundled rtk binary for RTK output compression, or undefined when it wasn't
+   * vendored for this platform (the agent then falls back to rtk on PATH). Keep
+   * in sync with the destDir in apps/code/vite-main-plugins.mts (copyRtkBinary).
+   */
+  private getRtkPath(): string | undefined {
+    const binary = process.platform === "win32" ? "rtk.exe" : "rtk";
+    const rtkPath = this.bundledResources.resolve(`.vite/build/rtk/${binary}`);
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/generalist-a]** 🟡 MEDIUM

this is the coupling that'd bite us — `getRtkPath()` hardcodes `.vite/build/rtk/${binary}` and the only thing keeping it aligned with `copyRtkBinary`'s destDir in `apps/code/vite-main-plugins.mts` is a "keep in sync" comment. if someone renames that stage dir the desktop app silently falls back to PATH with no error. same story for the `rtk.exe`/`rtk` win32 name now spelled out in three places (here, `bin.ts`, `build/rtk-binary.mjs`). i wonder if the stage path + bin name could be a single shared const both sides import, so it's OnceAndOnlyOnce instead of a comment we hope people read. not blocking for this PR.

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,200 @@
...
+    );
+    await downloadFile(url, archivePath);
+    try {
+      await verifyChecksum(archivePath, target);
+    } catch (error) {
+      // Don't leave an unverified archive on disk to be reused.
+      rmSync(archivePath, { force: true });
+      throw error;
+    }
+    await extractArchive(archivePath, cacheDir, target);
```
pauldambra: > [!NOTE]

Extraction into the cache directory is not atomic: `extractArchive` writes `rtk` directly into `cacheDir`, and a build process killed mid-extract leaves a partial binary that `existsSync(cachedBinary)` accepts on every future build — the corrupt binary is then bundled indefinitely (the archive checksum was verified, but the extracted artifact never is). Extract into a temp dir and rename into place, or re-verify a recorded hash of the extracted binary.

packages/agent/src/server/agent-server.ts
```diff
@@ -2838,6 +2840,11 @@ ${signedCommitInstructions}
       error: errorMessage ?? "Agent error",
     });
 
+    // Emit before the finally below stops the stream — failed runs still used
+    // RTK-compressed commands and must be counted. cleanupSession's emit runs
+    // after this stop on the error path, so it can't rely on that one.
+    await this.emitRtkSavings();
```
pauldambra: > [!NOTE]

**[qa-team/performance]** 🟢 LOW

On the error path this `await this.emitRtkSavings()` runs `rtk gain` (execFile, up to 5s timeout) *before* the `updateTaskRun` status write, so every failed run's terminal "failed" status can be delayed by up to 5 seconds. The emit only needs to land before the stream `stop()` in the finally block, not before the status API call — reorder it after `updateTaskRun` (or run them concurrently) to keep failure visibility prompt.

packages/agent/src/server/agent-server.test.ts
```diff
@@ -496,19 +501,51 @@ describe("AgentServer HTTP Mode", () => {
...
+            totalCommands: 4,
+            inputTokens: 1000,
+            outputTokens: 350,
+            tokensSaved: 650,
+            avgSavingsPct: 65,
+          }),
+        }),
+      );
+
+      const session = testServer.session;
```
pauldambra, line 516: > [!NOTE]

nice test — it names the real hazard and pins the emit-before-stop ordering with `invocationCallOrder`. two small readability notes from two reviewers: (1) the `testServer.session = session` re-arm reads as a puzzle until you trace that the first cleanup nulled `this.session`; a one-liner like "re-arm session so the second terminal cleanup actually runs its body" would save the next reader the trace. (2) the once-only assertion is only meaningful because the stubbed `stop` doesn't gate `enqueue` — the `rtkSavingsEmitted` guard is genuinely the thing under test, worth saying so. ship as you see fit.

apps/code/vite-main-plugins.mts
```diff
@@ -218,6 +221,65 @@ export function copyClaudeExecutable(): Plugin {
...
+      }
+
+      if (isStagedCopyCurrent(source, destBinary)) {
+        rtkCopied = true;
+        return;
+      }
+
+      copyFileSync(source, destBinary);
+      if (targetPlatform() !== "win32") {
+        execSync(`chmod +x "${destBinary}"`);
```
pauldambra, line 266: > [!NOTE]

**[qa-team/security]** ⚪ NIT

`execSync` with an interpolated path for the chmod (and the xattr/codesign lines) puts a filesystem path into a shell string; a path containing `$`, backticks, or quotes would break or inject. It mirrors the existing `copyClaudeExecutable` pattern so it's consistent, but `chmodSync(destBinary, 0o755)` (as `rtk-binary.mjs` already does) avoids the shell entirely, and `execFileSync("xattr", ["-cr", destBinary])` would do the same for the mac steps.

packages/agent/src/server/rtk-savings.ts
```diff
@@ -0,0 +1,86 @@
...
+    avgSavingsPct: toFiniteNumber(summary.avg_savings_pct),
+  };
+}
+
+async function defaultRunGain(
+  binary: string,
+  env: NodeJS.ProcessEnv,
+): Promise<string> {
+  const { stdout } = await execFileAsync(binary, ["gain", "--format", "json"], {
+    timeout: 5_000,
```
pauldambra, line 64: > [!NOTE]

**[qa-team/performance]** ⚪ NIT

`execFileAsync` uses the default 1MB `maxBuffer`. `rtk gain --format json` includes the full `daily` array, which grows on long-lived hosts; documented cloud-only (fresh DB per run) so fine today, but if this resolver is ever reused on desktop the exec starts throwing ENOBUFS and silently returns null via the catch. An explicit `maxBuffer` makes the bound intentional.

packages/agent/src/server/agent-server.test.ts
```diff
@@ -405,6 +405,10 @@ describe("AgentServer HTTP Mode", () => {
       mode: "interactive",
       taskId: "test-task-id",
       runId: "test-run-id",
+      // Never probe the machine's real rtk: with the default resolver these
+      // tests would emit real savings wherever rtk is installed and pass
+      // everywhere else.
+      resolveRtkSavings: async () => null,
```
pauldambra, line 411: > [!NOTE]

**[convergent: qa-team/generalist-b + xp]** ⚪ NIT

two adjacent notes on the `createServer` consolidation (which is textbook OnceAndOnlyOnce — nice): (1) `question-relay.test.ts` and `agent-server.configure-environment.test.ts` still construct `AgentServer` directly without this null resolver; they're safe today only because they never set `eventIngestToken` so `emitRtkSavings` no-ops — an implicit invariant nobody enforces. A shared test factory would close it. (2) routing the `broadcastTurnComplete` tests through `createServer` now assigns the outer `server`, so `afterEach` stops servers that previously went un-stopped — almost certainly more correct, just a silent teardown-coverage change worth being aware of.

packages/agent/src/server/agent-server.ts
pauldambra: > [!NOTE]

Renamed `rtkSavingsEmitted` -> `rtkSavingsAttempted` and added a one-liner clarifying that the flag is set before the await intentionally (a failed attempt still counts). Also updated the test comment to reference the new name. Fixed in 1fced7a8d.

packages/agent/build/rtk-binary.mjs
pauldambra: > [!NOTE]

Made extraction atomic: extract to `mkdtempSync` temp dir, verify the binary exists there, then `renameSync` into `cacheDir`. The temp dir is cleaned up in a `finally` block. Fixed in 1fced7a8d.

packages/agent/src/server/agent-server.ts
pauldambra: > [!NOTE]

Moved `emitRtkSavings` into the `finally` block after `updateTaskRun`, so the "failed" status lands before the up-to-5s rtk gain probe. Fixed in 1fced7a8d.

packages/agent/src/server/agent-server.test.ts
pauldambra: > [!NOTE]

Added both readability comments: the re-arm reason ("first cleanup nulled this.session") and the once-only assertion note ("rtkSavingsAttempted is the actual guard under test"). Fixed in 1fced7a8d.

packages/agent/src/server/rtk-savings.ts
pauldambra: > [!NOTE]

Added `maxBuffer: 10 * 1024 * 1024` with a comment noting the bound is intentional for potential desktop reuse. Fixed in 1fced7a8d.

apps/code/vite-main-plugins.mts
pauldambra: > [!NOTE]

Valid point on the shell-injection surface. Keeping as-is for now since it mirrors the existing `copyClaudeExecutable` pattern in the same file — changing only the new lines would create an inconsistency. Tracking as a follow-up to update both together. Resolving as NIT.

packages/agent/src/server/agent-server.test.ts
pauldambra: > [!NOTE]

Both notes acknowledged. The implicit-invariant risk in question-relay and configure-environment tests is real but a follow-up item (scope change). The silent teardown-coverage improvement from routing broadcastTurnComplete through createServer is intentional. Resolving as NIT.

packages/agent/src/server/bin.ts
pauldambra: > [!NOTE]

Design decision deferred to author — adding an exec probe changes the adoption behaviour meaningfully (adds latency, introduces new failure mode), and there are multiple reasonable approaches (probe vs trust existsSync, timeout length, error handling). Not auto-fixed.

packages/agent/src/acp-extensions.ts
pauldambra: > [!NOTE]

Deferred to author — verifying the cloud-side consumer for `_posthog/rtk_savings` requires checking another repo and is a stop-and-ask decision. Not auto-actioned.

packages/workspace-server/src/services/agent/agent.ts
pauldambra: > [!NOTE]

Deferred to author — consolidating the stage path + binary name into a shared const touches workspace-server, apps/code, and packages/agent/build. The change is correct in principle but spans multiple packages; not auto-applied.

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,212 @@
...
+    try {
+      await verifyChecksum(archivePath, target);
+    } catch (error) {
+      // Don't leave an unverified archive on disk to be reused.
+      rmSync(archivePath, { force: true });
+      throw error;
+    }
+    // Extract into a temp dir first so a killed build cannot leave a partial
+    // binary in cacheDir that existsSync accepts on the next run.
+    const tmpDir = mkdtempSync(join(tmpdir(), "posthog-rtk-"));
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + qa-team/generalist-b + security-audit/supply-chain]** 🟠 HIGH

The atomic-extraction temp dir is created in `os.tmpdir()`, but `renameSync(extractedBinary, cachedBinary)` targets `node_modules/.cache` — a different filesystem on many hosts (tmpfs `/tmp` on Linux, Docker overlay mounts, cross-drive Windows). `rename(2)` does not copy across devices; it throws EXDEV. Because bundling is best-effort, the caller converts that into a silent PATH fallback — deterministically defeating the pinning and checksum verification this PR exists to provide, on exactly the containerized build hosts it targets, on every cold-cache build. Fix: `mkdtempSync` beside the cache dir (e.g. `mkdtempSync(join(dirname(cacheDir), ".tmp-"))` — the parent already exists after `mkdirSync(cacheDir, {recursive:true})`), which guarantees same-filesystem rename; or catch EXDEV and fall back to `copyFileSync` + cleanup.

packages/agent/src/server/agent-server.test.ts
```diff
@@ -496,19 +501,55 @@ describe("AgentServer HTTP Mode", () => {
...
+        }),
+      );
+      expect(
+        testServer.eventStreamSender.enqueue.mock.invocationCallOrder[0],
+      ).toBeLessThan(
+        testServer.eventStreamSender.stop.mock.invocationCallOrder[0],
+      );
+    });
+
     it("writes terminal failure status before completing event ingest", async () => {
```
pauldambra, line 554: > [!NOTE]

**[qa-team/generalist-a]** 🟡 MEDIUM

The emit was moved into `signalTaskComplete`'s finally, but no test asserts the rtk_savings enqueue lands before `stop()` on that error path: this ordering test uses `createServer`'s default `resolveRtkSavings: async () => null`, so the emit enqueues nothing and the reorder is only exercised as "doesn't break enqueue/update/stop ordering". A future refactor placing the emit after `stop()` in `signalTaskComplete` would pass the suite (the emit-before-stop assertion exists only on the `cleanupSession` seam). Add a savings-returning stub to this test and assert the rtk enqueue's `invocationCallOrder` precedes `stop`'s.

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,212 @@
...
+    try {
+      await extractArchive(archivePath, tmpDir, target);
+      rmSync(archivePath, { force: true });
+      const extractedBinary = join(tmpDir, binName);
+      if (!existsSync(extractedBinary)) {
+        throw new Error(
+          `rtk binary missing after extraction: ${cachedBinary} — the release archive layout for ${target} may have changed`,
```
pauldambra: > [!NOTE]

This "missing after extraction" error interpolates `cachedBinary`, but the existence check now runs against `extractedBinary` in the temp dir — when it fires, the message points a debugger at a path that was never inspected. Interpolate `extractedBinary` (or both).

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,212 @@
...
+        );
+      }
+      renameSync(extractedBinary, cachedBinary);
```
pauldambra: > [!NOTE]

Two concurrent cold-cache builds can both pass the `!existsSync(cachedBinary)` gate; on Windows `renameSync` throws when the destination exists, so the loser now fails where the previous in-place write overwrote harmlessly. Caught upstream and degrades to PATH fallback, so impact is a spurious warn + unbundled rtk for that build — acceptable, but the hardening narrowed rather than closed the race. A `try { renameSync } catch { if (!existsSync(cachedBinary)) throw }` would close it.

packages/agent/src/server/rtk-savings.ts
```diff
@@ -0,0 +1,89 @@
...
+    // The daily array grows on long-lived hosts; cap the buffer explicitly so
+    // ENOBUFS never silently swallows savings on desktop reuse.
```
pauldambra: > [!NOTE]

**[qa-team/generalist-a]** ⚪ NIT

Node's maxBuffer overrun surfaces as `ERR_CHILD_PROCESS_STDIO_MAXBUFFER` ("stdout maxBuffer length exceeded"), not errno `ENOBUFS`. The cap is right and the failure was indeed silently swallowed by the catch — just name the real error so the comment doesn't mislead a future grep.

packages/agent/src/server/agent-server.ts
```diff
@@ -3373,6 +3381,51 @@ ${signedCommitInstructions}
...
+   * event, so PostHog can report how much context RTK saved. Best-effort and
+   * cloud-only (no-op when the event stream isn't configured). Fires at most
+   * once per run: it's called from both terminal seams (the error path stops
+   * the stream in signalTaskComplete, before cleanupSession runs) and must land
+   * before the stream is stopped, since enqueue is a no-op once stopped.
+   */
+  private async emitRtkSavings(): Promise<void> {
+    if (!this.eventStreamSender || this.rtkSavingsAttempted) return;
+    // Set before the await: a failed attempt still counts — prevents retry
+    // storms on the terminal path.  The flag name reflects this intent.
```
pauldambra: > [!NOTE]

**[qa-team/generalist-a]** ⚪ NIT

Double space in the new comment ("terminal path.  The flag"). Verified separately: the `rtkSavingsAttempted` rename left no stale references, and the `targetPlatform()` hoist is behavior-identical.

pauldambra (commented):
> [!NOTE]

QA Swarm round 3 (delta review of 1fced7a8d + 20cb39ca3). One convergent HIGH: the round-2 atomic-extraction fix introduced a cross-device rename (EXDEV) that silently defeats rtk bundling on tmpfs/Docker/cross-drive hosts. See inline comments.

packages/agent/build/rtk-binary.mjs
pauldambra: > [!NOTE]

Fixed in 1bcf3c5c5: `mkdtempSync` now creates the temp dir beside `cacheDir` (via `dirname(cacheDir)`) rather than in `os.tmpdir()`, so the `renameSync` stays on the same filesystem and never throws EXDEV in Docker/Linux containers.

packages/agent/src/server/agent-server.test.ts
pauldambra: > [!NOTE]

Fixed in 1bcf3c5c5: the test now passes a savings-returning `resolveRtkSavings` override so the rtk_savings enqueue fires on the `signalTaskComplete` error path. The order assertion is updated to `["enqueue", "update", "enqueue", "stop"]` and a explicit `invocationCallOrder` check asserts the last enqueue (rtk_savings) precedes `stop()`.

packages/agent/build/rtk-binary.mjs
pauldambra: > [!NOTE]

Fixed in 1bcf3c5c5: error message now interpolates `extractedBinary` (the path that was checked) instead of `cachedBinary` (the destination that was never inspected).

packages/agent/build/rtk-binary.mjs
pauldambra: > [!NOTE]

Fixed in 1bcf3c5c5: added `try { renameSync } catch (renameError) { if (!existsSync(cachedBinary)) throw renameError; }` to absorb the Windows destination-exists throw when the winner already placed the file.

packages/agent/src/server/rtk-savings.ts
pauldambra: > [!NOTE]

Fixed in 1bcf3c5c5: updated the comment to say `ERR_CHILD_PROCESS_STDIO_MAXBUFFER` (the real Node error code) instead of `ENOBUFS`.

packages/agent/src/server/agent-server.ts
pauldambra: > [!NOTE]

Fixed in 1bcf3c5c5: removed the double space in the comment.

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,222 @@
...
+  createWriteStream,
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  renameSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
```
pauldambra: > [!NOTE]

The temp-dir move left `import { tmpdir } from "node:os"` unused — Biome flags it. (The move itself verified correct: `mkdirSync(cacheDir, {recursive:true})` precedes `mkdtempSync`, so the parent exists, and same-filesystem rename eliminates EXDEV.)

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,222 @@
...
+    try {
+      await extractArchive(archivePath, tmpDir, target);
+      rmSync(archivePath, { force: true });
+      const extractedBinary = join(tmpDir, binName);
+      if (!existsSync(extractedBinary)) {
+        throw new Error(
+          `rtk binary missing after extraction: ${extractedBinary} — the release archive layout for ${target} may have changed`,
+        );
+      }
+      try {
```
pauldambra, line 197: > [!NOTE]

The rename-race guard is sound (only swallows when a whole, checksum-verified `cachedBinary` exists; finally cleans tmpDir on all paths), but two concurrent cold-cache builds still share `archivePath` inside cacheDir — one can rmSync/overwrite the archive while the other is verifying or extracting, failing loudly before the guarded rename. Downloading into the per-process tmpDir too would make the race real-safe rather than comment-safe.

packages/agent/build/rtk-binary.mjs
```diff
@@ -0,0 +1,222 @@
...
+      // Don't leave an unverified archive on disk to be reused.
+      rmSync(archivePath, { force: true });
+      throw error;
+    }
+    // Extract into a temp dir first so a killed build cannot leave a partial
+    // binary in cacheDir that existsSync accepts on the next run.
+    // Use dirname(cacheDir) so the temp dir and cacheDir are on the same
+    // filesystem: rename(2) fails with EXDEV across filesystems (e.g. tmpfs /tmp
+    // vs overlay node_modules/.cache in Docker / CI containers).
+    const tmpDir = mkdtempSync(join(dirname(cacheDir), ".tmp-"));
```
pauldambra, line 182: > [!NOTE]

**[qa-team/reliability]** ⚪ NIT

A build killed between `mkdtempSync` and the finally now leaks `.tmp-*` dirs under the cache (os.tmpdir() was OS-reaped). Harmless — nothing enumerates that dir — but an opportunistic sweep of stale `.tmp-*` siblings would keep the cache tidy.

packages/agent/src/server/agent-server.test.ts
```diff
@@ -554,7 +602,8 @@ describe("AgentServer HTTP Mode", () => {
         "boom",
       );
 
-      expect(order).toEqual(["enqueue", "update", "stop"]);
+      // error enqueue -> status update -> rtk_savings enqueue -> stop
+      expect(order).toEqual(["enqueue", "update", "enqueue", "stop"]);
       expect(testServer.eventStreamSender.enqueue).toHaveBeenCalledWith(
```
pauldambra, line 610: > [!NOTE]

The order assertion `["enqueue", "update", "enqueue", "stop"]` never pins the *second* enqueue as `_posthog/rtk_savings` — any payload passes. Today emitRtkSavings is the only other enqueue on this seam so the pin holds, but one `toHaveBeenCalledWith(... method: "_posthog/rtk_savings" ...)` makes it self-verifying against future enqueue additions.

packages/agent/src/server/agent-server.test.ts
```diff
@@ -572,20 +621,17 @@ describe("AgentServer HTTP Mode", () => {
           error_message: "boom",
         },
       );
+      // Assert rtk_savings enqueue precedes stop() on this seam.
+      const enqueueOrders =
+        testServer.eventStreamSender.enqueue.mock.invocationCallOrder;
+      const stopOrder =
+        testServer.eventStreamSender.stop.mock.invocationCallOrder[0];
+      // The last enqueue (rtk_savings) must land before stop()
+      expect(enqueueOrders[enqueueOrders.length - 1]).toBeLessThan(stopOrder);
```
pauldambra: > [!NOTE]

**[qa-team/reliability]** ⚪ NIT

The trailing `invocationCallOrder` assertion (last enqueue < stop) is fully implied by the order-array equality — both mocks push into the same array. Redundant per OnceAndOnlyOnce; drop one.

pauldambra (commented):
> [!NOTE]

QA Swarm round 4 (delta review of 1bcf3c5c5 + 258e019c0). All fixes verified correct; only polish-grade findings remain (unused import, test self-verification, log tag). Overall risk: LOW.

apps/code/vite-main-plugins.mts
```diff
@@ -218,6 +221,56 @@ export function copyClaudeExecutable(): Plugin {
...
+      }
+      signClaudeBinary(destBinary);
```
pauldambra: > [!NOTE]

Good dedupe, but `signClaudeBinary`'s warn/throw messages hardcode the "[copy-claude-executable]" tag, so a failed rtk signing logs under the wrong plugin during build triage. Also the "unreachable throw" claim isn't strict — copyClaudeExecutable early-returns without signing when its staged copy is current, so copyRtkBinary can be first to hit the darwin-on-non-darwin throw (arguably the right behaviour, but with a misleading claude-tagged message). A neutral "[sign-binary]" prefix + the path in the message fixes both.

packages/agent/build/rtk-binary.mjs
pauldambra: > [!NOTE]

Fixed in 99e4db68a — removed the unused `tmpdir` import.

packages/agent/build/rtk-binary.mjs
pauldambra: > [!NOTE]

Fixed in 99e4db68a — the archive now downloads/verifies/extracts wholly inside the per-process temp dir; concurrent builds no longer share archivePath, and the unverified-archive cleanup collapsed into the finally.

packages/agent/build/rtk-binary.mjs
pauldambra: > [!NOTE]

Accepted tradeoff — leaked `.tmp-*` dirs are inert (nothing enumerates the cache dir and rtk targets never match), and a sweep adds code for a kill-window edge; revisit if the cache ever grows an enumerator.

packages/agent/src/server/agent-server.test.ts
pauldambra: > [!NOTE]

Fixed in 99e4db68a — the test now pins the second enqueue as `_posthog/rtk_savings` via toHaveBeenCalledWith.

packages/agent/src/server/agent-server.test.ts
pauldambra: > [!NOTE]

Fixed in 99e4db68a — dropped the redundant invocationCallOrder assertion in favour of the order-array equality.

apps/code/vite-main-plugins.mts
pauldambra: > [!NOTE]

Fixed in 99e4db68a — renamed to `signDarwinBinary` with a neutral `[sign-binary]` tag that names the path being signed.

packages/agent/src/acp-extensions.ts
pauldambra: > [!NOTE]

Schema update relevant to the consumer work: as of `5fe748e62` the event carries gauge semantics — `cumulative_*` counter reads plus a `counter_id` group-by key — so the ingestion/query side must difference readings per counter (argMax - argMin per window, value drop = counter reset), never sum event values. Leaving unresolved for the consumer verification itself.

pauldambra (commented):
> [!NOTE]

QA Swarm round 6 (delta review of 5fe748e62, the gauge reshape). One convergent MEDIUM: counter_id = run_id breaks on warm snapshot-resumed sandboxes where the rtk tally survives into a new run_id.

packages/agent/src/server/agent-server.ts
```diff
@@ -3373,6 +3381,68 @@ ${signedCommitInstructions}
...
+          method: POSTHOG_NOTIFICATIONS.RTK_SAVINGS,
+          // snake_case: these land as PostHog event properties.
+          params: {
+            task_id: this.config.taskId,
+            run_id: this.config.runId,
+            team_id: this.config.projectId,
+            // The identity of the rtk database this reading came from — the
+            // group-by key for differencing. The cloud sandbox is fresh per
+            // run, so the run owns its counter.
+            counter_id: this.config.runId,
```
pauldambra: > [!NOTE]

**[convergent: qa-team/data-integrity + qa-team/generalist-a]** 🟡 MEDIUM

`counter_id: this.config.runId` rests on "the cloud sandbox is fresh per run", but the warm-reconnect path (agent-server.ts ~1368: sandbox restart with snapshot resume) preserves the sandbox filesystem — including rtk's tally database — under a NEW run_id. That resumed run's single reading re-reports the prior run's entire tally under a fresh counter, inflating aggregate savings exactly the way the gauge reshape set out to prevent. Key counter_id to something with the database's lifetime (the task id — a cold resume gets a fresh sandbox and reads a dropped value, which the reset rule already covers), or reset the tally at run start.

packages/agent/src/server/agent-server.ts
```diff
@@ -3373,6 +3381,68 @@ ${signedCommitInstructions}
     this.session = null;
   }
 
+  /**
+   * Emit a gauge snapshot of RTK's cumulative output-compression tally as a
+   * telemetry event, so PostHog can report how much context RTK saves.
+   *
+   * Gauge semantics: the `cumulative_*` properties are counter READS of the
+   * host's machine-global rtk tally, not per-run deltas. Consumers must group
+   * by `counter_id` and difference readings (argMax - argMin per window,
```
pauldambra: > [!NOTE]

**[qa-team/data-integrity]** 🟢 LOW

The differencing recipe as written (argMax - argMin per window) computes 0 for the only case emitted today: exactly one reading per counter. State the zero-baseline rule explicitly — an implicit 0 before a counter's first reading, so a single reading counts in full — or a consumer encoding the recipe literally reports zero savings.

packages/agent/src/server/agent-server.ts
```diff
@@ -3373,6 +3381,68 @@ ${signedCommitInstructions}
...
+            team_id: this.config.projectId,
+            // The identity of the rtk database this reading came from — the
+            // group-by key for differencing. The cloud sandbox is fresh per
+            // run, so the run owns its counter.
+            counter_id: this.config.runId,
+            cumulative_commands: savings.totalCommands,
+            cumulative_input_tokens: savings.inputTokens,
+            cumulative_output_tokens: savings.outputTokens,
+            cumulative_tokens_saved: savings.tokensSaved,
+            avg_savings_pct: savings.avgSavingsPct,
```
pauldambra: > [!NOTE]

**[qa-team/generalist-a]** 🟢 LOW

`avg_savings_pct` sits outside the gauge contract: it's derived from the same cumulative tally but an average can't be differenced, and the docstring is silent on how to read it. Either document it as a last-value snapshot, or drop it and let consumers derive the ratio from cumulative_tokens_saved / cumulative_input_tokens.

packages/agent/src/server/agent-server.test.ts
```diff
@@ -496,29 +501,73 @@ describe("AgentServer HTTP Mode", () => {
...
+        expect.objectContaining({
+          type: "notification",
+          notification: expect.objectContaining({
+            method: "_posthog/rtk_savings",
+            params: expect.objectContaining({
+              task_id: "test-task-id",
+              run_id: "test-run-id",
+              team_id: 1,
+              counter_id: "test-run-id",
+              cumulative_commands: 4,
```
pauldambra, line 539: > [!NOTE]

**[qa-team/generalist-a]** ⚪ NIT

The emit assertion pins counter_id, cumulative_commands, and cumulative_tokens_saved but not cumulative_input_tokens / cumulative_output_tokens — a partial revert of the rename would pass. Adding the remaining keys fully pins it.

packages/agent/src/server/agent-server.ts
pauldambra: > [!NOTE]

Fixed in 251681276 — counter_id is now the task id, matching the sandbox/rtk-database lifetime across warm snapshot-resumes; a cold resume reads a dropped value, which the reset rule already handles. Alternative considered was `rtk gain --reset` at run start — rejected because it destroys the counter a concurrent reader might be mid-diff on and adds a write to a read-only path.

packages/agent/src/server/agent-server.ts
pauldambra: > [!NOTE]

Fixed in 251681276 — the contract now states the implicit-zero baseline before a counter's first reading, so a single reading counts in full.

packages/agent/src/server/agent-server.ts
pauldambra: > [!NOTE]

Fixed in 251681276 — avg_savings_pct dropped end-to-end (emit, RtkSavingsSummary, parse, tests); the docstring notes the ratio is derivable from cumulative_tokens_saved / cumulative_input_tokens.

packages/agent/src/server/agent-server.test.ts
pauldambra: > [!NOTE]

Fixed in 251681276 — all four cumulative_* keys and counter_id are pinned in the assertion.

packages/workspace-server/src/services/agent/agent.ts
pauldambra: > [!NOTE]

Addressed in `e6bd53331` — the stage dir and binary name are now shared consts (`BUNDLED_RTK_DIR`, `bundledRtkBinName`) exported from `@posthog/agent/server/rtk-savings` and consumed by the workspace-server resolver, bin.ts, and the new desktop gauge host. The one remaining literal is the producer side in `vite-main-plugins.mts` (build-config eval can't safely import package dist), documented on the const.

packages/agent/src/server/bin.ts
pauldambra: > [!NOTE]

Fixed in `bc53fab59` — the rewrite hook now probes `rtk --version` lazily on the first command it would rewrite (memoized per session, injectable for tests); a failed probe logs once and downgrades the session to no-rewrite. Guarding the single hook seam covers both hosts, so neither adoption site needs its own check — and `rtk gain` already tolerates a broken binary.

packages/agent/src/acp-extensions.ts
pauldambra: > [!NOTE]

Verified against posthog master: the finding is real. The tasks backend's event ingest writes notifications to the run stream, and `relay_sandbox_events.py` only special-cases `_posthog/task_complete` and `_posthog/error` — nothing maps `_posthog/rtk_savings` into a captured analytics event, so cloud gauge readings are currently stored-but-not-reported. Follow-up needed in posthog: capture the notification params as an "RTK savings gauge" event (the exact shape is `RtkSavingsGaugeProperties` in @posthog/shared — desktop already emits it), and build the consuming insight as a per-counter diff (implicit zero baseline, drop = reset), never a sum. Leaving unresolved until that lands.

gantoine:
I explicitly didn't do this since I don't think bundling `rtk` is the right choice/should be a concern of posthog code, but I'll leave that up to the code team to make that product call.

tatoalo:
why are we bundling binaries at the agent-server level and not at the image deployed?

pauldambra:
> why are we bundling binaries at the agent-server level and not at the image deployed?

​I guess cos I was thinking local and worktree first. 

What's the best route?

tatoalo:
trying a different non-agent-package approach here for now 🙏🏻  (🧵 [thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1783509404871569))

PR #3152, opened 2026-07-05
fix(sessions): stale-conversation cost warning is a blocking prompt-input state

pauldambra:
## Problem

The staff-only warning for continuing a large, idle conversation (#3122) was an `AlertDialog` that flashed up and immediately disappeared.

Root cause: the gate was purely reactive on "newest event `ts` older than 60 min". Opening a stale session reconnects the agent, which immediately re-emits notifications (usage update, SDK handshake) that the workspace-server stamps with `ts: Date.now()` — so the conversation instantly looked active again and the dialog closed itself before it could be read.

## Change

- **Latch the gate** in the view-state store: once it engages for a session (recording the last-activity time observed at that moment), only an explicit user choice releases it. Freshly-stamped reconnect events can no longer dismiss it.
- **No more modal.** The warning renders as a state of the prompt input — the composer slot swaps to the same `ActionSelector` used by permission prompts, so you must choose before continuing:
  1. **Compact and continue** — sends `/compact`; pays the reload once, every later turn is cheaper
  2. **Continue anyway** — full-price reload, keeps the whole conversation
  3. **Start a new session** — avoids the cost entirely
- It takes precedence over a pending permission prompt, since answering one also resumes the costly turn. The thread stays visible and scrollable above, so you can review the conversation while deciding.
- `ActionSelector`'s footer now only advertises "Esc to cancel" when `onCancel` is actually wired (this selector has no cancel; permission prompts are unaffected).
- Adds a Storybook story: `Sessions/StaleConversationCostNotice` (thread backdrop capped at the real `CHAT_CONTENT_MAX_WIDTH`, same 750px wrapper the composer uses — the notice renders at exactly the prompt input's width).

## Testing

- Store latch behaviour covered in `staleConversationGateStore.test.ts` (engage snapshots, idempotence, acknowledged sessions can't re-engage)
- `pnpm --filter @posthog/ui typecheck`, Biome clean
- Visual check via the Storybook story in light and dark themes

---


pauldambra:
* **#3152** <a href="https://app.graphite.com/github/pr/PostHog/code/3152?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/3152?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -572,7 +577,29 @@ export function SessionView({
                       )}
                     </Flex>
                   </Flex>
-                ) : hideInput ? null : firstPendingPermission ? (
+                ) : hideInput ? null : staleGate.active ? (
+                  // Replaces the composer (and any pending permission — answering
```
pauldambra, line 662: > [!NOTE]

**[qa-team/frontend]** 🟡 MEDIUM

"Compact and continue" while a permission is pending pays the reload twice. This branch supersedes `PermissionSelector`, but `handleStaleCompact` queues `/compact` behind the still-unanswered permission. After the gate releases, the permission prompt appears; answering it resumes the costly turn at full price, and then the queued compaction re-processes the whole conversation again. The option's description ("Pays the reload once") is wrong in exactly this case. Consider surfacing/cancelling the pending permission as part of the compact choice, or suppressing the compact option while `firstPendingPermission` exists.

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -271,6 +271,11 @@ export function SessionView({
   // prompt cache has likely expired (see useStaleConversationGate).
   const staleGate = useStaleConversationGate(sessionId, events);
 
+  const handleStaleCompact = useCallback(() => {
```
pauldambra: > [!NOTE]

**[qa-team/generalist]** 🟡 MEDIUM

`handleStaleCompact` acknowledges the gate (`staleGate.onContinue()`) before the async `onSendPrompt("/compact")` has succeeded, and ignores its promise. If the send fails (offline — this notice has no `isOnline` guard, unlike `PromptInput`'s `submitDisabledExternal` — or any send error), the user gets a toast but the gate is permanently released for this session (per app run) and no compaction happened; their next message pays full price with no warning. Await the send and only acknowledge on success (or re-engage on failure).

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -572,7 +577,29 @@ export function SessionView({
...
+                  <Box className="min-h-0 shrink-0 overflow-y-auto">
+                    <Box
+                      className={compact ? "p-1" : "mx-auto px-2 pb-3"}
+                      style={
+                        compact
+                          ? undefined
+                          : { maxWidth: CHAT_CONTENT_MAX_WIDTH }
+                      }
+                    >
+                      <StaleConversationCostNotice
```
pauldambra, line 666: > [!NOTE]

**[convergent: qa-team/frontend + paul]** 🟡 MEDIUM

The gate branch renders regardless of `isRunning`, replacing the "Connecting to agent..." state, so the notice is actionable before the agent is connected. Choosing "Compact and continue" or "Continue anyway" then sends `/compact` into a session that may still be mid-reconnect — does the prompt land, or is it silently dropped? Worth verifying `sessionService.sendPrompt` queues this cleanly, or disabling the send-producing options until `isRunning`.

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -271,6 +271,11 @@ export function SessionView({
   // prompt cache has likely expired (see useStaleConversationGate).
   const staleGate = useStaleConversationGate(sessionId, events);
 
+  const handleStaleCompact = useCallback(() => {
+    staleGate.onContinue();
+    onSendPrompt("/compact");
```
pauldambra, line 345: > [!NOTE]

**[convergent: qa-team/generalist + xp]** 🟢 LOW

`onSendPrompt("/compact")` routes through the typed-input slash-command pipeline (`tryExecuteCodeCommand`, `rewriteLocalSkillCommandPrompt`, `resolveLocalSkillPrompt`). A local skill named `compact` would silently rewrite this button into a skill invocation instead of SDK compaction, and the literal string couples this view to the command parser's spelling with no test guarding it. A dedicated compaction call on the session service (or raw-send path) would keep the contract in one place.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
```diff
@@ -0,0 +1,72 @@
...
+  onNewSession?: () => void;
+}
+
+/**
+ * Composer state shown in place of the prompt input when PostHog staff return
+ * to a large, idle conversation whose prompt cache has likely expired. Uses
+ * the same ActionSelector as permission prompts, so the user must choose how
+ * to continue before they can type again.
+ */
+export function StaleConversationCostNotice({
```
pauldambra, line 31: > [!NOTE]

**[qa-team/frontend]** 🟢 LOW

Nothing focuses the `ActionSelector` when the notice mounts. Its capture-phase key handler ignores keys unless the container has focus, so the promised "Enter to select · Tab/Arrow keys to navigate" hint is dead until the user clicks the box (the permission path has a `requestFocus` effect keyed on `firstPendingPermission`; this path has no equivalent). Focus the selector container on mount.

packages/ui/src/features/sessions/useStaleConversationGate.ts
```diff
@@ -40,46 +36,40 @@ export function useStaleConversationGate(
...
-    (open: boolean) => setDismissedFor(open ? null : sessionId),
-    [sessionId],
-  );
-  const onReopen = useCallback(() => setDismissedFor(null), []);
 
   return {
-    active,
-    dialogOpen: active && !dismissed,
-    dismissed,
+    active: shouldEngage || engagement !== undefined,
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

`active: shouldEngage || engagement !== undefined` re-derives "engaged" two ways. Once the effect has run, `engagement` is the source of truth; `shouldEngage` only matters for the first render before the effect fires. It works, but a reader has to reason about effect timing to see why both terms are needed — one line of "why both" (`shouldEngage` covers the pre-effect first paint) would repay itself.

packages/ui/src/features/sessions/useStaleConversationGate.ts
```diff
@@ -40,46 +36,40 @@ export function useStaleConversationGate(
...
-  const [dismissedFor, setDismissedFor] = useState<string | null>(null);
-  const dismissed = active && dismissedFor === sessionId;
+  useEffect(() => {
+    if (shouldEngage) engage(sessionId, liveLastActivityAt);
+  }, [shouldEngage, engage, sessionId, liveLastActivityAt]);
 
   const onContinue = useCallback(
     () => acknowledge(sessionId),
     [acknowledge, sessionId],
   );
```
pauldambra, line 69: > [!NOTE]

**[paul]** 🟢 LOW

this is the staff-only costly-conversation gate — is there any way to tell in production whether folk are hitting it and which of the three options they pick? we've built a nice three-way choice (compact / continue / new session) but nothing captures the decision. a capture on select would let us build a funnel later — how often does the notice fire, and do people actually compact vs eat the cost. not blocking, but measure-first instinct is itching.

packages/ui/src/primitives/action-selector/ActionSelector.tsx
```diff
@@ -326,7 +326,8 @@ export function ActionSelector({
           </Flex>
 
           <Text color="gray" mt="2" as="p" className="text-[13px]">
-            Enter to select · Tab/Arrow keys to navigate · Esc to cancel
+            Enter to select · Tab/Arrow keys to navigate
+            {onCancel ? " · Esc to cancel" : ""}
```
pauldambra, line 339: > [!NOTE]

**[convergent: qa-team/frontend + paul + xp]** 🟢 LOW

Good fix hiding "Esc to cancel" when there is no `onCancel` — but the keydown handler still `preventDefault`s and `stopPropagation`s Escape at capture phase and invokes a no-op when `onCancel` is undefined. The new cancel-less usage makes this dead-but-greedy path reachable for the first time: while the selector is focused, Escape is silently swallowed (a problem if any parent overlay relies on it bubbling). Skip the Escape case when `onCancel` is undefined.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
```diff
@@ -0,0 +1,72 @@
...
+                label: "Start a new session",
+                description: "Avoids the cost entirely",
+              },
+            ]
+          : []),
+      ]}
+      onSelect={(optionId) => {
+        if (optionId === COMPACT_OPTION) onCompact();
+        else if (optionId === CONTINUE_OPTION) onContinue();
+        else onNewSession?.();
```
pauldambra: > [!NOTE]

**[xp]** ⚪ NIT

The `onSelect` dispatch's final `else` implicitly means "new-session". Since that option is conditionally present, an unexpected id would silently fall into `onNewSession?.()`. An explicit `else if (optionId === NEW_SESSION_OPTION)` reads truer to intent and won't misroute a future fourth option.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const activity =
+    lastActivityAt !== null
+      ? `was last active ${formatRelativeTimeLong(lastActivityAt)}`
+      : "has been idle";
+  const spent =
+    costUsd !== null ? ` (≈$${costUsd.toFixed(2)} spent so far)` : "";
+  return (
+    <ActionSelector
+      title="Continue this large, idle conversation?"
+      question={`This conversation holds about ${formatTokensCompact(usedTokens)} tokens and ${activity}${spent}. Its prompt cache has likely expired, so the next message re-processes everything at full input price instead of the ~10% cached rate. How do you want to continue?`}
```
pauldambra: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

The spend parenthetical now dangles off the activity clause — "…and was last active 2 hours ago (≈$12.34 spent so far)" — where the old dialog attached it to the pricing sentence it explains. Consider "…at full input price instead of the ~10% cached rate (≈$12.34 spent so far)" to keep cost next to cost.

pauldambra (commented):
> [!NOTE]

pauldambra:
> [!NOTE]
>

## Verdict: 💬 APPROVE WITH NITS <sub>(round 2 @ 92a819f)</sub>

The round-1 fixes are faithfully implemented with no regressions; the delta's new dynamism opens two narrow edge cases worth closing. Nothing blocks.

### Key findings

- 🟡 **MEDIUM** — the conditional compact option gives `ActionSelector` its first mutable options list; `selectedIndex` is never clamped, so a mid-decision permission arrival can make Enter throw or fire the wrong choice (`StaleConversationCostNotice.tsx:49`)
- 🟢 **LOW** — `staleGate.active && !isRunning` falls through to an answerable `PermissionSelector` during "connecting" windows where `pendingPermissions` isn't cleared, briefly uncovering the action the gate exists to block (`SessionView.tsx:609`)
- ⚪ NITs — derive the choice union from `StaleConversationGateChoiceProperties["choice"]`; `useMemo` vs `useCallback` inconsistency in the three gate handlers; no story for the two-option (permission-pending) layout; copy preference on the spend parenthetical placement

### Convergence

None this round — each finding came from a single reviewer.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Round-1 fixes verified (typecheck + store tests pass); two edge cases from the new conditional option and isRunning gating are worth closing. |
| 👤 paul | Every round-1 finding got a clean fix; loves the analytics wiring; only a copy nit on where "spent so far" landed. |
| 📐 xp | Small, well-motivated behaviour changes with load-bearing comments; two NITs about pre-existing duplication shapes the delta merely follows. |
| 🛡 security-audit | No findings: the new event carries only a closed enum and two numeric aggregates, staff-only, via the existing typed sink. |

<details>
<summary>Previous rounds (1)</summary>

round 1 @ f05d8c5 — 💬 APPROVE WITH NITS: 3 MEDIUM (compact/permission double reload, ack-before-send, mid-reconnect actionability), 5 LOW, 2 NIT; all fixed in `92a819fe2` or resolved as intentional.

</details>

---

packages/ui/src/features/sessions/components/SessionView.tsx
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — "Compact and continue" is now suppressed while a permission is pending (the notice's `onCompact` is omitted, so the option doesn't render). Did this rather than cancelling the pending permission because cancelling aborts the agent's in-flight work destructively — shout if you'd rather the cancel-then-compact behaviour.

packages/ui/src/features/sessions/components/SessionView.tsx
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — `handleStaleCompact` now guards on `isOnline` (offline shows the toast and keeps the gate engaged). Residual: a non-offline send failure still releases the gate before the toast; fixing that fully means making `onSendPrompt` awaitable through the SessionView contract, which felt out of scope — flagging in case you'd rather take that on.

packages/ui/src/features/sessions/components/SessionView.tsx
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — the gate branch now requires `isRunning`, so while the agent reconnects the existing "Connecting to agent..." state shows and the choices can't fire into a half-connected session. (`sessionService.sendPrompt` failures do surface as a toast, but gating on `isRunning` keeps the options truthful.)

packages/ui/src/features/sessions/components/SessionView.tsx
pauldambra: > [!NOTE]

Intentional — the button reproduces exactly what typing `/compact` does (the context-limit error view literally tells users to type it), so sharing the pipeline keeps button and typed behaviour identical, including any future command renames. A dedicated compaction call on the session service is a reasonable follow-up if one appears; keeping this PR at parity.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
pauldambra: > [!NOTE]

Not reproducible as described — `useActionSelectorState` focuses the container on mount (`useEffect` at useActionSelectorState.ts:106-110), exactly as it does for permission prompts; the only case it declines is when focus is already in an interactive element, which is the deliberate anti-focus-steal guard shared with the permission path.

packages/ui/src/features/sessions/useStaleConversationGate.ts
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — added the one-line "why both" comment (shouldEngage covers the pre-effect first paint; engagement covers every render after).

packages/ui/src/features/sessions/useStaleConversationGate.ts
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — added a typed `Stale conversation gate choice` analytics event (choice: compact/continue/new_session, plus used_tokens and cost_usd) fired from the SessionView handlers, following the existing `ANALYTICS_EVENTS`/`EventPropertyMap` convention.

packages/ui/src/primitives/action-selector/ActionSelector.tsx
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — the Escape case now breaks early (no preventDefault/stopPropagation) when `onCancel` is undefined, so the key bubbles instead of being swallowed by a no-op.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — explicit `else if (optionId === NEW_SESSION_OPTION)` branch.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
pauldambra: > [!NOTE]

Fixed in `92a819fe2` — the spend parenthetical now sits after the pricing sentence ("...instead of the ~10% cached rate (≈$12.34 spent so far)").

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -572,7 +577,29 @@ export function SessionView({
...
+                        usedTokens={staleGate.usedTokens}
+                        lastActivityAt={staleGate.lastActivityAt}
+                        costUsd={staleGate.costUsd}
+                        onContinue={staleGate.onContinue}
+                        onCompact={handleStaleCompact}
+                        onNewSession={onNewSession}
+                      />
+                    </Box>
+                  </Box>
+                ) : firstPendingPermission ? (
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Intentional precedence — answering a pending permission resumes the costly turn, which is exactly what the gate exists to warn about, so the warning must come first (the code comment at the branch notes this). The permission is deferred, never dropped: choosing any option releases the gate and the `PermissionSelector` renders as before. `92a819fe2` also suppresses "Compact and continue" while a permission is pending so the queued /compact can't pay the reload twice.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
```diff
@@ -0,0 +1,81 @@
...
+      ? `was last active ${formatRelativeTimeLong(lastActivityAt)}`
+      : "has been idle";
+  const spent =
+    costUsd !== null ? ` (≈$${costUsd.toFixed(2)} spent so far)` : "";
+  return (
+    <ActionSelector
+      title="Continue this large, idle conversation?"
+      question={`This conversation holds about ${formatTokensCompact(usedTokens)} tokens and ${activity}. Its prompt cache has likely expired, so the next message re-processes everything at full input price instead of the ~10% cached rate${spent}. How do you want to continue?`}
+      options={[
+        ...(onCompact
```
pauldambra, line 50: > [!NOTE]

**[qa-team/frontend]** 🟡 MEDIUM

This is the first `ActionSelector` consumer whose `options` array changes while mounted: the compact option toggles with `firstPendingPermission`, which can flip while the notice is on screen (a restored permission request typically arrives shortly after `isRunning` goes true). `useActionSelectorState` never resets or clamps `selectedIndex` when options change — if the user arrowed to index 2 and the list shrinks to 2 entries, Enter dereferences `allOptions[selectedIndex].id` as `undefined` and throws inside the document-capture keydown handler; short of that, a highlighted index silently changes meaning (index 0 flips from "Compact and continue" to "Continue anyway") so Enter/`1` fires the wrong choice. Fix: clamp `selectedIndex` in the hook when the options list changes, or force a remount here with `key={onCompact ? "with-compact" : "no-compact"}`.

pauldambra (commented):
> [!NOTE]

QA Swarm round 2 complete. See inline comments.

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -572,7 +606,38 @@ export function SessionView({
                       )}
                     </Flex>
                   </Flex>
-                ) : hideInput ? null : firstPendingPermission ? (
+                ) : hideInput ? null : staleGate.active && isRunning ? (
```
pauldambra: > [!NOTE]

`staleGate.active && isRunning` falls through to the `firstPendingPermission` branch when the gate is active but the session is not running. The handoff paths in sessionService set `status: "connecting"` without clearing `pendingPermissions`, so in that window an answerable `PermissionSelector` renders — and answering a permission resumes the costly turn, which is exactly what the gate exists to intercept. Rendering the connecting state instead of falling through when `staleGate.active && !isRunning` would keep the invariant that the gate always covers the pending permission.

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -271,6 +272,39 @@ export function SessionView({
   // prompt cache has likely expired (see useStaleConversationGate).
   const staleGate = useStaleConversationGate(sessionId, events);
 
+  const trackStaleGateChoice = useCallback(
+    (choice: "compact" | "continue" | "new_session") =>
```
pauldambra: > [!NOTE]

The `"compact" | "continue" | "new_session"` union is re-declared inline here when it already exists as `StaleConversationGateChoiceProperties["choice"]` in @posthog/shared. Deriving the parameter type from the shared interface keeps the event contract OnceAndOnlyOnce.

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -271,6 +272,39 @@ export function SessionView({
...
+    staleGate.onContinue();
+    onSendPrompt("/compact");
+  }, [isOnline, trackStaleGateChoice, staleGate.onContinue, onSendPrompt]);
+
+  const handleStaleContinue = useCallback(() => {
+    trackStaleGateChoice("continue");
+    staleGate.onContinue();
+  }, [trackStaleGateChoice, staleGate.onContinue]);
+
+  const handleStaleNewSession = useMemo(() => {
```
pauldambra: > [!NOTE]

**[xp]** ⚪ NIT

`handleStaleNewSession` uses `useMemo` returning a closure while its siblings use `useCallback` — the `useMemo` only exists to fold in the "undefined when `onNewSession` absent" guard. Keeping `useCallback` and doing the conditional at the call site (as `onCompact` already does) would make the three handlers read consistently.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+        ))}
+      </Flex>
+    </Box>
+  );
+}
+
+const TWO_HOURS_AGO = Date.now() - 2 * 60 * 60 * 1000;
+
+export const Default: Story = {
+  args: {
```
pauldambra, line 63: > [!NOTE]

**[qa-team/frontend]** ⚪ NIT

The new `onCompact: undefined` state (compact hidden while a permission is pending) has no story — both stories pass `onCompact`. A "PermissionPending" variant would document the two-option layout this round introduced.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const activity =
+    lastActivityAt !== null
+      ? `was last active ${formatRelativeTimeLong(lastActivityAt)}`
+      : "has been idle";
+  const spent =
+    costUsd !== null ? ` (≈$${costUsd.toFixed(2)} spent so far)` : "";
+  return (
+    <ActionSelector
+      title="Continue this large, idle conversation?"
+      question={`This conversation holds about ${formatTokensCompact(usedTokens)} tokens and ${activity}. Its prompt cache has likely expired, so the next message re-processes everything at full input price instead of the ~10% cached rate${spent}. How do you want to continue?`}
```
pauldambra, line 48: > [!NOTE]

**[paul]** ⚪ NIT

nit picking the copy move... `(≈$X spent so far)` now trails the cached-rate sentence — "…instead of the ~10% cached rate (≈$2.40 spent so far)." the parenthetical about money spent attaches to a sentence about the cache rate rather than the conversation. i think the round-1 placement (after the activity clause) read more naturally. not blocking, just where the eye lands.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
pauldambra: > [!NOTE]

Fixed in `3ec48e3df` — the notice now remounts when the option set changes (`key={firstPendingPermission ? "no-compact" : "with-compact"}`), resetting the selector's highlighted index. Chose the scoped remount over clamping `selectedIndex` inside `useActionSelectorState` to avoid touching every other consumer's behaviour — shout if you'd rather the general clamp in the primitive.

packages/ui/src/features/sessions/components/SessionView.tsx
```diff
@@ -572,7 +606,38 @@ export function SessionView({
...
+                        lastActivityAt={staleGate.lastActivityAt}
+                        costUsd={staleGate.costUsd}
+                        onContinue={handleStaleContinue}
+                        // A queued /compact would land after the pending
+                        // permission resumes the costly turn — paying the
+                        // reload twice — so the option hides until then.
+                        onCompact={
+                          firstPendingPermission
+                            ? undefined
+                            : handleStaleCompact
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `3ec48e3df` — the notice remounts when the option set changes (keyed on whether compact is offered), so the selector's highlighted index resets instead of dereferencing past the shrunk list.

packages/ui/src/features/sessions/components/SessionView.tsx
pauldambra: > [!NOTE]

Fixed in `3ec48e3df` — when the gate is active but the session isn't running, the composer slot now renders the connecting state instead of falling through, so a lingering pending permission can't render answerable underneath the gate.

packages/ui/src/features/sessions/components/SessionView.tsx
pauldambra: > [!NOTE]

Fixed in `3ec48e3df` — the choice parameter is now typed as `StaleConversationGateChoiceProperties["choice"]`.

packages/ui/src/features/sessions/components/SessionView.tsx
pauldambra: > [!NOTE]

Fixed in `3ec48e3df` — `handleStaleNewSession` is a `useCallback` like its siblings, with the undefined guard moved to the call site (`onNewSession ? handleStaleNewSession : undefined`), matching how `onCompact` is wired.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.stories.tsx
pauldambra: > [!NOTE]

Fixed in `3ec48e3df` — added the `PermissionPending` story (no `onCompact`) documenting the two-option layout.

packages/ui/src/features/sessions/components/StaleConversationCostNotice.tsx
pauldambra: > [!NOTE]

Intentional — round 1's qa-team/copy finding asked for exactly this placement (cost next to the pricing sentence it explains), so moving it back would just flip-flop between reviewer preferences. Keeping the round-1 convergence; happy to change if a human prefers the activity-clause placement.

charlesvien (approved):

PR #3163, opened 2026-07-06
fix(queued-messages): clamp long queued messages to 3 lines

harshjainnn:
Long queued messages (e.g. a pasted code block) previously rendered at full height, and this compounded with multiple queued messages, pushing the composer off-screen. Clamp to 3 lines with a fade mask and a 'Show more/less' toggle, mirroring the clamp pattern already used for sent messages in ChatThread's UserBubble.

Fixes #3150

## Problem

Queued follow-up messages render at full height with no clamp. A long queued message (e.g. a pasted code block) can grow the dock to hundreds of pixels tall, and this compounds when multiple long messages are queued at once — pushing the composer off-screen.

## Changes

Clamped `QueuedMessageView` content to 3 lines by default, with a bottom fade mask and a "Show more/less" toggle that only appears when the content actually overflows the clamp. This mirrors the existing clamp-by-measured-height pattern used in `ChatThread.tsx`'s `UserBubble` for sent messages — a `ResizeObserver` compares `scrollHeight` against the clamped `clientHeight`, since overflow can't be determined from character count alone (it depends on wrap width).

Also changed the outer row's alignment from `align="center"` to `align="start"` so the icon and action buttons (Steer/Edit/Discard) stay pinned to the top when content spans multiple lines, instead of being vertically centered against a tall block.

## How did you test this?

- Queued several short messages — dock renders unchanged, no toggle appears.
- Queued a message containing a long code block — content clamps to 3 lines with a fade, "Show more" expands it, "Show less" re-collapses it.
- Queued a mix of short and long messages together — short ones render untouched, long ones clamp independently of each other.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/ui/src/features/sessions/components/session-update/QueuedMessageView.tsx
+++ b/packages/ui/src/features/sessions/components/session-update/QueuedMessageView.tsx
@@ -53,7 +53,7 @@ export function QueuedMessageView({
   return (
     <Box className="rounded-lg border border-gray-5 bg-card px-3 py-2">
       <Flex align="start" gap="2">
-        <Stack size={14} className="shrink-0 text-gray-9 mt-0.5" />
+        <Stack size={14} className="mt-0.5 shrink-0 text-gray-9" />
         <Box className="min-w-0 flex-1 font-medium text-[13px] text-gray-12 [&>*:last-child]:mb-0">
           <div
             ref={textRef}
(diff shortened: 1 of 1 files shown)


charlesvien:
Hey @harshjainnn, closing this in favor of https://github.com/PostHog/code/pull/2904

Let me know if you disagree/have any thoughts!

harshjainnn:
Let me know the improvements for the issue required and I will commit them.Why did you close this PR?

PR #3167, opened 2026-07-06
fix(updates): make changelog cache and loading state version-aware

pauldambra:
## Problem

Opening the "What's New" modal (changelog notification -> click the version number) briefly shows the previously cached releases list before the fresh one loads. On a slow network this stale flash is confusing.

Two cache layers were version-blind:

1. **Renderer (TanStack Query)** — `githubReleases.list` had no input, so the query key was static: any prior cache entry rendered instantly while a background refetch swapped content underneath.
2. **Main process** — `GitHubReleasesService` served any cache within its 10-minute TTL, even when it predated the release being viewed. `UpdateAvailableModal` then silently fell back to the wrong ("latest") release's notes.

## Fix

Thread an optional `expectVersion` through `githubReleases.list`:

- The version joins the renderer query key, so a new app version gets a cold cache and the existing skeleton shows instead of stale content. Reopening on the same version stays an instant cache hit.
- The service refetches when its cache lacks the expected version, with a 60s cooldown so an unpublished release (or a local dev build version) doesn't hammer the GitHub API. Failure fallback to stale cache is unchanged.
- `WhatsNewModal` passes the current app version; `UpdateAvailableModal` passes the incoming update's version (also fixing its silent wrong-notes fallback).

## Tests

Parameterized cache-validity matrix plus discrete tests for the resolved-miss, cooldown, and failure-fallback paths in `github-releases.test.ts`.

---

### DIFF
--- a/packages/ui/src/features/updates/UpdateAvailableModal.tsx
+++ b/packages/ui/src/features/updates/UpdateAvailableModal.tsx
@@ -62,12 +62,14 @@ export function UpdateAvailableModal() {
   const downloadMutation = useMutation(
     hostTRPC.updates.download.mutationOptions(),
   );
+  const targetVersion = version ?? availableVersion;
   const { data: releasesData, isPending: isPendingReleases } = useQuery({
-    ...hostTRPC.githubReleases.list.queryOptions(),
+    ...hostTRPC.githubReleases.list.queryOptions(
+      targetVersion ? { expectVersion: targetVersion } : undefined,
+    ),
     enabled: isOpen,
   });
 
-  const targetVersion = version ?? availableVersion;
   const percent = Math.round(downloadPercent ?? 0);
   const sizeLabel = formatSize(downloadSizeBytes);
   const isDownloading = status === "downloading";
--- a/packages/ui/src/features/updates/WhatsNewModal.tsx
+++ b/packages/ui/src/features/updates/WhatsNewModal.tsx
@@ -45,7 +45,11 @@ export function WhatsNewModal() {
   const { data: currentVersion } = useQuery(
     hostTRPC.os.getAppVersion.queryOptions(),
   );
-  const { data, isLoading, isError } = useQuery({
+  const {
+    data,
+    isPending: isLoading,
+    isError,
+  } = useQuery({
     ...hostTRPC.githubReleases.list.queryOptions(
       currentVersion ? { expectVersion: currentVersion } : undefined,
     ),
--- a/packages/ui/src/features/updates/WhatsNewModal.tsx
+++ b/packages/ui/src/features/updates/WhatsNewModal.tsx
@@ -42,19 +42,20 @@ export function WhatsNewModal() {
   const isOpen = useWhatsNewStore((state) => state.isOpen);
   const close = useWhatsNewStore((state) => state.close);
   const hostTRPC = useHostTRPC();
-  const { data: currentVersion } = useQuery(
+  const { data: currentVersion, isError: isVersionError } = useQuery(
     hostTRPC.os.getAppVersion.queryOptions(),
   );
   const {
     data,
-    isPending: isLoading,
-    isError,
+    isPending,
+    isError: isReleasesError,
   } = useQuery({
     ...hostTRPC.githubReleases.list.queryOptions(
       currentVersion ? { expectVersion: currentVersion } : undefined,
     ),
     enabled: isOpen && currentVersion !== undefined,
   });
+  const isError = isVersionError || isReleasesError;
 
   const groups = groupReleases(data?.releases ?? []);
 
@@ -82,7 +83,7 @@ export function WhatsNewModal() {
           </Dialog.Close>
         </Flex>
 
-        {isLoading ? (
+        {isPending ? (
           <ChangelogSkeleton />
         ) : isError ? (
           <Text color="gray" size="2">
--- a/packages/ui/src/features/updates/WhatsNewModal.tsx
+++ b/packages/ui/src/features/updates/WhatsNewModal.tsx
@@ -42,13 +42,15 @@ export function WhatsNewModal() {
   const isOpen = useWhatsNewStore((state) => state.isOpen);
   const close = useWhatsNewStore((state) => state.close);
   const hostTRPC = useHostTRPC();
-  const { data, isLoading, isError } = useQuery({
-    ...hostTRPC.githubReleases.list.queryOptions(),
-    enabled: isOpen,
-  });
   const { data: currentVersion } = useQuery(
     hostTRPC.os.getAppVersion.queryOptions(),
   );
+  const { data, isLoading, isError } = useQuery({
+    ...hostTRPC.githubReleases.list.queryOptions(
+      currentVersion ? { expectVersion: currentVersion } : undefined,
+    ),
+    enabled: isOpen && currentVersion !== undefined,
+  });
 
   const groups = groupReleases(data?.releases ?? []);
 
--- a/packages/ui/src/features/updates/WhatsNewModal.tsx
+++ b/packages/ui/src/features/updates/WhatsNewModal.tsx
@@ -53,7 +53,7 @@ export function WhatsNewModal() {
     ...hostTRPC.githubReleases.list.queryOptions(
       currentVersion ? { expectVersion: currentVersion } : undefined,
     ),
-    enabled: isOpen && currentVersion !== undefined,
+    enabled: isOpen && !!currentVersion,
   });
   const isError = isVersionError || isReleasesError;
 
@@ -83,12 +83,12 @@ export function WhatsNewModal() {
           </Dialog.Close>
         </Flex>
 
-        {isPending ? (
-          <ChangelogSkeleton />
-        ) : isError ? (
+        {isError ? (
           <Text color="gray" size="2">
             Could not load releases. Please try again later.
           </Text>
+        ) : isPending ? (
+          <ChangelogSkeleton />
         ) : groups.length === 0 ? (
           <Text color="gray" size="2">
             No releases found.
--- a/packages/workspace-server/src/services/github-releases/github-releases.test.ts
+++ b/packages/workspace-server/src/services/github-releases/github-releases.test.ts
@@ -69,11 +69,75 @@ describe("GitHubReleasesService", () => {
     });
   });
 
-  it("caches results within the TTL", async () => {
+  it.each([
+    { expectVersion: undefined, expectedFetches: 1 },
+    { expectVersion: "1.2.0", expectedFetches: 1 },
+    { expectVersion: "1.3.0", expectedFetches: 2 },
+  ])(
+    "a fresh cache is a hit for expectVersion $expectVersion only when it contains it ($expectedFetches fetches)",
+    async ({ expectVersion, expectedFetches }) => {
+      const service = new GitHubReleasesService();
+      await service.listReleases();
+      await service.listReleases(expectVersion);
+      expect(fetchMock).toHaveBeenCalledTimes(expectedFetches);
+    },
+  );
+
+  it("caches the refetched list once it contains the expected version", async () => {
     const service = new GitHubReleasesService();
     await service.listReleases();
-    await service.listReleases();
+
+    fetchMock.mockResolvedValueOnce({
+      ok: true,
+      status: 200,
+      json: async () => [
+        {
+          tag_name: "v1.3.0",
+          name: "v1.3.0",
+          body: "new",
+          draft: false,
+          prerelease: false,
+          published_at: "2026-06-30T00:00:00Z",
+          html_url: "https://github.com/PostHog/code/releases/tag/v1.3.0",
+        },
+        ...sampleReleases,
+      ],
+    });
+    const second = await service.listReleases("1.3.0");
+    expect(second.releases[0].version).toBe("1.3.0");
+
+    const third = await service.listReleases("1.3.0");
+    expect(third).toEqual(second);
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+  });
+
+  it("waits out a cooldown before refetching a still-missing version", async () => {
+    const nowSpy = vi.spyOn(Date, "now").mockReturnValue(0);
+    const service = new GitHubReleasesService();
+    await service.listReleases("9.9.9");
+    await service.listReleases("9.9.9");
     expect(fetchMock).toHaveBeenCalledTimes(1);
+
+    nowSpy.mockReturnValue(61_000);
+    await service.listReleases("9.9.9");
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+    nowSpy.mockRestore();
+  });
+
+  it("serves stale cache when a version-miss refetch fails, without retrying within the cooldown", async () => {
+    const nowSpy = vi.spyOn(Date, "now").mockReturnValue(0);
+    const service = new GitHubReleasesService();
+    const first = await service.listReleases();
+
+    fetchMock.mockResolvedValueOnce({ ok: false, status: 500 });
+    const second = await service.listReleases("1.3.0");
+    expect(second).toEqual(first);
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+
+    const third = await service.listReleases("1.3.0");
+    expect(third).toEqual(first);
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+    nowSpy.mockRestore();
   });
 
   it("throws on non-ok responses", async () => {
--- a/packages/workspace-server/src/services/github-releases/github-releases.test.ts
+++ b/packages/workspace-server/src/services/github-releases/github-releases.test.ts
@@ -111,6 +111,17 @@ describe("GitHubReleasesService", () => {
     expect(fetchMock).toHaveBeenCalledTimes(2);
   });
 
+  it("dedupes concurrent cache misses into a single fetch", async () => {
+    const service = new GitHubReleasesService();
+    const [first, second] = await Promise.all([
+      service.listReleases(),
+      service.listReleases("1.3.0"),
+    ]);
+
+    expect(first).toEqual(second);
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+  });
+
   it("waits out a cooldown before refetching a still-missing version", async () => {
     const nowSpy = vi.spyOn(Date, "now").mockReturnValue(0);
     const service = new GitHubReleasesService();
--- a/packages/workspace-server/src/services/github-releases/github-releases.test.ts
+++ b/packages/workspace-server/src/services/github-releases/github-releases.test.ts
@@ -122,6 +122,22 @@ describe("GitHubReleasesService", () => {
     expect(fetchMock).toHaveBeenCalledTimes(1);
   });
 
+  it("concurrent callers both reject when the shared fetch fails, and inFlight is cleared so subsequent calls retry", async () => {
+    fetchMock.mockRejectedValueOnce(new Error("network error"));
+    const service = new GitHubReleasesService();
+
+    const [result1, result2] = await Promise.allSettled([
+      service.listReleases(),
+      service.listReleases("1.2.0"),
+    ]);
+
+    expect(result1.status).toBe("rejected");
+    expect(result2.status).toBe("rejected");
+    // inFlight must be cleared so the next call retries rather than hanging
+    await service.listReleases();
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+  });
+
   it("waits out a cooldown before refetching a still-missing version", async () => {
     const nowSpy = vi.spyOn(Date, "now").mockReturnValue(0);
     const service = new GitHubReleasesService();
--- a/packages/workspace-server/src/services/github-releases/github-releases.ts
+++ b/packages/workspace-server/src/services/github-releases/github-releases.ts
@@ -22,10 +22,11 @@ export class GitHubReleasesService {
     }
 
     // The fetch is version-agnostic, so any concurrent caller can share it.
-    if (this.inFlight === null) {
-      this.inFlight = this.fetchAndCacheReleases(now);
+    let promise = this.inFlight;
+    if (promise === null) {
+      promise = this.fetchAndCacheReleases();
+      this.inFlight = promise;
     }
-    const promise = this.inFlight!;
     try {
       const data = await promise;
       this.updateMissingVersionCooldown(normalizedVersion, now);
@@ -43,9 +44,7 @@ export class GitHubReleasesService {
     }
   }
 
-  private async fetchAndCacheReleases(
-    now: number,
-  ): Promise<ListReleasesOutput> {
+  private async fetchAndCacheReleases(): Promise<ListReleasesOutput> {
     const response = await fetch(RELEASES_URL, {
       headers: { Accept: "application/vnd.github+json" },
       signal: AbortSignal.timeout(FETCH_TIMEOUT_MS),
@@ -70,7 +69,7 @@ export class GitHubReleasesService {
       }));
 
     const data: ListReleasesOutput = { releases };
-    this.cache = { fetchedAt: now, data };
+    this.cache = { fetchedAt: Date.now(), data };
     return data;
   }
 
--- a/packages/workspace-server/src/services/github-releases/github-releases.ts
+++ b/packages/workspace-server/src/services/github-releases/github-releases.ts
@@ -21,9 +21,13 @@ export class GitHubReleasesService {
       return cached;
     }
 
+    // The fetch is version-agnostic, so any concurrent caller can share it.
+    if (this.inFlight === null) {
+      this.inFlight = this.fetchAndCacheReleases(now);
+    }
+    const promise = this.inFlight!;
     try {
-      this.inFlight ??= this.fetchAndCacheReleases(now);
-      const data = await this.inFlight;
+      const data = await promise;
       this.updateMissingVersionCooldown(normalizedVersion, now);
       return data;
     } catch (error) {
@@ -33,7 +37,9 @@ export class GitHubReleasesService {
       }
       throw error;
     } finally {
-      this.inFlight = null;
+      if (this.inFlight === promise) {
+        this.inFlight = null;
+      }
     }
   }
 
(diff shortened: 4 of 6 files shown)


packages/ui/src/features/updates/WhatsNewModal.tsx
```diff
@@ -42,13 +42,15 @@ export function WhatsNewModal() {
...
-    enabled: isOpen,
-  });
   const { data: currentVersion } = useQuery(
     hostTRPC.os.getAppVersion.queryOptions(),
   );
+  const { data, isLoading, isError } = useQuery({
+    ...hostTRPC.githubReleases.list.queryOptions(
+      currentVersion ? { expectVersion: currentVersion } : undefined,
+    ),
+    enabled: isOpen && currentVersion !== undefined,
```
pauldambra: > [!NOTE]

**[qa-team/frontend]** 🟡 MEDIUM

Gating with `enabled: isOpen && currentVersion !== undefined` breaks the loading state this PR is about: in TanStack Query v5 a disabled query has `isLoading === false` (it is pending but not fetching), so while `getAppVersion` is resolving — or indefinitely if it errors — the modal skips the `isLoading` and `isError` branches and renders "No releases found." instead of the skeleton. Use `isPending` instead of `isLoading` for the skeleton branch (or fold the version query's pending state into the loading check), and consider what the user should see if `getAppVersion` fails, since today the releases query stays disabled forever with no error surfaced. (Flagged independently by three qa-team reviewers — convergent finding.)

packages/ui/src/features/updates/WhatsNewModal.tsx
```diff
@@ -42,13 +42,15 @@ export function WhatsNewModal() {
...
-    ...hostTRPC.githubReleases.list.queryOptions(),
-    enabled: isOpen,
-  });
   const { data: currentVersion } = useQuery(
     hostTRPC.os.getAppVersion.queryOptions(),
   );
+  const { data, isLoading, isError } = useQuery({
+    ...hostTRPC.githubReleases.list.queryOptions(
+      currentVersion ? { expectVersion: currentVersion } : undefined,
+    ),
```
pauldambra, line 55: > [!NOTE]

The two modals now handle "version not yet known" differently: here the query is blocked until `currentVersion` resolves (good — avoids a wasted version-agnostic fetch and an `undefined`-keyed cache entry), while `UpdateAvailableModal` stays enabled and fires with `undefined` input when `targetVersion` is missing, keying a separate cache entry. Both shapes are defensible given the modals differ, but a reader comparing them will wonder if the divergence is intentional — worth aligning or a word in the PR description.

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -4,14 +4,22 @@ import { githubReleasesApiResponse, type ListReleasesOutput } from "./schemas";
 const RELEASES_URL =
   "https://api.github.com/repos/PostHog/code/releases?per_page=30";
 const CACHE_TTL_MS = 10 * 60_000;
+const MISSING_VERSION_RETRY_MS = 60_000;
 const FETCH_TIMEOUT_MS = 10_000;
 
 @injectable()
 export class GitHubReleasesService {
   private cache: { fetchedAt: number; data: ListReleasesOutput } | null = null;
+  private missingVersionRefetchNotBefore = 0;
```
pauldambra, line 13: > [!NOTE]

**[convergent: qa-team/reliability + xp]** 🟢 LOW

`missingVersionRefetchNotBefore` is a single scalar shared across all expected versions, and it is only ever armed, never reset: a cooldown set while probing version A also suppresses the first refetch for version B for up to 60s, and a later successful fetch that does contain the version leaves the stale timestamp behind. Harmless today because `cacheContains` short-circuits before the cooldown term — but that invariant ("cooldown only matters while the version is missing") is unstated and would bite a future edit to the guard order. Either key the cooldown by version, reset it when the version appears, or state the invariant in a comment.

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -4,14 +4,22 @@ import { githubReleasesApiResponse, type ListReleasesOutput } from "./schemas";
...
 
-  async listReleases(): Promise<ListReleasesOutput> {
-    if (this.cache && Date.now() - this.cache.fetchedAt < CACHE_TTL_MS) {
+  async listReleases(expectVersion?: string): Promise<ListReleasesOutput> {
```
pauldambra, line 16: > [!NOTE]

**[convergent: qa-team/reliability + paul]** 🟢 LOW

Two related gaps on the missing-version refetch path, both about GitHub's unauthenticated 60 req/hr/IP limit. (1) No in-flight dedupe: the two modals now use different renderer query keys (`{expectVersion: targetVersion}` vs `{expectVersion: currentVersion}`), so concurrent opens against a cold or version-missing cache fire two parallel GitHub fetches — memoizing the in-flight promise would share one fetch. (2) No observability: when a version can never be found (a leaked/dev version string), we quietly re-poll GitHub every 60s within each TTL window with no signal — a log line on "refetched and version still missing" would let us spot it.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments and the summary comment.

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -4,14 +4,22 @@ import { githubReleasesApiResponse, type ListReleasesOutput } from "./schemas";
...
+    if (
```
pauldambra: > [!NOTE]

**[xp]** 🟡 MEDIUM

The cache-hit guard now expresses three distinct ideas in one boolean pileup: "cache is fresh", "cache has what you asked for", and "we're inside a don't-bother-refetching cooldown". Consider extracting a named predicate — `private canServeFromCache(expectVersion?: string)` — so `listReleases` reads as `if (this.canServeFromCache(expectVersion)) return this.cache.data;`. Same logic, but the intent is stated once and named, and the two `Date.now()` reads live next to the thing they gate. ComposedMethod: keep the top-level method at one level of abstraction.

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -41,12 +49,31 @@ export class GitHubReleasesService {
 
       const data: ListReleasesOutput = { releases };
       this.cache = { fetchedAt: Date.now(), data };
+      this.updateMissingVersionCooldown(expectVersion);
       return data;
     } catch (error) {
```
pauldambra, line 34: > [!NOTE]

**[paul]** 🟢 LOW

In the catch branch the cooldown is armed off `Date.now()` at failure time while the cache keeps its original `fetchedAt` — so once the 10-min TTL expires, the next call retries regardless of the 60s cooldown (the TTL check short-circuits first). I think that's the behaviour you want (TTL is the real ceiling), but it means the cooldown only ever throttles *within* a still-valid TTL window. Worth a one-line comment so the future traveller doesn't think the two timers compose differently than they do.

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -41,12 +49,31 @@ export class GitHubReleasesService {
...
       if (this.cache) {
+        this.updateMissingVersionCooldown(expectVersion);
         return this.cache.data;
       }
       throw error;
     }
   }
+
+  private cacheContains(version: string): boolean {
+    return (
```
pauldambra, line 98: > [!NOTE]

**[qa-team/compatibility]** 🟢 LOW

`cacheContains` compares `expectVersion` verbatim against `release.version`, which has already had its `v` prefix stripped from the tag name. Current callers pass unprefixed versions (`app.getVersion()`, electron-updater), but a future caller passing "v1.3.0" would never match and would silently refetch once per cooldown window with no error. Normalizing the input with the same `replace(/^v/, "")` makes the contract self-defending.

packages/ui/src/features/updates/UpdateAvailableModal.tsx
```diff
@@ -62,12 +62,14 @@ export function UpdateAvailableModal() {
   const downloadMutation = useMutation(
     hostTRPC.updates.download.mutationOptions(),
   );
+  const targetVersion = version ?? availableVersion;
   const { data: releasesData, isPending: isPendingReleases } = useQuery({
-    ...hostTRPC.githubReleases.list.queryOptions(),
+    ...hostTRPC.githubReleases.list.queryOptions(
+      targetVersion ? { expectVersion: targetVersion } : undefined,
```
pauldambra, line 68: > [!NOTE]

**[qa-team/frontend]** ⚪ NIT

If `targetVersion` resolves after the modal is already open, the query input flips from `undefined` to `{expectVersion: ...}`, which changes the TanStack Query cache key and re-runs the query, briefly re-showing the release-notes skeleton mid-modal. In practice the modal opens with a version already known, so only worth addressing if you see skeleton flicker — e.g. by keeping `placeholderData` from the previous key.

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE <sub>(round 6 @ 3e965ea)</sub>

The round-5 hardening (identity-guarded `inFlight` clear, concurrent-failure test, version-agnostic-fetch comment) plus the simplify pass (`fetchedAt` stamped at data arrival) were verified correct and complete by all three reviewers. No actionable findings remain — review has converged.

### Key findings

- ⚪ **NIT** (record only, not posted inline): the identity guard's protected interleaving (a late joiner's `finally` racing a replacement fetch) has no direct test — acceptable given the contrived setup it would need; the clear-on-failure regression *is* tested
- Timestamp asymmetry (cooldown uses caller entry `now`, `fetchedAt` uses arrival time) traced by all three reviewers: errs only in the conservative direction

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Both commits correct and complete; dedupe test not timing-flaky; nothing blocks merging. |
| 👤 paul | Traced cooldown, stale-serve, and expiry paths against the tests — all reconcile. Ship it. |
| 📐 xp | Reviewed against the four rules of simple design; comments earn their place documenting subtle invariants. No standing or new issues. |
| 🛡 security-audit | Ran round 1 (no findings); dedupe further reduces outbound request amplification. |

<details>
<summary>Previous rounds (5)</summary>

round 1 @ 2bbb66c — 💬 APPROVE WITH NITS: convergent MEDIUM loading-state issue + hardening notes; fixed in be5ee05/fd84320.
round 2 @ fd84320 — ✅ APPROVE: fixes verified; 1 LOW + 2 NITs, fixed in c693e6c.
round 3 @ c693e6c — ⚠️ REQUEST CHANGES: convergent HIGH — error branch unreachable; fixed in 91c3001.
round 4 @ 91c3001 — ✅ APPROVE: fix verified by all reviewers; converged.
round 5 @ b5c5c71 — 💬 APPROVE WITH NITS: dedupe concurrency verified sound; 2 LOW + 1 NIT hardening, fixed in 55f9430/3e965ea.

</details>

---

packages/ui/src/features/updates/WhatsNewModal.tsx
```diff
@@ -42,13 +42,15 @@ export function WhatsNewModal() {
...
   const close = useWhatsNewStore((state) => state.close);
   const hostTRPC = useHostTRPC();
-  const { data, isLoading, isError } = useQuery({
-    ...hostTRPC.githubReleases.list.queryOptions(),
-    enabled: isOpen,
-  });
   const { data: currentVersion } = useQuery(
     hostTRPC.os.getAppVersion.queryOptions(),
   );
+  const { data, isLoading, isError } = useQuery({
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `be5ee055b` — changed `isLoading` to `isPending: isLoading` in `packages/ui/src/features/updates/WhatsNewModal.tsx`.

packages/ui/src/features/updates/WhatsNewModal.tsx
pauldambra: > [!NOTE]

Fixed in `be5ee055b` — changed `isLoading` to `isPending: isLoading` in `packages/ui/src/features/updates/WhatsNewModal.tsx` so the skeleton renders while the query is in pending/idle state.

packages/ui/src/features/updates/WhatsNewModal.tsx
pauldambra: > [!NOTE]

Intentional divergence — the two modals have different lifecycle contracts. WhatsNewModal blocks its query until currentVersion resolves to avoid a version-agnostic cache entry; UpdateAvailableModal fires with the version already known from the updater. The asymmetry reflects the source difference, not an oversight.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Fixed in `be5ee055b` — added a comment in `canServeFromCache` stating the invariant: the single-scalar cooldown is safe because `cacheContains` short-circuits above it, so the cooldown is only consulted while the version is absent.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Fixed in `be5ee055b` — extracted `private canServeFromCache(expectVersion?: string): boolean` so `listReleases` reads as a single guard at one level of abstraction, with each cache-hit condition named and commented.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Fixed in `be5ee055b` — added a comment at the top of `canServeFromCache` explaining that the cooldown only ever throttles within a still-valid TTL window; once the cache expires the TTL check short-circuits first.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Fixed in `be5ee055b` — normalising `expectVersion` with `replace(/^v/, "")` at the top of `listReleases` so a "v"-prefixed version string matches the stored format.

packages/ui/src/features/updates/UpdateAvailableModal.tsx
pauldambra: > [!NOTE]

Out of scope for this PR — the modal opens only when the updater has already resolved a target version, so the flicker path is not reachable in practice. Worth revisiting if that assumption ever changes.

packages/ui/src/features/updates/WhatsNewModal.tsx
```diff
@@ -42,13 +42,19 @@ export function WhatsNewModal() {
...
   );
+  const {
+    data,
+    isPending: isLoading,
+    isError,
+  } = useQuery({
+    ...hostTRPC.githubReleases.list.queryOptions(
+      currentVersion ? { expectVersion: currentVersion } : undefined,
+    ),
+    enabled: isOpen && currentVersion !== undefined,
```
pauldambra: > [!NOTE]

**[convergent: qa-team/frontend + paul]** 🟢 LOW

With the query gated on `currentVersion !== undefined` and the skeleton driven by `isPending` (true while disabled), an error from `os.getAppVersion` now leaves the skeleton up indefinitely with no error text — the releases query never enables. Vanishingly rare for a local IPC call on desktop, but cheap to handle: destructure `isError` from the getAppVersion query and OR it into the existing "Could not load releases" branch.

packages/ui/src/features/updates/WhatsNewModal.tsx
```diff
@@ -42,13 +42,19 @@ export function WhatsNewModal() {
...
-  const { data, isLoading, isError } = useQuery({
-    ...hostTRPC.githubReleases.list.queryOptions(),
-    enabled: isOpen,
-  });
   const { data: currentVersion } = useQuery(
     hostTRPC.os.getAppVersion.queryOptions(),
   );
+  const {
+    data,
+    isPending: isLoading,
```
pauldambra: > [!NOTE]

`isPending: isLoading` aliases the v5 `isPending` back to the old name `isLoading`, which in TanStack Query v5 means something narrower ("pending AND fetching") — a future reader may assume the wrong semantics. Prefer using `isPending` directly in the JSX.

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -40,13 +45,53 @@ export class GitHubReleasesService {
         }));
 
       const data: ListReleasesOutput = { releases };
-      this.cache = { fetchedAt: Date.now(), data };
+      this.cache = { fetchedAt: now, data };
```
pauldambra: > [!NOTE]

**[qa-team/performance]** ⚪ NIT

`fetchedAt: now` (and the cooldown deadline) use the timestamp captured before the awaited fetch, so a slow fetch (up to the 10s timeout) backdates the TTL/cooldown by its duration. Negligible against a 10-minute TTL — just noting the intentional trade-off the threaded `now` makes.

pauldambra (commented):
> [!NOTE]

QA Swarm round 2 complete (re-review after round-1 fixes). Round-1 findings verified addressed; remaining inline comments are LOW/NIT.

packages/ui/src/features/updates/WhatsNewModal.tsx
pauldambra: > [!NOTE]

Fixed in `c693e6c` — destructured `isError` from `os.getAppVersion` query and OR'd it into the existing error branch. Also removed the `isPending: isLoading` alias in the same commit.

packages/ui/src/features/updates/WhatsNewModal.tsx
pauldambra: > [!NOTE]

Fixed in `c693e6c` — removed the `isPending: isLoading` alias and use `isPending` directly in the JSX.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Intentional trade-off — against a 10-minute TTL the backdate from a slow fetch is negligible and the simpler pre-fetch timestamp avoids a second `Date.now()` call after the await. Leaving as-is.

packages/ui/src/features/updates/WhatsNewModal.tsx
```diff
@@ -76,7 +83,7 @@ export function WhatsNewModal() {
           </Dialog.Close>
         </Flex>
 
-        {isLoading ? (
+        {isPending ? (
```
pauldambra: > [!NOTE]

**[convergent: qa-team/frontend + paul + xp]** 🟠 HIGH

The ternary order makes the version-error state unreachable — the exact case c693e6c was written for. When `os.getAppVersion` errors, `currentVersion` stays undefined, the releases query stays disabled, and in TanStack Query v5 a never-enabled query has `isPending === true` forever. Since `isPending` is checked before `isError`, the user gets an infinite `ChangelogSkeleton` and the `isVersionError` wiring (line 58) never renders — also a regression vs main, where the changelog rendered even without a version. Fix: check `isError` before `isPending` (`isError ? errorText : isPending ? skeleton : ...`), or gate the skeleton with `isPending && !isVersionError`.

packages/ui/src/features/updates/WhatsNewModal.tsx
```diff
@@ -42,13 +42,20 @@ export function WhatsNewModal() {
...
   );
+  const {
+    data,
+    isPending,
+    isError: isReleasesError,
+  } = useQuery({
+    ...hostTRPC.githubReleases.list.queryOptions(
+      currentVersion ? { expectVersion: currentVersion } : undefined,
+    ),
+    enabled: isOpen && currentVersion !== undefined,
```
pauldambra: > [!NOTE]

Line 54 gates the query input on truthiness (`currentVersion ? {...} : undefined`) while line 56 gates `enabled` on `!== undefined`. An empty-string version would enable the query without `expectVersion` pinning. Make the two conditions agree (e.g. `!!currentVersion` in both).

pauldambra (commented):
> [!NOTE]

QA Swarm round 3 complete (re-review of the getAppVersion-error fix). One convergent HIGH: the new error branch is unreachable.

packages/ui/src/features/updates/WhatsNewModal.tsx
pauldambra: > [!NOTE]

Fixed in `91c3001` — moved `isError` check before `isPending` in the render ternary so a version error renders the error text immediately instead of an infinite skeleton.

packages/ui/src/features/updates/WhatsNewModal.tsx
pauldambra: > [!NOTE]

Fixed in `91c3001` — aligned `enabled` to use `!!currentVersion` so both the query input and the enabled gate use the same truthiness condition, preventing an empty-string version from enabling the query without expectVersion pinning.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Dedupe fixed in `b5c5c71` — the in-flight fetch is now memoized so concurrent callers share one GitHub request (covered by a new test). The observability log line was consciously skipped per the author: it needs ROOT_LOGGER injection and the refetch cadence is already bounded (60s cooldown within a 10-min TTL, queries only fire on modal open), so it can wait until there's evidence dev-build versions are common.

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -4,49 +4,105 @@ import { githubReleasesApiResponse, type ListReleasesOutput } from "./schemas";
...
+      this.updateMissingVersionCooldown(normalizedVersion, now);
       return data;
     } catch (error) {
       if (this.cache) {
+        this.updateMissingVersionCooldown(normalizedVersion, now);
         return this.cache.data;
       }
       throw error;
+    } finally {
+      this.inFlight = null;
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + paul]** 🟢 LOW

Every waiter on the shared `inFlight` promise unconditionally runs `finally { this.inFlight = null; }`. This is safe today only because nothing awaits between `await this.inFlight` and the `finally` (catch and cooldown update are synchronous), so all waiters' finally blocks drain as consecutive microtasks before any new call can assign a fresh `inFlight`. If a future edit inserts an await on that path (e.g. an async fallback in the catch), a late waiter could null out a newer caller's in-flight promise and silently break the dedupe. Capture the promise locally and guard the clear: `const promise = (this.inFlight ??= this.fetchAndCacheReleases(now)); ... finally { if (this.inFlight === promise) this.inFlight = null; }` — correctness becomes local instead of resting on a non-obvious event-loop invariant.

packages/workspace-server/src/services/github-releases/github-releases.test.ts
```diff
@@ -69,11 +69,86 @@ describe("GitHubReleasesService", () => {
...
+    });
+    const second = await service.listReleases("1.3.0");
+    expect(second.releases[0].version).toBe("1.3.0");
+
+    const third = await service.listReleases("1.3.0");
+    expect(third).toEqual(second);
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+  });
+
+  it("dedupes concurrent cache misses into a single fetch", async () => {
```
pauldambra, line 114: > [!NOTE]

**[convergent: qa-team + xp]** 🟢 LOW

The new concurrency test covers only the success path. The concurrent *failure* path — two racing callers sharing a fetch that rejects — is exactly where the per-caller catch/finally interplay lives, and where a leaked `inFlight` would strand every future caller. Add one concurrent-failure case: `Promise.all` over two calls with `fetchMock` rejecting, asserting both reject with no cache (or both receive stale data with a warm cache), and that a subsequent call fetches again (proving `inFlight` was cleared).

packages/workspace-server/src/services/github-releases/github-releases.ts
```diff
@@ -4,49 +4,105 @@ import { githubReleasesApiResponse, type ListReleasesOutput } from "./schemas";
...
-              : release.tag_name,
-          notes: release.body ?? "",
-          date: release.published_at,
-          isPrerelease: release.prerelease,
-          htmlUrl: release.html_url,
-        }));
-
-      const data: ListReleasesOutput = { releases };
-      this.cache = { fetchedAt: Date.now(), data };
+      this.inFlight ??= this.fetchAndCacheReleases(now);
```
pauldambra: > [!NOTE]

**[convergent: xp + paul]** ⚪ NIT

The dedupe silently attaches a caller with one `expectVersion` to an in-flight fetch started for a different (or no) version. That's safe *only* because `fetchAndCacheReleases` ignores `expectVersion` and always returns the full list — a load-bearing invariant that isn't stated near this line. A one-line comment ("the fetch is version-agnostic, so any concurrent caller can share it") would protect a future maintainer tempted to parameterise the fetch by version.

pauldambra (commented):
> [!NOTE]

QA Swarm round 5 complete (review of the in-flight dedupe commit). Concurrency semantics verified sound; findings are LOW/NIT hardening.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Fixed in `55f943046` — captured the promise before the try block and guard the clear with `if (this.inFlight === promise) this.inFlight = null;` so a late waiter from an earlier fetch cannot zero out a newer caller's in-flight promise.

packages/workspace-server/src/services/github-releases/github-releases.test.ts
pauldambra: > [!NOTE]

Fixed in `55f943046` — added a concurrent-failure test: two racing callers share a rejected fetch, both reject (no cache), and a subsequent call fetches again proving inFlight was cleared.

packages/workspace-server/src/services/github-releases/github-releases.ts
pauldambra: > [!NOTE]

Fixed in `55f943046` — added the comment alongside the identity-guard refactor.

PR #3177, opened 2026-07-06
fix(skills): ignore hidden directories when walking skill files

adboio:
## Problem

A user hit `Local skill bundle contains more than 1000 files` when invoking a custom skill on a cloud run. Two issues behind it:

- The bundler walks the entire skill directory tree and its ignore list was only `.DS_Store` / `.git` / `node_modules` / `__pycache__`, so hidden directories like `.venv` (easily 5–20k files) counted toward the limit and got uploaded.
- The error named neither the offending skill nor a remedy — and the offender can be a skill pulled in via `dependencies:` that the user never tagged.

## Changes

- Skip hidden directories when walking skill files, via a shared `isIgnoredSkillEntry` predicate. This extends the existing `findSkillDirs` rule that hidden dirs are never skills. Hidden *files* (`.gitignore`) still count.
- Apply the same predicate in `listSkillFiles`, so the skill-contents UI and team-skill exports match what a cloud bundle uploads (previously that walker skipped nothing — publishing a team skill would ship a `.venv` to the skills store).
- The bundle limit errors now name the skill, its resolved path, and what to do: `Skill "alpha" (/home/u/.claude/skills/alpha) contains more than 1000 files. Cloud runs upload every file in the skill folder, so move data and build artifacts out of it.`

Left alone: the 1000-file cap itself (only the PostHog server's 30MB artifact size limit is a real platform constraint; raising/dropping the count cap can be a follow-up if legit >1000-file skills show up), and the marketplace install direction (unpacking downloaded archives stays a faithful copy).

## Testing

- New `skill-bundler.test.ts` unzips real bundles: hidden dirs and junk excluded while hidden files and nested content survive; 1100 files inside `.venv` bundle fine (the reported scenario); the limit error names skill and path.
- Matching `listSkillFiles` case in `skill-discovery.test.ts`.
- All 90 workspace-server skills tests pass; `tsc --noEmit` and Biome clean.

### DIFF
--- a/packages/workspace-server/src/services/skills/skill-bundler.test.ts
+++ b/packages/workspace-server/src/services/skills/skill-bundler.test.ts
@@ -1,0 +1,81 @@
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { unzipSync } from "fflate";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { bundleLocalSkill } from "./skill-bundler";
+
+let root: string;
+let skillPath: string;
+
+beforeEach(async () => {
+  root = await mkdtemp(path.join(tmpdir(), "skill-bundler-test-"));
+  skillPath = path.join(root, "alpha");
+  await mkdir(skillPath, { recursive: true });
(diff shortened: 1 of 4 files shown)


charlesvien:

PR #3179, opened 2026-07-06
fix(workspace): make branch mismatch "Switch branch" actually send the message

adboio:
## Problem

In the branch-mismatch dialog, the **Switch branch** button was wrapped in `AlertDialog.Action`, which auto-closes the dialog on click. Closing fires `onOpenChange(false)` → `onCancel()`, which:

- discarded the pending message + clear-editor refs while the checkout mutation was still in flight, so **the user's message was never sent after a successful switch** — it silently stayed in the composer;
- meant the `loading` state on the button was never visible (dialog already gone);
- rendered checkout **errors into an already-closed dialog** — failed switches were completely silent;
- double-fired analytics: every `switch` also emitted a `cancel`.

The double-fire is visible in production: over the last 60 days, `Branch mismatch warning shown` fired 1,249× while actions total 1,513 (continue 737 / cancel 505 / switch 271). The 264-event excess matches the 271 switches almost exactly — corrected, the real distribution is ~59% continue / ~22% switch / ~19% cancel.

## Fix

- Unwrap the Switch button from `AlertDialog.Action` so clicking it no longer closes the dialog; it now stays open through the mutation (spinner visible, errors surface inline) and closes via state when the checkout succeeds.
- Ignore `onOpenChange` close requests while `isSwitching`, so Escape can't wipe the pending message mid-checkout.
- Clear a stale `switchError` when a retry succeeds.

## Tests

New `BranchMismatchDialog.test.tsx` component test clicks the real Radix buttons and asserts `onSwitch` doesn't co-fire `onCancel` (the regression), plus coverage for Cancel/Continue/Escape/error states. Existing hook tests unchanged and passing.

Part of a stack reworking the branch-mismatch UX:
1. **→ this PR** — fix the broken Switch action
2. Auto-unlink stale linked branches
3. Replace the modal with a non-blocking pre-send banner
4. Staff-gated auto-switch for clean worktree tasks

### DIFF
--- a/packages/ui/src/features/task-detail/BranchMismatchDialog.test.tsx
+++ b/packages/ui/src/features/task-detail/BranchMismatchDialog.test.tsx
@@ -1,0 +1,90 @@
+import { Theme } from "@radix-ui/themes";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, expect, it, vi } from "vitest";
+import { BranchMismatchDialog } from "./BranchMismatchDialog";
+
+function renderDialog(
+  overrides?: Partial<Parameters<typeof BranchMismatchDialog>[0]>,
+) {
+  const handlers = {
+    onSwitch: vi.fn(),
+    onContinue: vi.fn(),
+    onCancel: vi.fn(),
+  };
+  render(
+    <Theme>
+      <BranchMismatchDialog
+        open
+        linkedBranch="feat/foo"
+        currentBranch="main"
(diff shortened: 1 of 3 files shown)


adboio:
Closing unmerged: the branch-mismatch dialog this fixes is deleted wholesale by #3183 (the modal → banner replacement), so merging this would only add review load for code that dies two PRs up the stack. The stack is now #3180 → #3183, rebased so this PR's commits are no longer part of it. If the banner direction gets dropped, this fix stands alone and can be reopened.

PR #3181, opened 2026-07-06
feat(browser): in-app browser tab with webview security hardening

MattPua:
## Problem

Working on web changes means bouncing between the app and an external browser. There is no way to view docs, a local dev server, or another page inside the app next to a task.

## Changes

https://github.com/user-attachments/assets/0a6b4ef1-62ac-48e3-bfba-288e23f7e44b

Adds an embedded browser tab, gated behind `posthog-code-browser-tab` and enabled by default in local development.

- Adds a host-injected browser capability so unsupported hosts never mount Electron's `<webview>`.
- Adds address, back, forward, reload, stop, loading, title, and persisted URL handling.
- Normalizes public HTTP addresses to HTTPS while preserving HTTP for loopback development servers.
- Keeps restored browser tabs visible with an unavailable state when the host or flag does not support them.
- Waits for Electron `dom-ready`, surfaces attachment failures, and queues early Enter navigation until the webview is ready.
- Adds a main-process feature gate synchronized with the renderer flag.
- Hardens webview attachment preferences and uses a dedicated persisted partition.
- Denies all guest permissions through an explicit empty allowlist.
- Allows only HTTPS, loopback HTTP, or exactly `about:blank` for main-frame navigation.
- Applies the same URL policy to popups and blocks unsafe programmatic requests, custom schemes, files, and metadata endpoints.

## How did you test this?

- Browser normalization, interaction, availability, permission, attachment, navigation, request, and feature-gate unit tests.
- Desktop and UI typechecks, Biome, and host-boundary validation.
- Packaged Electron E2E covering upgraded webview APIs, navigation, unsafe URL rejection, reload, and recovery.
- Real development app over CDP, including immediate Enter navigation on a brand-new browser tab before `dom-ready`.

## Automatic notifications

- [x] Publish to changelog?

---


packages/core/src/panels/panelLayoutTransforms.test.ts
```diff
@@ -75,6 +77,52 @@ describe("panelLayoutTransforms", () => {
...
   });
 
+  describe("addBrowserTab", () => {
+    it("adds a browser tab carrying the initial url", () => {
+      const layout = createInitialTaskLayout();
+      const next = applyUpdates(
+        layout,
+        addBrowserTab(layout, "main-panel", "https://posthog.com"),
+      );
+
```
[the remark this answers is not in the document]
MattPua: Fixed — added `expect(next.panelTree.type).toBe("leaf")` before the type-narrowing guard in both tests. (PostHog Code, on behalf of the author)

packages/core/src/panels/panelLayoutTransforms.test.ts
```diff
@@ -75,6 +77,52 @@ describe("panelLayoutTransforms", () => {
...
+  describe("updateBrowserTabUrl", () => {
+    it("updates the url of an existing browser tab", () => {
+      const layout = createInitialTaskLayout();
+      const added = applyUpdates(
+        layout,
+        addBrowserTab(layout, "main-panel", "about:blank"),
+      );
+      if (added.panelTree.type !== "leaf") return;
+      const tabId = added.panelTree.content.tabs.find(
```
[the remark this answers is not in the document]
MattPua: Fixed alongside the sibling comment — explicit assertion added before the narrow. (PostHog Code, on behalf of the author)

MattPua:
Addressed the review feedback (PostHog Code, on behalf of the author):

- Both P1 silent test bail-outs: explicit `expect(...).toBe("leaf")` assertions added before the type-narrowing guards
- P2 (session handlers reset on every attach): permission handlers now install once per session via a `WeakSet` guard (`hardenWebviewSession`), so a future per-guest divergence can't silently replace an earlier handler
- Rebased onto latest main — resolved conflicts with the new `autoresearch` tab type, conflicts are cleared

packages/ui/src/features/browser/BrowserPanel.tsx
```diff
@@ -0,0 +1,313 @@
...
+    <button
+      type="button"
+      aria-label={ariaLabel}
+      data-attr={dataAttr}
+      onClick={onClick}
+      disabled={disabled}
+      className="flex h-[24px] w-[24px] cursor-pointer items-center justify-center rounded-(--radius-2) border-0 bg-transparent text-(--gray-11) hover:bg-(--gray-4) disabled:cursor-default disabled:opacity-40"
+    >
+      {children}
+    </button>
```
MattPua: does it not make sense to use a quill button here?

packages/ui/src/features/browser/BrowserPanel.tsx
```diff
@@ -0,0 +1,313 @@
...
+            <Globe size={24} className="text-gray-10" />
+            <Text color="gray" className="text-sm">
+              {loadError}
+            </Text>
+          </Flex>
+        )}
+        {/* Shared persisted profile across all browser tabs/tasks is intentional
+            (stay logged in to e.g. GitHub); trade-off: shared cookies/storage.
+            No `allowpopups` — popups are denied and routed to the OS browser by
+            the guest's window-open handler (window.ts). */}
```
MattPua: could we adda. button to the browser tabs to quicly clear cache?

packages/ui/src/features/browser/BrowserPanel.tsx
```diff
@@ -0,0 +1,313 @@
...
+          >
+            <Globe size={24} className="text-gray-10" />
+            <Text color="gray" className="text-sm">
+              {loadError}
+            </Text>
+          </Flex>
+        )}
+        {/* Shared persisted profile across all browser tabs/tasks is intentional
+            (stay logged in to e.g. GitHub); trade-off: shared cookies/storage.
+            No `allowpopups` — popups are denied and routed to the OS browser by
```
MattPua: why do we deny popups?

packages/ui/src/features/browser/BrowserPanel.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+    return () => {
+      webview.removeEventListener("did-navigate", onNavigate);
+      webview.removeEventListener("did-navigate-in-page", onNavigate);
+      webview.removeEventListener("page-title-updated", onTitle);
+      webview.removeEventListener("did-fail-load", onFailLoad);
+      webview.removeEventListener("did-start-loading", onStartLoading);
+      webview.removeEventListener("did-stop-loading", onStopLoading);
+    };
+  }, [persistUrl]);
```
MattPua: was a useEffect necsesary?

apps/code/src/main/window.ts
```diff
@@ -120,6 +120,116 @@ function setupExternalLinkHandlers(window: BrowserWindow): void {
   });
 }
 
+// The authoritative gate for the in-app browser guest: main process, where a
+// guest page can't route around it. The renderer's normalizeAddress is only a
+// convenience on top of this.
+const ALLOWED_WEBVIEW_SCHEMES = new Set(["http:", "https:", "about:"]);
```
MattPua: why did we need about?

apps/code/src/main/window.ts
```diff
@@ -120,6 +120,116 @@ function setupExternalLinkHandlers(window: BrowserWindow): void {
...
+
+// The link-local range (incl. cloud metadata 169.254.169.254) can hand out
+// instance credentials. Loopback and LAN are deliberately allowed — reaching a
+// local dev server is a first-class use of a coding tool's browser.
+function isBlockedWebviewHost(hostname: string): boolean {
+  return /^169\.254\./.test(hostname);
+}
```
MattPua: not entirely clear from the comment why this is needed

apps/code/src/main/window.ts
```diff
@@ -120,6 +120,116 @@ function setupExternalLinkHandlers(window: BrowserWindow): void {
...
+  "usb",
+  "pointerLock",
+  "idle-detection",
+  "openExternal", // popups are already routed through our own handler
+]);
+
+// setPermissionRequestHandler replaces (not composes with) any previous
+// handler on the session, and guests share one persisted session — install
+// once per session so a future per-guest divergence can't silently drop an
+// earlier handler.
```
MattPua: not super clear

apps/code/src/main/window.ts
```diff
@@ -120,6 +120,116 @@ function setupExternalLinkHandlers(window: BrowserWindow): void {
...
+
+  // Deny at both request time (prompts) and check time (sync fast-paths like
+  // navigator.permissions.query).
+  session.setPermissionRequestHandler((_wc, permission, callback) => {
+    callback(!DENIED_WEBVIEW_PERMISSIONS.has(permission));
+  });
+  session.setPermissionCheckHandler(
+    (_wc, permission) => !DENIED_WEBVIEW_PERMISSIONS.has(permission),
+  );
+}
```
MattPua, line 157: what is this doing

packages/ui/src/features/browser/BrowserPanel.tsx
MattPua: Done in d850a44 — replaced the hand-rolled NavButton with quill `Button size="icon-sm"` (same pattern as TabStrip). (PostHog Code, on behalf of the author)

packages/ui/src/features/browser/BrowserPanel.tsx
MattPua: Deferred for now per Matt — clearing the shared `persist:browser` session needs a main-process hop (platform interface + host-router procedure), so holding off until we want it. (PostHog Code, on behalf of the author)

packages/ui/src/features/browser/BrowserPanel.tsx
MattPua: A popup would be a brand-new native child window hosting untrusted web content — it would live outside the webview, so none of the hardening here (nav guards, permission denials, preload stripping) applies to it, and it would look like a legit app window for spoofing/phishing. Denying `allowpopups` and routing http(s) targets to the OS browser keeps arbitrary web content confined to the sandboxed webview. (PostHog Code, on behalf of the author)

packages/ui/src/features/browser/BrowserPanel.tsx
MattPua: Yes — Electron `<webview>` emits plain DOM events (`did-navigate`, `page-title-updated`, ...) on the element, which React has no declarative props for. The element only exists after mount (via ref), so `addEventListener` in an effect with cleanup on unmount is the standard way to wire these. (PostHog Code, on behalf of the author)

apps/code/src/main/window.ts
MattPua: `about:` is needed for `about:blank` — the src a new blank browser tab mounts with before a url is entered. Without it the guard would veto the initial load of every new tab. Expanded the comment in d850a44. (PostHog Code, on behalf of the author)

apps/code/src/main/window.ts
MattPua: Reworded in d850a44: this blocks 169.254.0.0/16 because on cloud VMs 169.254.169.254 is the instance-metadata endpoint, which hands IAM/service-account credentials to any local HTTP request — a hostile page could redirect the webview there to exfiltrate them (classic SSRF target). Loopback/LAN stay allowed for local dev servers. (PostHog Code, on behalf of the author)

apps/code/src/main/window.ts
MattPua: Clarified in d850a44: all browser webviews share one `persist:browser` session, and `setPermissionRequestHandler` replaces (not stacks) the previous handler — so installing on every attach would let the latest attach silently win. The WeakSet makes it install once per session. (PostHog Code, on behalf of the author)

apps/code/src/main/window.ts
MattPua: Comment expanded in d850a44: Chromium has two permission paths — `setPermissionRequestHandler` decides explicit requests (prompt-style, e.g. getUserMedia), `setPermissionCheckHandler` answers sync status probes (`navigator.permissions.query`). Both deny the same list so a page can't see "granted" on one path while the other refuses. (PostHog Code, on behalf of the author)

charlesvien:
Did we implement these already in https://github.com/PostHog/code/pull/3221?

MattPua:
@charlesvien i believe those are "tab implementations" and this is actually running the browser

adboio (dismissed):
(sry agent went rogue)

charlesvien:

PR #3205, opened 2026-07-06
feat(autoresearch): improve new-task composer controls (GROW-124)

fercgomes:
## What

Polish pass on the **autoresearch** mode in the new-task composer (follow-up to the autoresearch feature, GROW-103).

### 1. Autoresearch moved into the agent-mode dropdown
- Removed the standalone "Autoresearch" toolbar button.
- It's now the last item of the mode dropdown (after a separator), as a checkbox toggle applied on menu close — the same deferral the mode radio items use, so the composer doesn't relayout under the closing menu.
- Arming it defaults the agent mode to the most hands-off option available — `bypassPermissions` when the user has it enabled, otherwise `acceptEdits` — since an unattended iteration loop would stall on permission prompts.

### 2. Fixed the target/iterations inputs stealing focus on click
- Root cause: they lived inside the composer's `InputGroup`, whose click handler refocuses the editor on any non-button click, so clicking an input bounced focus straight to the prompt.
- Fixed structurally by relocating them (see #3) into the outer container, which already respects focusable elements.

### 3. Controls moved outside the input box
- The autoresearch controls now render as a bar **between the dropdown row and the prompt input**, instead of inside the composer's header addon.

### 4. Controls rewritten to be self-explanatory
- Before: a bare `Maximize`, an unlabeled `Target` box, a stray `≤ 10 iterations`, and a `Claude Opus 4.8 · Max` line that didn't look clickable.
- After: reads as one sentence — **Optimize to `maximize` until it reaches `[optional]` or after `[10]` iterations using `[model]`** — with a tooltip on each label and an explainer on the Autoresearch chip. Tooltips wrap only plain label spans (never the inputs) to avoid reintroducing focus flicker.

## Feature-flag gating
Both new surfaces stay behind `useAutoresearchEnabled()` (staff-gated flag; always on in dev):
- The dropdown toggle is passed to `PromptInput`/`ModeSelector` only when `autoresearchService && autoresearchEnabled`.
- The controls bar renders only when `autoresearchDraft` is set, which is forced `null` while the flag is off.
- `ModeSelector` and `PromptInput` stay flag-agnostic; the gate lives solely in `TaskInput`.

## Files
- `packages/ui/src/features/message-editor/components/ModeSelector.tsx`
- `packages/ui/src/features/message-editor/components/PromptInput.tsx`
- `packages/ui/src/features/task-detail/components/TaskInput.tsx`
- `packages/ui/src/features/autoresearch/AutoresearchComposerControls.tsx`

## Testing
- `pnpm typecheck` (full monorepo, 22/22) and Biome lint pass on all changed files.
- ⚠️ **Not yet driven in the running app** — live check of the focus fix, the dropdown toggle applying the hands-off mode, and tooltip/label wrapping still pending.

Fixes GROW-124

---

### DIFF
--- a/packages/shared/src/analytics-events.ts
+++ b/packages/shared/src/analytics-events.ts
@@ -993,6 +993,28 @@ export interface ClaudeSessionImportFailedProperties {
   failed_step?: string;
 }
 
+/** Fired when a user arms autoresearch mode on the new-task composer. */
+export interface AutoresearchArmedProperties {
+  /** Hands-off mode auto-applied on arm so the unattended loop isn't blocked on permission prompts. */
+  default_mode: "bypassPermissions" | "acceptEdits";
+  workspace_mode?: "local" | "worktree" | "cloud";
+}
+
+/** Fired when an armed autoresearch task is submitted and its run kicks off. */
+export interface AutoresearchRunStartedProperties {
+  direction: "maximize" | "minimize";
+  /** Whether the user set a target metric value to stop early at. */
+  has_target: boolean;
+  max_iterations: number;
+  /** Build and measure stages differ, so each iteration splits into a build turn and a measure turn. */
+  stages_split: boolean;
+  implement_model?: string;
+  measure_model?: string;
+  implement_effort?: string;
+  measure_effort?: string;
+  workspace_mode?: "local" | "worktree" | "cloud";
+}
+
 // Event names as constants
 export const ANALYTICS_EVENTS = {
   // App lifecycle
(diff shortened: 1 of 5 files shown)


fercgomes:
Added product-analytics instrumentation for autoresearch usage (commit 4319801).

**Two events** (defined in `packages/shared/src/analytics-events.ts`, fired from `TaskInput.tsx`):

- **`Autoresearch armed`** — fired when the user turns the mode on from the agent-mode dropdown.
  - `default_mode` (`bypassPermissions` | `acceptEdits`), `workspace_mode`
- **`Autoresearch run started`** — fired when an armed task is submitted and the run kicks off.
  - `direction`, `has_target`, `max_iterations`, `stages_split`, `implement_model`, `measure_model`, `implement_effort`, `measure_effort`, `workspace_mode`

Together they give an arm → run funnel. Properties are compile-time-checked via `EventPropertyMap`.

Both fire only through the feature-flag-gated arm/submit paths, so nothing emits when autoresearch is disabled.

rafaeelaudibert (approved):

PR #3221, opened 2026-07-07
Merge bluebird/code + Browser tabs – bulletproof persistence, skeleton-first navigation, bounded memory

adamleithp:
Migrating browser tabs out of bluebird into the merged Code app surfaced persistence bugs and sluggish tab switching. This branch fixes them, with regression tests and live CDP profiling behind each change.

## (fix) Tab persistence — nav clicks opened duplicate tabs, switches overwrote other tabs
- **Stale renderer mirror mis-targeted writes.** The route→tab effect makes persistent writes (`setTabTarget`/`openOrFocus`) from the renderer mirror, which lagged the server by an IPC round-trip. A navigation landing in that window wrote to the *previous* active tab (tab contents "copied over") or saw no active tab and opened a duplicate. Every mutation now applies its returned snapshot to the mirror synchronously, and activate/replace also apply optimistically before the round-trip.
- **Dead history tags dangled focus.** Back/forward can replay an entry tagged with a since-closed tab id; persisting it left a dangling `activeTabId`, after which every nav looked like "no active tab" and opened new tabs. Guarded at three layers: `decideTabNavigation` validates tags against live tab ids, a new `setWindowActiveTab` shared primitive rejects unknown/foreign tab ids, and the service refuses to persist them.
- **appView as a tab identity dimension** (home, inbox, agents, skills, MCP servers, command center) so top-level pages dedup and navigate in place instead of spawning tabs; navigation is exhaustiveness-checked so a new page kind can't silently no-op. The incomplete inline blank-tab check in `__root.tsx` (missed `channelId`/`appView`) is replaced with the shared hook.
- 29 new browser-tabs tests, including a regression suite reproducing each reported symptom end-to-end.

## (perf) Tab switching — click to first paint went from ~570ms of frozen UI to a next-frame skeleton
- Profiled over CDP: loaders were synchronous, so the router committed navigation inside one main-thread block while the destination's heavy mount (chat-thread measurement effects, canvas grid) ran — the old route stayed frozen the whole time.
- New `yieldToPaint()` single-frame loader + per-route-kind `pendingComponent` skeletons (task detail, canvas, channel pages, app pages), applied via a `withRouteSkeleton()` factory. The click now commits the URL, moves the tab highlight, and paints a skeleton before the heavy mount runs behind it.
- Router: `defaultPendingMinMs: 0` (no artificial skeleton hold) and `defaultPreloadStaleTime: 0` (hover-preload can't satisfy the loader and silently reintroduce the freeze).

## (perf) Memory — bounded per-visit caches
- **Freeform canvas threads** (up to 50 whole-document version copies each) capped at 8 MRU; eviction skips threads with saves in flight and lives in store state (devtools/HMR-safe). `revert()` now guards against evicted threads — previously it could persist an empty canvas over real content.
- **File-content queries** (full bodies + base64 blobs, `staleTime: Infinity`) bounded to 30s gcTime after unmount instead of the 5-minute default.
- **Terminal parking bounding was attempted and reverted.** An LRU cap on parked xterm instances (and a follow-on `display:none` parking tweak) froze the app when leaving a tab whose terminal was on screen — `term.dispose()` on switch synchronously re-rendered/serialized. Reverted to main's terminal behavior; a future attempt needs a teardown path that doesn't do synchronous work on switch.

## (migration)
- `browser_tabs.app_view` column (migration 0018) backing the new tab identity dimension.
- One-time localStorage adoption of the legacy Code sidebar width (`sidebar-storage`) into the unified `channels-sidebar` store, so users' resized width survives the merge instead of resetting to 240px.

## (ui) Sidebar-merge regressions restored
- Title-bar-left width transition matched to the sidebar's curve again, suppressed during drag (it snapped while the sidebar animated).
- Legacy `enter_space`/`leave_space` analytics re-fired from the Channels toggle (its merged-world successor) so space-adoption dashboards stay continuous.
- Skeletons use `@posthog/quill` primitives per the design-system rule.

## Verification
- Full shared + ui test suites pass; typecheck + biome clean across shared, ui, workspace-server.
- Live app driven over CDP throughout: frame-level screencasts confirm the skeleton paints mid-switch; the terminal-freeze regression was caught and reverted the same way.

---


adamleithp:
@k11kirky pushed a rebase + a few fixes on this branch:

**Rebased onto `main`** (was 45 behind) and resolved the conflicts:
- **Migration index collision** — both branches minted `0018`. Kept main's `0018_add_pr_urls`, renumbered ours to `0019_add_app_view`, and rebuilt `0019_snapshot.json` on top of main's 0018 (so it carries *both* `pr_urls` + `app_view`), with the prevId chain + journal fixed.
- **`sidebarStore`** — kept both sides' independent fields (`taskTypeFilter` from main, `channelsEnabled` from us).
- **`ChannelsSidebar`** — took our unified Bluebird chrome. Main had added the **Activity** nav item into the old `ChannelsNav`, which we'd replaced with `SidebarNavSection` — so that was the "missing activity". Ported it back in as a new `ActivityItem` (bell + unread-mentions badge), sitting **under the Channels toggle** and gated on `channelsEnabled` (only shows when channels are on).

**New-tab / last-tab-close default** (`BrowserTabStrip`): both were routing through the `/website` index → `channels[0]`, so a fresh tab or an emptied strip opened a random first channel. Unified them through one `landOnDefault()` keyed off the channels toggle (not the current route, which lags a flip):
- channels **on** → the private **#me** channel (provisioned lazily)
- channels **off** → the Code **new-task** screen
- never touches the channels index, so no more random-channel landing.

Typecheck + biome green across `ui` / `core` / `workspace-server`. Heads-up: this was a **force-push** (history rewritten by the rebase).

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
```diff
@@ -101,21 +186,43 @@ export function BrowserTabStrip() {
...
-    trpc.browserTabs.setActiveTab.mutationOptions(),
-  );
+  // Every tab mutation returns the fresh authoritative snapshot. Apply it to
+  // the renderer mirror synchronously: the snapshot-change subscription also
+  // delivers it, but only after an IPC round-trip, and the navigation effect
+  // below makes *persistent writes* (setTabTarget/openOrFocus) from the mirror.
+  // A stale mirror mis-targets those writes — the classic symptom is a
+  // navigation replacing some other tab's contents, or opening a duplicate tab
+  // because the mirror still says "no active tab".
+  const applySnapshot = (next: TabsSnapshot) =>
```
k11kirky: Every tab mutation's `onSuccess` unconditionally calls `applySnapshot` → `setSnapshot`, which does a bare `set({ snapshot })` with no equality check. Since the "activate"/"replace" cases below already write the mirror optimistically (lines ~338-341, ~357-360) before firing the mutation, every tab switch/navigation double-writes and double-renders every `useTabsSnapshot()` subscriber — `setTabTarget` always stamps a fresh `lastActiveAt` via `input.now()`, so the server-echoed snapshot here is never reference- or value-equal to the optimistic write.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
```diff
@@ -194,34 +317,59 @@ export function BrowserTabStrip() {
...
+        const store = browserTabsStore.getState();
+        store.setSnapshot(
+          setWindowActiveTab(store.snapshot, windowId, decision.tabId),
+        );
         setActiveTab.mutate({ windowId, tabId: decision.tabId });
         break;
-      case "replace":
-        setTabTarget.mutate({
+      }
+      case "replace": {
```
k11kirky, line 363: The "activate" and "replace" cases hand-inline an identical optimistic-mirror-write pattern (read `store.getState().snapshot`, apply a pure transform, `setSnapshot`) with duplicated safety comments, and no shared helper enforces this for future mutation types. A future mutation added to this switch would need this pattern hand-copied a third time from memory.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
```diff
@@ -237,7 +385,11 @@ export function BrowserTabStrip() {
     params.dashboardId,
     params.taskId,
     routeChannelSection,
+    routeAppView,
     activeTab,
+    // The tab LIST feeds windowTabIds (dead-tag validation); activeTab alone
+    // doesn't change when an inactive tab closes in another window.
+    snapshot,
```
k11kirky: This effect depends on the `snapshot` it itself writes optimistically (lines ~338-341, ~357-360), so it re-runs and recomputes `decideTabNavigation` 2-3x per navigation instead of once. Each extra pass resolves to a cheap `stamp`/`noop` (not a correctness bug), but it's wasted recomputation on the exact hot path this PR is otherwise optimizing for paint latency.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
```diff
@@ -473,6 +671,71 @@ export function BrowserTabStrip() {
...
+    if (!channelsEnabled) {
+      navigate({ to: "/code", state });
+      return;
+    }
+    // #me is provisioned lazily the first time (same bridge the sidebar's #me
+    // row uses); fall back to the new-task screen if it can't be created.
+    void (async () => {
+      try {
+        const existing = channels.find((c) => c.name === PERSONAL_CHANNEL_NAME);
+        const folder = existing ?? (await createChannel(PERSONAL_CHANNEL_NAME));
```
k11kirky: `landOnDefault()`'s `#me` provisioning (`channels.find(...) ?? await createChannel(...)`) is an unguarded check-then-create, same pattern as `ChannelsList.tsx`'s `PersonalChannelRow.open()`. The `/desktop_file_system/` folder-creation endpoint this hits doesn't appear to have server-side dedup (unlike the separate backend `task_channels` "me" channel, which is explicitly idempotent server-side) — concurrent invocations before the first mutation's cache update lands could both see no existing "me" folder and both create one. Wiring this into the Cmd+T new-tab hotkey (trivially double-fired/held) raises the odds of hitting this versus the original single click-handler call site. Also: there's no test coverage for this new `landOnDefault`/`applyCloseResult` logic (no test file exists for `BrowserTabStrip.tsx`), despite this PR's own commit history emphasizing regression tests for each fixed bug.

packages/ui/src/features/sidebar/components/SidebarNavSection.tsx
```diff
@@ -170,6 +187,56 @@ export function SidebarNavSection({
...
+              setChannelsEnabled(checked);
+              track(ANALYTICS_EVENTS.CHANNEL_ACTION, {
+                action_type: "toggle_channels",
+                surface: "nav",
+              });
+              // The unified sidebar removed the Code↔Channels space boundary;
+              // this toggle is its successor. Keep firing the legacy
+              // enter/leave events so space-adoption dashboards stay continuous.
+              track(ANALYTICS_EVENTS.CHANNEL_ACTION, {
+                action_type: checked ? "enter_space" : "leave_space",
```
k11kirky, line 221: The merged Channels toggle re-fires legacy `enter_space`/`leave_space` analytics with only `surface: "nav"`, silently dropping the `title_bar` and `header_button` surface values the old Exit button and `BluebirdButton` used to send (both removed by this PR). Any dashboard/breakdown by `surface` will see `title_bar`/`header_button` stop appearing entirely after this ships — a shape-change in historical analytics data, not quite the "stay continuous" the comment above claims.

packages/ui/src/features/sidebar/components/ProjectSwitcher.tsx
```diff
@@ -214,21 +214,14 @@ export function ProjectSwitcher({
...
         side="bottom"
-        className={
-          // The `button` trigger spans the full sidebar width, so binding the
-          // menu to `--anchor-width` would stretch it and break the layout. Pin
-          // a fixed width there; the `item` trigger matches its anchor.
-          triggerVariant === "button"
-            ? "w-64 min-w-64 pt-0"
-            : "w-(--anchor-width) max-w-(--anchor-width) pt-0"
-        }
+        className="w-(--anchor-width) max-w-(--anchor-width) pt-0"
```
k11kirky, line 196: Now that the `triggerVariant === "button"` branch of this ternary is gone, the `triggerVariant="button"` variant of `ProjectSwitcher` (prop default at line 65, JSX branch at ~line 180-192) has no remaining caller — the only caller (`ChannelsSidebar.tsx`) renders `<ProjectSwitcher />` with no props. Worth deleting the dead branch/prop rather than leaving it to rot.

packages/ui/src/router/routes/__root.tsx
```diff
@@ -131,14 +140,12 @@ function RootLayout() {
...
-  // perform their navigation only once it's submitted or skipped.
+  // "PostHog Web" opens the feedback modal first and performs its navigation
+  // only once the modal is submitted or skipped.
   const handleFeedbackFinished = () => {
     const finishedMode = feedbackMode;
     setFeedbackMode(null);
-    if (finishedMode === "leaving") {
-      navigate({ to: "/code" });
-    } else if (finishedMode === "posthog-web" && posthogWebUrl) {
+    if (finishedMode === "posthog-web" && posthogWebUrl) {
```
k11kirky, line 148: `FeedbackModalMode`'s `"leaving"` mode (in `FeedbackModal.tsx`, untouched by this PR) is now unreachable dead code — this PR removes the only call site that ever set it (the old "Exit" button in the now-deleted `isChannelsSpace` branch). `FeedbackModal.tsx`/`FeedbackModal.test.tsx` still fully implement and test it, so it'll silently rot with no signal that it can never execute.

packages/ui/src/features/sidebar/components/SidebarMenu.tsx
```diff
@@ -35,12 +34,18 @@ import { useQueryClient } from "@tanstack/react-query";
 import { memo, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { ArchiveRunningTaskDialog } from "./ArchiveRunningTaskDialog";
 import { SidebarItem } from "./SidebarItem";
-import { SidebarNavSection } from "./SidebarNavSection";
 import { TaskListView } from "./TaskListView";
 import { TasksHeader } from "./TasksHeader";
 
 const log = logger.scope("sidebar-menu");
 
+function isEditableTarget(target: EventTarget | null): boolean {
```
k11kirky, line 42: `isEditableTarget` is restored verbatim from the deleted `MainSidebar.tsx` as a fresh local copy — at least an 8th independent copy of the same editable-target check in this codebase (`AgentBuilderDockLayout.tsx`, `HomeView.tsx`, `useInboxReportListSelection.ts`, `useAutoFocusOnTyping.ts`, `PlanApprovalSelector.tsx`, `useActionSelectorState.ts`, `SpaceSwitcher.tsx`), with no shared helper in `@posthog/shared` or `@posthog/ui`. Pre-existing duplication, not introduced here, but this was a natural point to extract one.

k11kirky (commented):
Re-reviewed against the latest 2 commits (`dd65e61b`, `15e6e46f`). The new `landOnDefault()` fix (default new tab / last-tab-close to `#me`/new-task instead of the `/website` index) is a real, well-targeted fix for a genuine bug, and I traced the resulting `decideTabNavigation` flow end-to-end — it correctly fills the blank tab. Two notes on it plus my earlier findings are inline below. Nothing else material changed; the `ChannelsSidebar.tsx`/`SidebarNavSection.tsx` diff is just the merge-conflict resolution re-homing the pre-existing "Activity" nav item under the new unified nav (correctly gated on `channelsEnabled`).

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
adamleithp: Fixed in d0ae33387 — `browserTabsStore.setSnapshot` now bails when the incoming snapshot is value-equal to the current one (reference check, then a structural compare), so a mutation's authoritative echo no longer re-renders subscribers when an optimistic write already set the same value. Caveat you correctly flagged: `setTabTarget` stamps a fresh `lastActiveAt`, so that echo still differs and writes once (as intended) — the guard eliminates the redundant *identical* echoes (activate / close / openOrFocus / setOrder / setActiveTab).

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
adamleithp: Fixed in d0ae33387 — extracted `applyOptimistic(transform)` (stable `useCallback`) used by both the activate and replace cases, so the read-current-snapshot → transform → `setSnapshot` rule lives in one place instead of being hand-inlined twice.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
adamleithp: Fixed in d0ae33387 — with the `setSnapshot` value-equality guard (see the thread on ~L204), an echo equal to the current snapshot no longer changes the store, so this effect doesn't re-run for it. Genuinely-changed writes still run once.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
adamleithp: Fixed in d0ae33387 — `#me` provisioning is now deduped through a shared module-level in-flight promise, so a double-fired/held Cmd+T can't create two `me` folders. On tests: no `BrowserTabStrip` test file exists yet, so standing up that harness is a follow-up rather than part of this PR.

packages/ui/src/features/sidebar/components/ProjectSwitcher.tsx
adamleithp: Fixed in d0ae33387 — removed the `triggerVariant` prop and the dead `button` branch (the only caller is `<ProjectSwitcher />`), plus the now-unused `Button` / `ChevronDownIcon` imports.

packages/ui/src/router/routes/__root.tsx
adamleithp: Fixed in d0ae33387 — removed the `"leaving"` mode from `FeedbackModal` (type union, copy map, and `FEEDBACK_SOURCE_BY_MODE`) and its test case; the skip-behaviour test now runs against `posthog-web`.

packages/ui/src/features/sidebar/components/SidebarNavSection.tsx
adamleithp: Good catch. Leaving this as a follow-up — whether to keep re-firing the legacy `enter_space`/`leave_space` events at all, and whether to preserve the old `title_bar`/`header_button` surface values, is an analytics data-shape decision I don't want to make silently in this PR. Will raise it separately.

packages/ui/src/features/sidebar/components/SidebarMenu.tsx
adamleithp: Agreed it's the natural extraction point, but since it's pre-existing duplication spanning ~8 files and out of scope for this PR, I'll do the shared-helper extraction as a separate follow-up.

k11kirky (commented):
Re-reviewed after `d0ae3338` ("address PR review") and `b693feb9` ("completely rebuilt tabs"). This isn't a patch pass — it's a real architectural fix, and it resolves nearly everything I flagged:

**Fixed, verified by reading the new code + tests:**
- **Double mirror-write / no equality check** → `browserTabsStore.setSnapshot` now guards on `snapshotsEqual` (reference check, then structural `JSON.stringify` compare — correctly reasoned as "can produce a false mismatch, never a false match," so it's safe to skip only genuine no-ops).
- **Duplicated optimistic-apply pattern** → extracted into a proper `tabsSync.ts` module (`applyLocalTransform` + `persistWrite` + `applyRemoteSnapshot`, with an `inFlight` counter and last-settle-wins reconciliation). This is a legitimate synchronization primitive, not just a helper function — I traced the FIFO-IPC assumption it depends on and it holds for this single-window Electron transport. It's also now the *only* path for writes (`BrowserTabsDnd.tsx` and the events contribution were migrated too), and it has 7 dedicated tests (`tabsSync.test.ts`) covering in-flight dedup, last-settle-wins, failed-write re-pull, and re-pull superseded by a newer write. Good coverage of the exact race conditions that mattered.
- **Self-retriggering nav-sync effect** → the effect no longer depends on the reactive `snapshot`; it reads `readMirror()` fresh inside and only re-runs on genuine location changes. The doc comment on this is explicit about why (`snapshot` in the deps would make a handler's own optimistic write re-trigger the effect against a location that hasn't changed). This is the correct fix, not a workaround.
- **`#me` channel creation race** (from the `landOnDefault` commit) → `personalChannelInFlight` is now a shared in-flight promise; a double-fired Cmd+T can't create two `#me` folders. Traced the sync-up-to-first-`await` semantics — this is race-free for the realistic double-fire scenario.
- **Dead `ProjectSwitcher` "button" variant** → deleted (prop, branch, unused imports all cleaned up).
- **Dead `FeedbackModalMode.leaving`** → deleted, including the survey-source mapping constant and its test case.
- Bonus: `decideTabNavigation` also gained a `windowTabs` dedup pass (activate an existing tab that already matches the route instead of replacing/duplicating) — a real gap the old logic had, now covered by 6 new test cases.

**Still open, unchanged, lower severity:**
- `SidebarNavSection.tsx`'s toggle still only fires `surface: "nav"` for the legacy `enter_space`/`leave_space` events — `title_bar`/`header_button` surface values are still gone from that breakdown. Not addressed; likely a deliberate call that this is acceptable, but flagging again in case it wasn't.
- `isEditableTarget` in `SidebarMenu.tsx` is still its own copy (8th in the repo) — pre-existing duplication, not a regression from this PR, not worth blocking on.
- No direct test for `BrowserTabStrip.tsx`'s `landOnDefault`/hotkey wiring itself, though the risky concurrency logic it depends on (`tabsSync.ts`) is now well covered, which was the actual risk surface.

No new bugs found in the rebuild. Nice fix.

k11kirky (approved):

PR #3228, opened 2026-07-07
fix(channels): keep the backend task channel in sync on rename and delete

adamleithp:
## Problem

Channels exist in two places: the desktop file-system folder (the sidebar entry) and a backend task channel that owns the task feed and thread messages — and the UI maps one onto the other **by name**. That made two channel actions lossy:

- **Rename** only moved the folder. The name no longer matched any backend channel, so the feed silently auto-provisioned a brand-new empty channel, orphaning every existing task and message under the old name.
- **Delete** only removed the folder, leaving the backend channel live but unreachable with all its tasks stranded on it.

## Changes

- Rename now moves the data-bearing backend channel **by id** in lockstep with the folder: backend first (if it fails, nothing changes), and the backend name is reverted if the folder rename then fails. Server-side name errors (invalid/taken) surface in the error toast.
- Delete now soft-deletes the backend channel's tasks and the channel itself before removing the folder — both are soft deletes server-side, so nothing is lost and everything is recoverable. Individual task deletes are best-effort so one flaky request can't block the delete; a backend failure aborts before the folder is touched.
- Added `renameTaskChannel` / `deleteTaskChannel` to the api-client (the backend endpoints already existed; the client never called them).
- Updated the delete confirmation copy, which previously claimed tasks were not deleted.

## How did you test this?

- 8 new unit tests in `useChannels.test.tsx` covering the rename lockstep/rollback paths and the delete ordering/failure paths; full `@posthog/ui` suite passes.
- `turbo typecheck` for `@posthog/api-client` + `@posthog/ui`, Biome clean.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/canvas/hooks/useChannels.test.tsx
+++ b/packages/ui/src/features/canvas/hooks/useChannels.test.tsx
@@ -1,4 +1,5 @@
 import type { Schemas } from "@posthog/api-client";
+import type { TaskChannel } from "@posthog/shared/domain-types";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { act, renderHook, waitFor } from "@testing-library/react";
 import type { ReactNode } from "react";
@@ -7,6 +8,13 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 const mockClient = vi.hoisted(() => ({
   getDesktopFileSystemChannels: vi.fn(),
   createDesktopFileSystemChannel: vi.fn(),
+  renameDesktopFileSystemChannel: vi.fn(),
+  deleteDesktopFileSystem: vi.fn(),
+  getTaskChannels: vi.fn(),
+  renameTaskChannel: vi.fn(),
+  deleteTaskChannel: vi.fn(),
+  getTasks: vi.fn(),
+  deleteTask: vi.fn(),
 }));
 vi.mock("@posthog/ui/features/auth/authClient", () => ({
   useOptionalAuthenticatedClient: () => mockClient,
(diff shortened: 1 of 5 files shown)


packages/ui/src/features/canvas/hooks/useChannels.test.tsx
```diff
@@ -99,3 +107,205 @@ describe("useChannelMutations", () => {
...
   });
 });
+
+describe("useChannelMutations rename", () => {
+  function taskChannel(
+    id: string,
+    name: string,
+    channel_type: TaskChannel["channel_type"] = "public",
+  ): TaskChannel {
+    return { id, name, channel_type, created_at: "2026-01-01T00:00:00Z" };
```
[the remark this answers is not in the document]
adamleithp: Lifted the `taskChannel` factory to module scope alongside `folder`; removed both per-describe copies. Tests still green.

packages/ui/src/features/canvas/hooks/useChannels.ts
```diff
@@ -99,27 +105,93 @@ export function useChannelMutations() {
...
+          (c) => c.channel_type === "public" && c.name === normalized,
+        );
+        if (backendChannel) {
+          const tasks = await client.getTasks({ channel: backendChannel.id });
+          // Best-effort per task — one failed soft delete shouldn't strand
+          // the rest or block the channel delete (the failed task stays
+          // attached to the soft-deleted channel, still recoverable).
+          await Promise.allSettled(
+            tasks.map((task) => client.deleteTask(task.id)),
+          );
```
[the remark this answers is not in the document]
adamleithp: Documented the cap at the call site: `getTasks` caps at 500 and doesn't paginate, so a >500-task channel soft-deletes only the first page here. The channel itself is still soft-deleted below, so stragglers stay attached to it and remain recoverable — nothing lost, just left behind. Left pagination out of scope for this fix.

packages/ui/src/features/canvas/hooks/useChannels.ts
```diff
@@ -99,27 +105,97 @@ export function useChannelMutations() {
...
+        }
+      }
+      try {
+        return await client.renameDesktopFileSystemChannel(id, name);
+      } catch (error) {
+        // The folder rename failed after the backend channel moved — put the
+        // backend name back so the two sides stay in sync.
+        if (backendChannel) {
+          await client
+            .renameTaskChannel(backendChannel.id, from)
```
k11kirky, line 180: Once the folder rename fails and this compensating revert is attempted, its own failure is discarded here (`.catch(() => undefined)`). If the revert also fails (e.g. a concurrent rename already took `from`, or a transient error), the backend channel is left permanently stuck on `to` while the folder still shows `from` — with no log or user-visible signal beyond the generic "Couldn't rename channel" toast. Worth at least logging via `ROOT_LOGGER` so the desync is diagnosable.

packages/ui/src/features/canvas/hooks/useChannels.ts
```diff
@@ -99,27 +105,97 @@ export function useChannelMutations() {
...
+          // Best-effort per task — one failed soft delete shouldn't strand
+          // the rest or block the channel delete (the failed task stays
+          // attached to the soft-deleted channel, still recoverable).
+          await Promise.allSettled(
+            tasks.map((task) => client.deleteTask(task.id)),
+          );
+          await client.deleteTaskChannel(backendChannel.id);
+        }
+      }
       return client.deleteDesktopFileSystem(id);
```
k11kirky, line 137: This delete path soft-deletes the backend channel and its tasks unconditionally before an unguarded `deleteDesktopFileSystem(id)` call — unlike `renameMutation` below, there's no rollback if this final call fails. If it fails after the backend soft-delete already succeeded, the folder stays in place but its backend channel is now soft-deleted and absent from `getTaskChannels()`. Reopening that folder later will make `useBackendChannel` silently provision a brand-new empty channel, permanently orphaning the original tasks with no error surfaced about that specific consequence.

packages/ui/src/features/canvas/hooks/useChannels.ts
```diff
@@ -99,27 +105,97 @@ export function useChannelMutations() {
   });
 
   const deleteMutation = useMutation({
-    mutationFn: async (id: string) => {
+    mutationFn: async ({ id, name }: { id: string; name: string }) => {
       if (!client) throw new Error("Not authenticated");
```
k11kirky, line 109: Both `deleteMutation` and `renameMutation` implement multi-step, cross-backend orchestration inline in a `useMutation` — list backend channels, find a match by normalized name, conditionally mutate, and (for rename) manually compensate on failure. Per this repo's CLAUDE.md, that's business orchestration ("orchestration, retries, dedupe, sagas... Inversify services only" is `packages/core`'s job), and Rule 5 says hooks should wrap exactly one mutation, with multi-source orchestration belonging in a service method. The repo already has a `Saga` abstraction (`packages/shared/src/saga.ts`) used by other `packages/core` sagas, and a sibling `ChannelTasksService` in `packages/core/src/canvas/` — this logic seems like a natural fit there instead, and would be reusable/testable outside a React hook.

packages/ui/src/features/canvas/hooks/useChannels.ts
```diff
@@ -99,27 +105,97 @@ export function useChannelMutations() {
...
+        const backendChannels = await client.getTaskChannels();
+        const backendChannel = backendChannels.find(
+          (c) => c.channel_type === "public" && c.name === normalized,
+        );
+        if (backendChannel) {
+          // getTasks caps at 500 and does not paginate, so a channel with more
+          // than 500 tasks soft-deletes only the first page here. The channel
+          // itself is still soft-deleted below, so any stragglers stay attached
+          // to it and remain recoverable — nothing is lost, just left behind.
+          const tasks = await client.getTasks({ channel: backendChannel.id });
```
k11kirky, line 127: This fires one concurrent HTTP DELETE per task with no batching — for a channel near the 500-task `getTasks` cap, that's up to 500 simultaneous requests. `Promise.allSettled` handles individual failures fine, but there's no cap on concurrency, which could trigger backend rate-limiting during the burst on a large channel. (Not a new pattern in this codebase, but worth a concurrency limit here given the potential scale.)

packages/ui/src/features/canvas/hooks/useChannels.ts
```diff
@@ -99,27 +105,97 @@ export function useChannelMutations() {
...
       if (!client) throw new Error("Not authenticated");
+      // The task feed lives on a separate backend channel. Deleting only the
+      // folder would leave that channel live but unreachable, with its tasks
+      // orphaned under it. Instead, soft-delete the tasks and the backend
+      // channel first (both are soft deletes server-side, so nothing is
+      // lost), then remove the folder. Backend first: if it fails, nothing
+      // local has changed.
+      const normalized = normalizeChannelName(name);
+      if (normalized !== PERSONAL_CHANNEL_NAME) {
+        const backendChannels = await client.getTaskChannels();
```
k11kirky, line 118: This "find the public backend channel by normalized name" lookup is duplicated here and in `renameMutation` below, and a very similar lookup already exists in `useBackendChannel` (useTaskChannels.ts). Worth extracting to one shared helper (e.g. `findPublicBackendChannel(channels, normalizedName)`) so the matching rule only needs to change in one place.

PR #3229, opened 2026-07-07
refactor: extract shared Adapter type (claude|codex)

jonathanlab:
Introduces `packages/shared/src/adapter.ts` as the single source of truth for the "claude" | "codex" runtime-adapter union, previously duplicated (often under different local names: AgentAdapter, AdapterType, RuntimeAdapter, PreviewAdapter, CloudRuntimeAdapter) across core, ui, workspace-server, agent, api-client, and the desktop/mobile apps.

All call sites now import Adapter from @posthog/shared instead of redeclaring the literal union.


tatoalo (approved):
thanks!

PR #3232, opened 2026-07-07
feat(local-mcp): forward importable local MCP servers into cloud runs

richardsolomou:
Part 1b of bringing local MCP servers to cloud task runs. **Stacked on #3231** (config reading + classification) — review that first; this PR's own diff is the creation-payload wiring, composer UX, and the Django-side spec. Behind the `posthog-code-local-mcp-import` feature flag.

## What

- **Creation payload**: cloud run creation (`buildCloudRunRequestBody`) gains an optional `imported_mcp_servers` field carrying the user's importable local servers in exactly the shape the sandbox agent server's `remoteMcpServerSchema` accepts, so the backend can merge them into `--mcpServers` without transformation. Threaded `TaskInput` → `useTaskCreation` → `prepareTaskInput` → `TaskCreationSaga` → `createTaskRun`, cloud-only. Until the Django half lands the backend ignores the field, so this degrades gracefully (and the flag defaults off anyway).
- **UX**: when the composer is in cloud mode, a `LocalMcpServersButton` lists the user's local MCP servers annotated "Available in cloud" / "Requires your machine" / "Not available in cloud".
- **`docs/cloud-mcp-import.md`**: the design doc, including the precise Django-side spec (validation incl. server-side SSRF re-checks, encrypted write-only header storage, `--mcpServers` merge rules, codex-acp reachability caveat) and the header-rotation design built on the existing `refresh_session` command.

## Why rotation ships as a design, not code

`~/.claude.json` static headers carry no expiry metadata to watch, and OAuth-backed servers managed by Claude Code keep tokens in Claude's credential store, not in `mcpServers.headers` — so those aren't importable this way at all (the relay design covers them). Static headers ship first; the doc specifies the `refresh_session` client path (today the desktop never sends that command for cloud runs — `sendCommandInput` in `packages/core/src/cloud-task/schemas.ts` stops at `set_config_option`, a small discrepancy from the task brief) plus the Django allowlist change.

## Notes for reviewers

- Cloud runs currently import **user-scoped** servers only: cloud task creation picks a GitHub repository, not a local checkout, so there's no `cwd` to key `projects[cwd].mcpServers` off. Documented follow-up.
- Codex config (`~/.codex/config.toml`) import and `${VAR}` header expansion are documented follow-ups.

## Validation

`pnpm typecheck` / `pnpm lint` green; all suites touched by this PR pass (remaining full-suite failures are pre-existing sandbox-environment failures, identical on a clean checkout).

### DIFF
--- a/docs/cloud-mcp-import.md
+++ b/docs/cloud-mcp-import.md
@@ -1,0 +1,171 @@
+# Importing local MCP servers into cloud task runs
+
+Status: client side implemented behind the `posthog-code-local-mcp-import`
+feature flag. The Django side (spec below) is not implemented; until it lands,
+the backend ignores the extra creation-payload field and cloud runs behave as
+before.
+
+## Problem
+
+A local task run gets all of the user's MCP servers: the PostHog MCP plus
+`MCPServerInstallation` records (built by `AgentAuthAdapter.buildMcpServers` in
+`packages/workspace-server/src/services/agent/auth-adapter.ts`), and — for the
+Claude adapter — the user's own servers from `~/.claude.json`
+(`loadUserClaudeJsonMcpServers` in
+`packages/agent/src/adapters/claude/session/mcp-config.ts`).
+
+A cloud run's sandbox only gets what the backend bakes into the agent server's
+`--mcpServers` flag at spawn (`remoteMcpServerSchema` in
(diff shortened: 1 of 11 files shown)


richardsolomou:
Consolidated into #3231 per review feedback — all changes now live there.

PR #3233, opened 2026-07-07
docs: design for relaying local MCP servers into cloud task runs

richardsolomou:
Part 2 of bringing local MCP servers to cloud task runs: the design doc for relaying **stdio servers and private-network URLs** back to the desktop app. **Design only — implementation is gated on review of this document** and will ship behind a `posthog-code-mcp-relay` feature flag. Complements the import work in #3231 / #3232.

## What's in the doc (`docs/cloud-mcp-relay.md`)

- **Protocol**: mirrors the existing question-relay pattern (`permission_request` event / `permission_response` command in `agent-server.ts`) with a new pair — an `mcp_request` event (uuid `requestId`, server name, JSON-RPC payload, `expiresAt`) over the durable event stream, answered by an `mcp_response` command through the run command endpoint. Zod schemas included. (Note: the brief pointed at commit `ff4a7334f` as prior art; that commit isn't in this repo's history — single-commit clone — but the mechanism it introduced is fully present in `agent-server.ts` and is what the design mirrors.)
- **Sandbox side**: loopback HTTP MCP endpoints per relayed server so the Claude/Codex adapters need no changes; correlation map with 60s timeouts; 503 when no desktop has been seen recently so Codex-style reachability pruning treats a disconnected relay as unreachable and Claude gets a clean MCP error instead of a hang.
- **Desktop side**: a workspace-server executor that owns the real connections (stdio spawn, private-URL dial) resolved from *local* config by name — commands, args, env, and headers never cross the wire.
- **Security posture**: relayed tools default to always-ask via the existing tool-approval + permission-relay mechanism, regardless of the run's permission mode; designation is explicit per run (names only); headless runs (web/mobile/Slack) simply have no relay.
- **Durability decision**: `mcp_request` events stay persisted with the run (audit value, same semantics as the question relay) with 64 KB request / 256 KB response caps; responses travel the non-persisted command path and the spec requires that to remain true.
- **Django-side spec**: `relayed_mcp_servers` on run creation, `mcp_response` command allowlisting, event-ingest note — for the `posthog/posthog` repo, not implemented here.
- Failure-mode table, testing plan, staged rollout.

### DIFF
--- a/docs/cloud-mcp-relay.md
+++ b/docs/cloud-mcp-relay.md
@@ -1,0 +1,245 @@
+# Design: relaying local MCP servers into cloud task runs
+
+Status: **design — not implemented**. Implementation is gated on review of
+this document, ships behind a `posthog-code-mcp-relay` feature flag, and
+follows [cloud-mcp-import.md](./cloud-mcp-import.md) (which handles the easy
+case: url-based servers on public hosts).
+
+## Problem
+
+Two classes of local MCP servers cannot be imported into a cloud sandbox by
+forwarding a URL:
+
+- **stdio servers** — a process spawned on the user's machine
+  (`npx @playwright/mcp`, a company-internal CLI, ...).
+- **private-URL servers** — `http`/`sse` servers only reachable from the
+  user's machine or network (localhost, RFC1918, tailnet, VPN).
+
+Both classes are classified `requires_desktop` by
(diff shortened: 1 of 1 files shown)


richardsolomou:
Consolidated into #3231 per review feedback — the relay design doc now lives there.

PR #3235, opened 2026-07-07
feat: add conversations

veryayskiy:
Add conversations to gateway

### DIFF
--- a/packages/agent/src/utils/gateway.ts
+++ b/packages/agent/src/utils/gateway.ts
@@ -3,7 +3,8 @@ export type GatewayProduct =
   | "background_agents"
   | "signals"
   | "slack_app"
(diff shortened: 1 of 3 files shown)


packages/agent/src/utils/gateway.ts
```diff
@@ -21,6 +22,9 @@ export function resolveGatewayProduct({
   if (originProduct === "signal_report" || originProduct === "signals_scout") {
     return "signals";
   }
+  if (originProduct === "support_reply") {
+    return "conversations";
```
skoob13, line 26: Either support_reply or conversations – whatever works better for you, but I'd not change the value here.

skoob13 (approved):
Exactly this

PR #3237, opened 2026-07-07
feat(agent-builder): author, save, and dry-run custom tools

dmarticus:
## What

Wires the backend's custom-tool authoring loop ([posthog#66584](https://github.com/PostHog/posthog/pull/66584)) into PostHog Code. Until now the agent builder could only *read* tool source (`tools/<id>/source.ts`, read-only) — this adds the client + GUI to **edit, save, and dry-run** an existing custom tool end to end, behind the existing `agent-platform` flag.

Scope: this edits tools already declared in the spec — there's no create-a-new-tool or delete-from-UI flow yet, and it edits `source.ts` only (`description` / `args_schema` are round-tripped unchanged).

## How

**Part A — API client + types** (hand-written; agent-platform types are *not* generated from OpenAPI)
- `@posthog/shared/agent-platform-types`: `ToolCompileError`/`ToolCompileErrorKind`, `ToolCapabilities`, `WriteToolRequest`/`WriteToolResult`, `DryRunToolRequest`/`DryRunToolEnvelope`/`DryRunToolResult`.
- `posthog-client.ts`:
  - `putRevisionTool` — 422 returns a typed compile-failed result (rendered inline); other non-2xx (400/409 sealed/…) still throw.
  - `deleteRevisionTool` — 404 (`tool_not_found`) treated as success.
  - `dryRunRevisionTool` — reads `ok` from the **body** (a throwing tool is HTTP 200 with `ok:false`); 200 and 500 both return the envelope; 429 → `throttled` (with optional `max_concurrent`), 503 → `unavailable`. Never thrown, never retried.
  - `parseFailedRequest` helper recovers status + body from the shared fetcher's `Failed request: [status] body` throw.

**Part B — hooks**
- `useSaveRevisionTool` (invalidates the `bundle(...)` query only when the compile persisted), `useDeleteRevisionTool` (invalidates bundle), `useDryRunRevisionTool` (no invalidation). All expose `isPending`.

**Part C — UI** (agent builder)
- `ToolSourcePanel` — **read-first**: the source is a collapsed, read-only block at rest; **Edit source** (draft-only) and **Test tool** are explicit toggles, so a tool node stays as compact as every other config node.
- `ToolSourceEditor` (behind the edit toggle) — compile-on-save with inline `ToolCompileError` diagnostics (kind + message + 1-based line/column), Revert, double-submit guard, and a `secret_refs` / `dynamic_secret_refs` readout from the save response.
- `ToolDryRunPanel` (behind the test toggle) — args JSON editor, optional `mock_secrets` key/value rows, and a result area handling success (result + `duration_ms`), tool failure (`error.code`/`message`), **429 busy**, and **503 unavailable**.
- Everything is gated behind `AGENT_PLATFORM_FLAG`; non-draft / non-custom tools stay read-only.

## Testing
- `pnpm typecheck` (23/23 packages) and `pnpm biome check` on all touched files pass.
- New client tests: 200/422/409 writes, delete 404-as-success, dry-run 200-ok / 200-fail / 500 / 429 (with and without `max_concurrent`) / 503 + `mock_secrets` passthrough. New hook test for the invalidate-only-when-persisted branch. The `@posthog/api-client` and `@posthog/ui` suites pass.

## Note
End-to-end dry-run requires the deployed backend (janitor + main app) to include [posthog#66584](https://github.com/PostHog/posthog/pull/66584) with the sandbox backend configured. Until then dry-run returns **503**, which is the expected "not available here" state this UI is built against.


daniloc (approved):
thanks for your patience. Yeah, I know what you mean. it's fine, it works, and also I can't imagine wanting to do the edit/test loop myself at ALL, that's a robot job

so approving with the same faint enthusiasm as the PR itself

PR #2723, opened 2026-06-17
feat(agent): Default Codex to the native app-server harness

charlesvien:
## Problem

Codex ran through the Zed codex-acp ACP adapter, an extra translation layer that weakens native Codex capabilities (thread resume/fork, typed item and tool events, detailed usage, structured outputs, approvals).

## Changes

1. Bundle the native `codex` CLI binary (download-binaries + vite copy)
2. Add `CodexAppServerAgent` speaking Codex's native app-server JSON-RPC under an unchanged ACP surface
3. Add the app-server JSON-RPC client, ACP event mapping, and spawn helper
4. Default `createAcpConnection` to the native agent, with `POSTHOG_CODEX_USE_ACP=1` to fall back to codex-acp

## How did you test this?

Manually

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #2975 
- <kbd>&nbsp;1&nbsp;</kbd> #2723 👈

PR #3239, opened 2026-07-07
fix(agent): stop phantom desktop app boots from agent subprocesses

gantoine:
## Problem

Users sometimes see dozens of empty dark windows cascade onto the screen with dozens of hung "node" processes in Activity Monitor (hit again today — screenshots in the internal thread). Root cause was established in #3108: the agent host puts a `node -> app binary` shim on PATH that only behaves as node while `ELECTRON_RUN_AS_NODE=1` survives the caller's environment. Claude's Bash tool replays commands under a login-shell snapshot of the user's dotfiles, so direnv/mise/flox or a stray `unset` strips the var, and every `node` invocation then boots a full desktop app instance: while the real app runs, phantoms lose the single-instance lock and exit invisibly; when the lock is free (quit, crash, update restart) each spawn opens an empty window and never exits.

#3108 and its follow-up #3113 were closed unmerged, so main is still exposed (#3114 remains open). This PR re-lands the #3108 scope, rebased onto current main.

## Changes

Cherry-picked from `feat/vendored-node-runtime` (original work by @charlesvien):

- The `node` shim is now a self-contained wrapper script that exports `ELECTRON_RUN_AS_NODE=1` itself before exec'ing the app binary (POSIX; Windows keeps the symlink). Also heals stale shims pointing at a moved binary, which the old EEXIST-skip never did. (`node-shim.ts`)
- Defense in depth: the workspace-server child env carries `POSTHOG_CODE_INTERNAL_CHILD=1`, and `bootstrap.ts` refuses to boot the packaged app when the marker is present without `ELECTRON_RUN_AS_NODE`, failing fast on stderr instead of racing the single-instance lock.
- Integrated terminals stop inheriting `ELECTRON_RUN_AS_NODE`, so Electron-based CLIs launched from app terminals don't silently run as node. (`shell.ts`)

Adapted to current main (the old `adapters/codex` was replaced by `codex-app-server` in #2723):

- The codex local-tools MCP helper runs `command: process.execPath` — the app binary in packaged installs — while codex's own env has the var deleted. Its server config now carries `ELECTRON_RUN_AS_NODE=1` explicitly, porting the old-adapter fix. Inert on real node (cloud runs).

Not included: the #3113 vendored Node runtime (tracked in #3114).

## How did you test this?

- Targeted suites all green: `node-shim.test.ts` (7), `agent.test.ts` (25), `shell.test.ts` (5) in workspace-server; `local-tools-mcp.test.ts` (4, incl. new env assertion), `spawn.test.ts` (7), `codex-app-server-agent.test.ts` (57) in agent.
- Full `pnpm typecheck` and biome on all changed files pass.
- Unrelated local failures (db repository tests, claude settings, posthog-analytics) are byte-identical files to main and fail due to the local better-sqlite3 Electron/Node ABI flip-flop — fittingly, the dev machine's agent shell itself runs vitest under the app binary via the shim this PR hardens. CI runs them clean.

Closes #3108-scope; refs #3114.

---

### DIFF
--- a/apps/code/src/main/bootstrap.ts
+++ b/apps/code/src/main/bootstrap.ts
@@ -22,6 +22,19 @@ import path from "node:path";
 import { app, crashReporter, protocol } from "electron";
 import { fixPath } from "./utils/fixPath";
 
+// The workspace-server tree runs with a `node -> app binary` PATH shim that
+// relies on ELECTRON_RUN_AS_NODE. If a descendant strips that var and runs
+// `node` (or process.execPath), this binary boots as a full app: it races the
+// single-instance lock and, whenever the lock is free, opens phantom windows.
+// POSTHOG_CODE_INTERNAL_CHILD marks that tree so we can refuse to boot here.
+if (app.isPackaged && process.env.POSTHOG_CODE_INTERNAL_CHILD) {
+  process.stderr.write(
+    "[posthog-code] Refusing to start the desktop app from inside its own " +
+      "child process tree (expected ELECTRON_RUN_AS_NODE=1).\n",
+  );
+  process.exit(1);
+}
+
 const isDev = !app.isPackaged;
 
 // Set app name for single-instance lock, crashReporter, etc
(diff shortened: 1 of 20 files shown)


charlesvien:
note to all: this needs careful review

charlesvien (approved):

PR #3242, opened 2026-07-07
feat(sessions): reorder and edit queued messages in place

gantoine:
## What

Two additions to the queued-messages dock (the follow-ups pinned above the composer while a turn is running):

1. **Drag to reorder** — each queued card is now draggable. The queue drains in array order, so the order shown is the order the messages send. Uses the existing `@dnd-kit/react` sortable pattern (same as sidebar folders / browser tabs), with a `PointerSensor` movement threshold so the card's buttons still take clicks.
2. **Edit in place** — the pencil now loads a queued message back into the composer for editing **without removing it from the queue**. Submitting the edit updates the message where it sits (same id, position, and queue timestamp) instead of appending a new one. If the message already drained or was discarded while being edited, it falls back to sending as a fresh message so nothing is lost. The card being edited is highlighted and gains a cancel (✕) affordance.

Previously the pencil ("return to editor") pulled the message out of the queue entirely; that hook (`useReturnQueuedMessageToEditor`) is replaced by `useEditQueuedMessage` / `useCancelQueuedMessageEdit`.

Screenshot 2026-07-08 at 4 45 37 PM

## How

- **core** — `moveQueuedMessage` and `updateQueuedMessage` store setters; `SessionService.updateQueuedMessage` mirrors the enqueue normalization (cloud recomputes transport display + raw payload via `getCloudPromptTransport`; local stores the serialized text).
- **ui** — `sessionViewStore` tracks the active edit target per task (`editingQueuedIdByTaskId`); `useSessionCallbacks.handleSendPrompt` routes to `updateQueuedMessage` when an edit is in progress; `QueuedMessagesDock` wires the `DragDropProvider` + sortable cards and clears a stale edit target if the message leaves the queue.

## Testing

- `pnpm --filter @posthog/core typecheck` + `@posthog/ui typecheck` — clean
- Biome lint — clean
- New `packages/core/src/sessions/sessionQueueEditing.test.ts` (9 tests) for the store setters; updated `QueuedMessagesDock.test.tsx`
- Full suites: core sessions 267 pass, UI sessions 348 pass

Not yet dogfooded in the running app — happy to do that next.

---


charlesvien:
hey @gantoine, what's the status of this one?

gantoine:
i'm still wresting with the UX a bit, namely around multiple stacked prompts executing as sa group and cutting off the edit (racing to save before the prompt is sent)

charlesvien (approved):

PR #3271, opened 2026-07-08
fix(sessions): highlight every skill chip in submitted prompts

adboio:
## Problem

Reported in Slack: once a prompt is submitted, only the *first* skill in the prompt keeps its chip highlight — later skills render as plain text. In the input box, all of them are highlighted. This is a purely cosmetic inconsistency in how a submitted user message is rendered in the thread.

## Changes

Submitted prompts flatten their `<skill … />` tags to plain `/name` text before the thread renders them, but the rendered-message chip parser (`parseFileMentions.tsx`) only matched a slash command anchored to the very start of the string (a non-global regex used with a single `.match()`). Any second-or-later `/skill` fell through to plain markdown.

The parser now scans for every slash command with a global, non-anchored regex and merges those matches with the existing mention-tag matches (files, folders, GitHub refs, error context), sorted by position, so each chip renders regardless of order.

## How did you test this?

Added `parseFileMentions.test.tsx` covering multiple slash commands, non-leading commands, mid-word slashes (not matched), and slash commands interleaved with file mentions. Ran:
- `vitest run parseFileMentions` — 6 passed
- `vitest run UserMessage` — 5 passed (existing suite still green)
- `biome lint` on the changed files — clean

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1783513480844189?thread_ts=1783513480.844189&cid=C09G8Q32R6F)*

### DIFF
--- a/packages/ui/src/features/sessions/components/session-update/parseFileMentions.test.tsx
+++ b/packages/ui/src/features/sessions/components/session-update/parseFileMentions.test.tsx
@@ -1,10 +1,7 @@
 import { Theme } from "@radix-ui/themes";
 import { render, screen } from "@testing-library/react";
 import { describe, expect, it } from "vitest";
-import {
-  hasFileMentions,
-  parseFileMentions,
-} from "./parseFileMentions";
+import { hasFileMentions, parseFileMentions } from "./parseFileMentions";
 
 function renderParts(content: string) {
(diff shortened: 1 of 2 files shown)


charlesvien:

PR #3284, opened 2026-07-08
chore(agents): rename the "Applications" tab to "Fleet"

dmarticus:
## Summary

The second Agents tab was labelled **Applications** — vague, and it clashed with **Agent Builder**, which is both mentioned in the page header *and* one of the listed agents. Renames the visible label to **Fleet**, which the app already uses elsewhere for the collection of running agents ("In-flight agent sessions across the fleet") and which parallels the sibling **Scouts** tab tonally.

## What actually changes

Three user-visible strings:
- Tab label in `AgentsTabLayout` (`Applications` → `Fleet`)
- Back-nav link on the agent detail page (`AgentDetailLayout`)
- Back-nav link on the fleet-approvals page (`AgentFleetApprovalsPane`)

Plus six stale JSDoc / inline comments that referred to "the Applications tab" — updated so future readers aren't confused when the visible label no longer matches the code comments.

## What doesn't change

None of the internal names:
- Route: `/code/agents/applications`
- Feature directory: `packages/ui/src/features/agent-applications/`
- Enum value: `AgentsTab = "scouts" | "applications"`
- Feature flag: `AGENT_PLATFORM_FLAG` / `"agent-platform"`
- Hook names: `useAgentApplications`, `agentApplicationsKeys`, etc.
- Query keys, DB schema, tRPC procedures

Renaming those would be scope creep and land in follow-ups if we want to align internal vocabulary. This PR only touches what a user sees on screen.

## Test plan

- [ ] Load the Agents section, confirm the tab reads "Fleet"
- [ ] Click into an agent, confirm the back-nav link reads "← Fleet"
- [ ] Open Fleet approvals, confirm the back-nav link reads "← Fleet"
- [ ] Deep-linking to `/code/agents/applications` still resolves (route unchanged)

### DIFF
--- a/packages/ui/src/features/agent-applications/components/AgentAnalyticsView.tsx
+++ b/packages/ui/src/features/agent-applications/components/AgentAnalyticsView.tsx
@@ -43,7 +43,7 @@ const tileTitle = (label: string): ReactNode => (
  * HogQL and shapes the data; this renders one of loading / error / empty /
  * populated.
  *
- * `scope` "overview" is the fleet board blended into the Applications landing
+ * `scope` "overview" is the fleet board blended into the Fleet landing
  * (KPIs + spend-by-agent + cost-by-model; the agent list below carries the
  * per-agent breakdown). "agent" is the per-agent Observability tab (KPIs +
  * cost-by-model + tool reliability — spend-by-agent is meaningless for one).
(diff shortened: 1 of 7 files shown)


daniloc (approved):
tasteful

PR #3308, opened 2026-07-09
fix(chat): float agent-response copy button instead of new line

MattPua:
## Problem

The copy button on agent responses in the new chat thread renders on its own line below the message, and reserves that empty line even when hidden.

Root cause: it was placed in a `ChatMessageFooter`, which is a flex-column child of `ChatMessageContent`, so it always stacks below the bubble.

## Changes

- `AgentProse` (`ChatThread`): float the copy button at the message top-right on hover (absolute), matching the classic `AgentMessage`. No longer consumes a row.

## How did you test this?

Biome lint + `@posthog/ui` typecheck pass (pre-existing `defaultEligibleModel` errors are unrelated). Not yet verified in the running app.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/sessions/components/chat-thread/ChatThread.tsx
+++ b/packages/ui/src/features/sessions/components/chat-thread/ChatThread.tsx
@@ -563,23 +563,21 @@ const AgentProse = memo(function AgentProse({
             )}
           </ChatBubbleContent>
         </ChatBubble>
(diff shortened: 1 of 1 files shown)


MattPua:
Closing — the reported discrepancy was in the classic thread view, not the new ChatThread these changes touched. Reverting.

PR #3324, opened 2026-07-09
Channels: persistent reply row to stop feed scroll jitter

adboio:
## Problem

In channels, scrolling up through the feed caused visible layout jitter: a message with a human thread would come into view, and a beat later its "N replies" teaser would pop in and push the feed down. The teaser was only mounted once a row scrolled near the viewport, and rendered nothing until its thread fetch resolved — so on the first scroll up from the bottom (rows that have never been fetched) the fetch routinely landed *after* the row was already visible.

**Why:** Reported from the Channels (project-bluebird) experience — the late-attaching reply teaser makes scrolling feel unstable, and the existing way to start a thread was hidden in the hover toolbar.

## Changes

- Replace the in-view-gated `RepliesRow` with a persistent `ReplyFooter`, rendered at a constant height for every message: the reply teaser once the thread has messages, and a quiet always-visible **Reply** affordance otherwise. Both states share the same avatar-sized row, so the teaser swaps in after the fetch with **no layout shift**.
- The Reply affordance is now always visible, instead of being tucked into the hover toolbar.
- Add an `enabled` flag to `useTaskThread` so only near-viewport rows fetch/poll; off-screen rows show the static affordance and idle, keeping a long feed from polling every row.

Height parity between the two states relies on both using the same `AvatarGroup size="xs"` row — worth an eyeball in review, but the swap is now a content change, not a `0 → teaser` height jump.

## How did you test this?

- `pnpm --filter @posthog/ui typecheck` — passes.
- Biome `check` on the two changed files — clean.
- No manual/visual run of the app in this environment; the persistent-height claim should be confirmed visually before merge.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/canvas/components/ChannelFeedView.tsx
+++ b/packages/ui/src/features/canvas/components/ChannelFeedView.tsx
@@ -383,9 +383,6 @@ function ReplyFooter({
             </AvatarFallback>
           </Avatar>
         </AvatarGroup>
-        {/* Muted, not the teaser's --primary accent: on an always-present row
-            the accent should stay earned by real replies, not shout on every
-            card. Matches the grey of the populated state's "Last reply" meta. */}
         <ThreadItemRepliesLabel className="text-(--muted-foreground)">
           Reply
         </ThreadItemRepliesLabel>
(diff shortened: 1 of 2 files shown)


adboio:
### Merge activity

* **Jul 10, 5:19 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/code/3324).

PR #3335, opened 2026-07-10
fix(agent): seed hydrated transcripts with estimated context usage

tatoalo:
## Problem

When a cold resume rebuilds the session transcript from S3 logs, every rebuilt message carries `usage: 0`. The resumed SDK session therefore believes its context is nearly empty: it cannot trigger auto-compaction proactively and only learns the real context size from the first API response — or from a "prompt is too long" rejection, at which point compaction itself is already impossible. This was visible in a real failure where the SDK reported a compact boundary of ~16k pre-tokens while the actual request weighed over 1.1M tokens.

Follow-up to #3331.

## Changes

- After building the hydrated JSONL, seed the last assistant message's `usage.input_tokens` with the estimated token count of the rebuilt conversation, so the SDK's context accounting starts near reality and auto-compaction can fire before the first API call when warranted.

## How did you test this?

- New test: only the last assistant entry carries the non-zero estimate; earlier entries stay at zero.
- `pnpm exec vitest run src/adapters/claude/session/jsonl-hydration.test.ts` — 58 passed, 1 pre-existing failure (`sanitizeSessionJsonl` warn-path test) that also fails on `main`.
- `tsc --noEmit` on the agent package.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/agent/src/adapters/claude/session/jsonl-hydration.test.ts
+++ b/packages/agent/src/adapters/claude/session/jsonl-hydration.test.ts
@@ -496,6 +496,29 @@ describe("conversationTurnsToJsonlEntries", () => {
     expect(conv[1].parentUuid).toBe(conv[0].uuid);
   });
 
+  it("seeds only the last assistant message with the estimated context size", () => {
+    const lines = conversationTurnsToJsonlEntries(
+      [
+        { role: "user", content: [{ type: "text", text: "q".repeat(400) }] },
+        {
+          role: "assistant",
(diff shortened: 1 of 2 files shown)


tatoalo:
with the hydration budgets capped at ~half the window (#3331) and the workspace transcript mirror making cold hydration rare (#3333), the SDK learns its real context size from the first API response anyway, so seeding estimated usage adds little. The prompt-too-long fallback (#3334) covers the residual risk.

PR #3338, opened 2026-07-10
feat(browser-tabs): split panes — drag tabs to edges for an IDE layout

adamleithp:
Stacked on #3221.

True IDE-like split panes for the browser-tab system: drag a tab pill onto the left/right/top/bottom edge of the content area (or of any pane) to split the UI, with arbitrary VS Code-style nesting. The nav sidebar and title bar stay stationary; each pane gets its own tab strip and its own back/forward history, and the global title-bar back/forward buttons drive whichever pane is focused.

## How it works

**Data model** (`@posthog/shared` + migration `0020`): windows now hold a recursive pane layout (leaf/split nodes with persisted sizes) plus flat pane rows carrying per-pane `activeTabId`; tabs gain `paneId`; `window.focusedPaneId` names the pane that owns shortcuts and imperative navigation. All tab transforms are pane-scoped, and `ensureSnapshotIntegrity` heals the snapshot on boot (migration backfills one root pane per window — existing tabs are preserved).

**One router per pane**: the window-wide hash router is gone. Each pane hosts its own TanStack router over the shared route tree with in-memory history (`createAppRouter`), so per-pane back/forward falls out by construction. Chrome inverted out of `__root` into `AppShell`, wrapped in a `RouterContextProvider` bound to the focused pane's router — every existing router hook in the sidebar/title bar follows pane focus automatically. Pane locations persist to sessionStorage so Cmd+R/HMR reloads restore; across relaunches each pane's active tab is the source of truth. Settings stays a route in the pane's history but renders as a full-window portal overlay.

**Interactions**:
- drag to a pane edge / window edge → split (translucent preview of the resulting pane)
- drag onto another pane's strip or center → move the tab there
- a pane emptied by a move/split collapses; closing a pane's last tab backfills a blank tab; panes close explicitly via the strip's X
- navigating to something already open in another pane focuses that pane instead of mounting it twice (no double PTY attach)
- resize handles persist sizes (committed on gesture end)
- Cmd+T/W/1-9 act on the focused pane only

**Sync** keeps the local-first model from #3221: every mutation is one synchronous pure-transform apply to the mirror + one background persist, with renderer-minted ids (`paneId`, `blankTabId`) so optimistic and durable state agree.

## Verification

- 638 shared tests (incl. new tree-math, healing, and pane-transform suites) + 1441 UI tests green; typecheck/lint/host-boundary clean.
- Migration exercised against a real SQLite upgrade path (tabs preserved into a root pane).
- Verified live over CDP: edge-drag split (+restart persistence), nested splits, cross-pane strip drop with source collapse, resize persistence, close-pane collapse, focus-follow history buttons, and a full TaskDetail (with terminals) running per pane.


charlesvien:

PR #3345, opened 2026-07-10
fix(tabs): repair stale browser-tabs schema, never-empty strip; move usage to title bar

adamleithp:
## Browser tabs — dead "+" on an empty strip (bluebird follow-up)

**Root cause:** migration `0013` was amended in place on the bluebird branch (same folderMillis). Profiles that dogfooded a pre-merge build have it recorded as applied but kept the old panes-era `browser_windows` schema — no `active_tab_id`. Every browser-tabs query then throws `no such column: active_tab_id`, the whole service dies at boot, the renderer's snapshot seed swallowed the failure, and the strip rendered a permanently dead "+" (the handler bailed on a windowless mirror, and no mutation could bootstrap a window — deadlock).

image

image

**Fix, four layers:**
- **Migration `0020`**: tolerated `ALTER TABLE ADD COLUMN` repairs (`browser_windows.active_tab_id`, plus `browser_tabs.task_id`/`channel_section`/`app_view` for other stale variants). No-op on healthy DBs (the runner tolerates "duplicate column name"); heals divergent ones.
- **Server invariants**: `BrowserTabsService` seeds a blank tab at boot when none survived — the strip should never be empty — and `newBlankTab`/`openOrFocus` heal an unknown `windowId` to the primary window.
- **Seed resilience**: the renderer boot seed retries 3× with backoff and logs terminal failure instead of `.catch(() => undefined)`.
- **Self-healing "+"**: with no window in the mirror, the handler re-pulls the authoritative snapshot (`reseedMirror()`) and creates the tab against the fresh primary window instead of silently returning.

Verified live on an affected profile: schema healed on boot, persisted tabs restored, "+" and Cmd+T work.

⚠️ Anyone who dogfooded early bluebird builds has the stale DB — this should ship before/with the next build. Amending shipped migration files in place is the underlying foot-gun.

## Title bar

- **Usage moved to the title bar**: the sidebar usage bar is now a `Usage: N%` outline button left of "PostHog Web"; hover opens the full plan card (plan · % used, progress bar, reset time, Upgrade) in a quill Popover (`openOnHover` on the trigger). Same gating as before (`posthog-code-billing` + free-plan usage). New `titlebar`/`titlebar_card` analytics surfaces.
- **Native drag region**: the empty title-bar space around the tab strip now drags the window — the strip container no longer carves a `no-drag` rectangle; pills and "+" opt out individually (click/close/reorder unaffected).

## Tests
- `service.test.ts`: boot seeds primary window + blank tab from empty/tab-less stores, leaves populated snapshots untouched, creation mutations heal stale window ids.


.claude/skills/quill-code/scripts/sync-quill.sh
```diff
@@ -0,0 +1,49 @@
...
+HASH="$(md5 -q "$RAW" | cut -c1-8)"
+HASHED="posthog-quill-local-$HASH.tgz"
+cp "$RAW" "$DEST/$HASHED"
+rm -f "$RAW"
+# drop stale local tarballs so the pnpm store can't resolve an old integrity
+find "$DEST" -name 'posthog-quill-local-*.tgz' ! -name "$HASHED" -delete
+
+echo "==> pointing override at .local-quill/$HASHED"
+# single override line: '@posthog/quill': file:./.local-quill/...
+sed -i '' -E "s#('@posthog/quill': file:\./\.local-quill/)[^']*#\1$HASHED#" "$WS"
```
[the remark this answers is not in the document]
adamleithp: Fixed in 9cf2736: `shasum -a 256` replaces `md5 -q` and the `sed -i ''` rewrite now goes through a temp file + `mv`, so the script runs on both macOS and Linux.

packages/ui/src/features/billing/UsageButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <Button
+            variant="outline"
+            size="sm"
+            onClick={() => handleOpenPlan("titlebar")}
+          >
+            {exceeded ? "Usage: limit reached" : `Usage: ${usagePercent}%`}
+          </Button>
+        }
+      />
+      <PopoverContent side="bottom" align="end" sideOffset={6}>
```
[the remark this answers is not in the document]
adamleithp: Fixed in 9cf2736: `PopoverContent` now carries `no-drag` (same treatment as the tab strip's `ContextMenuContent`) so clicks near the card's top edge aren't swallowed by the title-bar drag region.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
```diff
@@ -750,16 +755,36 @@ export function BrowserTabStrip() {
...
+    }
+    // No window means the mirror never seeded (the boot fetch raced or
+    // failed) — the click must not die. Re-pull the authoritative snapshot
+    // (the server always has a primary window) and append into it.
+    void reseedMirror()
+      .then(() => {
+        const win = primaryWindow(readMirror());
+        if (win) createBlankTab(win.id);
+      })
+      .catch(() => undefined);
```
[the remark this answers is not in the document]
adamleithp: Fixed in 9cf2736: `reseedMirror()` now returns the fetched snapshot even when the store apply is skipped (write in flight / newer remote version), and the new-tab handler resolves the primary window from that snapshot instead of re-reading the possibly-still-windowless mirror.

packages/workspace-server/src/db/migrations/0020_repair_browser_tabs_schema.sql
```diff
@@ -0,0 +1,46 @@
...
+	`scroll_state` text,
+	`created_at` integer NOT NULL,
+	`last_active_at` integer NOT NULL,
+	`task_id` text,
+	`channel_section` text,
+	`app_view` text,
+	FOREIGN KEY (`window_id`) REFERENCES `browser_windows`(`id`) ON UPDATE no action ON DELETE cascade
+);--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS `browser_tabs_window_idx` ON `browser_tabs` (`window_id`);--> statement-breakpoint
+ALTER TABLE `browser_windows` ADD COLUMN `active_tab_id` text;--> statement-breakpoint
```
k11kirky, line 43: Heads up on the failure mode here: `migrate.ts` wraps *all* pending migrations in one `sqlite.transaction()`, and only tolerates `duplicate column name` (confirmed by `migrate.test.ts`'s "propagates errors other than duplicate-column conflicts" case). If a dogfooded DB's divergence doesn't fit this ALTER-ADD-COLUMN healing model (some other SQLite error), the whole batch rolls back and the exception propagates uncaught up through `DatabaseService.initialize()` into the boot chain. It's caught only by the global `uncaughtException`/`unhandledRejection` handlers in `apps/code/src/main/index.ts`, so the process survives, but later boot steps (auth, deep links, etc.) never run — a worse, harder-to-diagnose state than the dead "+" button this PR fixes. This is pre-existing, deliberately-tested runner behavior, not something this PR introduces, but 0020 is the first migration to exercise it against real-world divergent data, so it's worth double-checking that every affected 0013 variant is covered before shipping.

packages/workspace-server/src/services/browser-tabs/service.ts
```diff
@@ -113,6 +131,7 @@ export class BrowserTabsService
     const providedId = input.tabId;
     const { snapshot } = openOrFocusTab(this.snapshot, {
       ...input,
+      windowId: this.resolveWindowId(input.windowId),
```
k11kirky, line 138: `resolveWindowId` heals a stale/unknown `windowId` here and in `newBlankTab` (line 149), but `setOrder` (line 176) and `setActiveTab` (line 182) still pass `input.windowId` straight through to `setTabOrder`/`setWindowActiveTab` in `packages/shared/src/browser-tabs.ts`, which silently no-op when the window doesn't match. So a mirror desynced on windowId (the exact scenario this PR heals for tab creation) still silently drops a reorder or a tab-focus mutation. Worth either extending the healing to those two, or a short comment explaining why creation paths are the only ones that need it.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
```diff
@@ -750,16 +758,43 @@ export function BrowserTabStrip() {
...
+    // (the server always has a primary window) and append into it. Resolve
+    // the window from the FETCHED snapshot, not the mirror: reseedMirror
+    // skips the store apply when a local write or newer remote push raced
+    // the fetch, and the mirror could still be windowless then.
+    void reseedMirror()
+      .then((server) => {
+        const win = server
+          ? primaryWindow(server)
+          : primaryWindow(readMirror());
+        if (win) createBlankTab(win.id);
```
k11kirky: Narrow edge case: if `reseedMirror()` resolves with `server === null` (no fetcher registered — `BrowserTabsEventsContribution.start()` never ran) and `readMirror()` is also windowless, `win` is `undefined` and this line silently skips with no log/error. Given `registerSnapshotFetcher` is called synchronously inside `start()`, this should be effectively unreachable once boot runs normally — but since it reproduces exactly the dead-"+" symptom this PR sets out to fix, might be worth a `logger.error` fallback here too, matching the `.catch` below.

packages/ui/src/router/routes/__root.tsx
```diff
@@ -373,8 +374,9 @@ function RootLayout() {
               noops on param-less routes (inbox, agents, new-task), so it's safe
               to mount everywhere. */}
           <BrowserTabStrip />
-          {channelsEnabled && (
-            <Flex align="center" className="no-drag ml-auto pr-3">
+          <Flex align="center" gap="2" className="no-drag ml-auto pr-3">
```
k11kirky: This `Flex` (and its `no-drag` class) used to only exist when `channelsEnabled` was true. Now it always renders. When `BILLING_FLAG` is off (`UsageButton` returns `null`) *and* `channelsEnabled` is false, this renders an empty `Flex` that still claims a `no-drag` rect in the title bar — silently shrinking the new native drag region for no visible reason. Might be worth gating the wrapper itself on `billingEnabled || channelsEnabled`, or accepting the small dead zone as intentional (worth a one-line comment either way).

packages/ui/src/features/browser-tabs/browser-tabs-events.contribution.ts
```diff
@@ -39,7 +50,46 @@ export class BrowserTabsEventsContribution implements Contribution {
...
+        applyRemoteSnapshot(snapshot);
+        return;
+      } catch (error) {
+        if (signal.aborted) return;
+        if (attempt === SEED_ATTEMPTS) {
+          this.logger.error("browser-tabs snapshot seed failed", { error });
+          return;
+        }
+        await new Promise<void>((resolve) => {
+          const timer = setTimeout(
```
k11kirky: `packages/shared/src/backoff.ts` already has `getBackoffDelay`/`sleepWithBackoff` (used by core/auth, oauth, sessionService, autoresearch) computing the same `initialDelayMs * multiplier ** attempt` formula. This reimplements it locally because `sleepWithBackoff` has no `AbortSignal` support — might be worth adding that to the shared helper instead, so the abort-cancelable sleep pattern isn't reinvented per-caller.

packages/ui/src/features/billing/UsageButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+            <Circle
+              size={4}
+              weight="fill"
+              className="mx-1.5 inline text-muted-foreground"
+            />
+            <span className="font-normal text-muted-foreground">
+              {exceeded ? "Limit reached" : `${usagePercent}% used`}
+            </span>
+          </span>
+          <button
```
k11kirky: Per CLAUDE.md, "CTAs are quill `Button`s: primary action `variant=\"primary\"`, secondary `variant=\"outline\"`" — this "Upgrade" CTA is a raw `<button>` with hand-rolled classes (carried over unchanged from the deleted `SidebarUsageBar.tsx`), while the rest of this file uses quill primitives throughout (`Popover`, `Progress`, the trigger `Button`). Worth swapping to a quill `Button` (e.g. `variant="ghost"`/`size="sm"`) for consistency and to pick up quill's focus/hover states for free.

k11kirky (commented):
Automated review pass (high effort, 8 finder angles + verification). Overall the fix is solid and well-tested for the reported symptom; flagging a few gaps in the fix's scope and one cleanup nit. Nothing here blocks merge in my view, but worth a look before shipping to the affected dogfood population.

k11kirky:
One more from the same review pass, not inline-able since the file isn't part of this diff: `packages/ui/src/features/browser-tabs/BrowserTabsDnd.tsx`'s `onDragStart`/`onDragOver`/`onDragEnd` (lines 40, 61, 89) all do `if (!win) return;` when `primaryWindow(snapshot)` is undefined — the same windowless-mirror dead end this PR fixes for the "+" button, just untreated for drag-and-drop. Not a regression from this PR, but worth a follow-up if the windowless-mirror state is expected to recur (e.g. new profiles created before boot finishes).

packages/workspace-server/src/db/migrations/0020_repair_browser_tabs_schema.sql
adamleithp: Fixed in 0e104ba: 0020 is now registered as a best-effort repair migration in migrate.ts — its statements tolerate ANY SQLite error, so an unforeseen divergence degrades to the pre-repair status quo instead of rolling back the batch and killing boot. Covered in migrate.test.ts: panes-era lineage, tables-missing lineage, and a pathological browser_tabs-as-VIEW variant (which emits 'views may not be indexed' / 'Cannot add a column to a view' — exactly the class the strict runner would have died on).

packages/workspace-server/src/services/browser-tabs/service.ts
adamleithp: Comment added in 0e104ba on resolveWindowId: healing is deliberately creation-only. A desynced mirror's setOrder/setActiveTab carries stale TAB ids too, so retargeting them at the primary window would apply wrong state — the shared transforms no-op safely and the snapshot reconcile heals the mirror. Creating a tab is window-independent intent.

packages/ui/src/features/browser-tabs/BrowserTabStrip.tsx
adamleithp: Fixed in 0e104ba: the no-window-after-reseed case now logs 'browser-tabs: new-tab found no window after reseed' instead of silently skipping.

packages/ui/src/router/routes/__root.tsx
adamleithp: Fixed in 0e104ba: the right-side group is gated on billingEnabled || channelsEnabled with a comment, so an empty Flex can't claim a no-drag rect.

packages/ui/src/features/browser-tabs/browser-tabs-events.contribution.ts
adamleithp: Fixed in 0e104ba: sleepWithBackoff now takes an optional AbortSignal (clears the timer and resolves on abort), and the seed loop uses the shared helper.

packages/ui/src/features/billing/UsageButton.tsx
adamleithp: Fixed in 0e104ba: Upgrade is now a quill Button variant=link size=sm — verified live it picks up --primary and the hover underline from quill.

k11kirky:
Re-reviewed the two follow-up commits (`0e104ba7`, `63f69217`). All six inline findings from the last pass are addressed, and well-tested:

- **Migration boot-abort risk** → fixed properly: `migrate.ts` now has an explicit `BEST_EFFORT_MIGRATIONS` set (keyed by the exact 0020 timestamp from `_journal.json`) that tolerates *any* error for listed migrations, not just duplicate-column. New `migrate.test.ts` cases cover the panes-era variant, a variant missing tables entirely, and a deliberately unhealable case (`browser_tabs` as a VIEW) that now degrades to "still broken" instead of killing boot. This is a nice generalization — turns the one-off repair migration into a documented, tested pattern for future repair migrations.
- **`resolveWindowId` asymmetry** → not code-changed, but now has a clear comment explaining why `setOrder`/`setActiveTab` are deliberately excluded (they carry stale *tab* ids too, so retargeting would apply wrong state; the shared transforms already no-op safely). Reasonable, and resolves my concern.
- **Silent no-op in `handleNewTab`'s fallback** → fixed, now logs `logger.error` when no window is found after reseed instead of silently skipping.
- **Title-bar drag-region regression** → fixed, the wrapper `Flex` is now gated on `billingEnabled || channelsEnabled` so an empty group can't claim a no-drag rect.
- **Backoff reuse** → fixed, `sleepWithBackoff` in `packages/shared/src/backoff.ts` now takes an optional `AbortSignal` and `seedWithRetry` uses it instead of hand-rolling the timer/listener logic.
- **Raw `<button>` CTA** → fixed, now a quill `Button variant="link"`. Bonus: the trigger button was also converted to `render={<Link .../>}` per the CLAUDE.md link-CTA convention (there's existing precedent for `Button render={<Link/>}` in `ChannelTabs.tsx`, though nesting it inside `PopoverTrigger`'s own `render` prop is new — worth a quick visual check that click/hover still behave through that triple-nested composition, just to be safe).

The `BrowserTabsDnd.tsx` gap from my general comment above is unaddressed, but that's expected — it's a pre-existing, non-regressing follow-up item, not something this PR needs to fix.

Traced the `openSettings`/`prepareSettingsPage` split against every other call site (`billing.contribution.ts`, `UsageLimitModal.tsx`, `SignalSourcesSettings.tsx`, `WorkspaceModeSelect.tsx`, `ProjectSwitcher.tsx`, `UsageView.tsx`, `TaskInput.tsx`) — all still call `openSettings` unchanged, so the extraction is safe. No new issues found. CI's `unit-test`/`typecheck`/`build` were still pending at review time — worth confirming green before merge given the new migrate.ts logic path.

k11kirky (approved):

PR #3350, opened 2026-07-10
Clear channel composer instantly and show optimistic kickoff in feed

raquelmsmith:
## Problem

In a Bluebird channel, submitting a task had a visible lag: the composer only cleared and the card only appeared *after* the task-creation round trip (and its feed invalidate/refetch) resolved. Because the input still showed the typed prompt, it looked like the keystroke hadn't registered, so people pressed enter twice.

**Why:** Raquel reported hitting enter twice because the input didn't clear and the message didn't show up in the channel messages immediately — the ask was to clear the input right away, surface the message in the feed instantly, and show a loading state while the task actually "takes" rather than leaving text sitting in the box.

## Changes

- The channel composer now clears the instant a submit is accepted, snapshotting the editor content and passing it to `handleSubmit` as an override so clearing early can't race the content read.
- An optimistic "Starting…" kickoff row (the user's message + a spinner card) is posted to the feed immediately, then swapped for the real card once the task is created.
- On creation, the real task is spliced straight into the feed cache instead of waiting on the invalidate refetch/poll.
- On failure, the optimistic row is pulled and the prompt is handed back to the composer so the user can retry.

Scoped to the channel feed surface (`WebsiteChannelHome` / `ChannelHomeComposer` / `ChannelFeedView`).

## How did you test this?

- `pnpm --filter @posthog/ui typecheck` — passes.
- `biome check` on the changed files — passes.
- Not yet exercised in the running app; please pull down and try submitting in a channel.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/canvas/components/ChannelFeedView.tsx
+++ b/packages/ui/src/features/canvas/components/ChannelFeedView.tsx
@@ -33,6 +33,7 @@ import {
   ThreadItemRepliesLabel,
   ThreadItemRepliesMeta,
   ThreadItemTimestamp,
+  useChatMessageScroller,
 } from "@posthog/quill";
 import { formatRelativeTimeShort, getLocalDayDiff } from "@posthog/shared";
 import type { Task, TaskRunStatus } from "@posthog/shared/domain-types";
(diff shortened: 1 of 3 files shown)


raquelmsmith:
tested locally, it works

PR #3354, opened 2026-07-10
feat(stamphog): route PR review through the ai-gateway

brandonleung:
## Problem

Stamphog (the vendored PR-approval agent) calls Anthropic directly, so its LLM usage bypasses PostHog's ai-gateway and shows up as pre-gateway `stamphog` traffic on the cutover dashboards. This routes it through the gateway, matching the other migrated products.

## Changes

- Add `gateway.py`: resolves `AI_GATEWAY_URL`/`AI_GATEWAY_API_KEY`, points the Claude Agent SDK at the gateway via `ANTHROPIC_*` env, tags `aio_stamphog`. Verbatim copy from the posthog monorepo `tools/pr-approval-agent`.
- `reviewer.py` uses the plain SDK query in gateway mode so the gateway's own `$ai_generation` is not double-counted; falls back to the traced query (then the plain query) when the gateway is not configured.
- Workflow passes `AI_GATEWAY_URL`/`AI_GATEWAY_API_KEY` from `STAMPHOG_AI_GATEWAY_*` secrets to the review step.

Env-gated: both secrets set and valid routes through the gateway; unset or malformed falls back to direct Anthropic.

## Not in this PR

- Populating the two GitHub secrets (needs the gateway URL and a `phs_` project secret).
- The other stamphog surfaces (MLHog, the monorepo) and summarization; those are tracked separately.

## Rollout

Set repo/org secrets `STAMPHOG_AI_GATEWAY_URL` (gateway URL ending in `/v1`) and `STAMPHOG_AI_GATEWAY_API_KEY` (`phs_` project secret) to arm the cutover; until then it stays on direct Anthropic. Rollback: clear either secret.

## How did you test this?

`pytest test_gateway.py test_reviewer_gateway.py` (16 tests): gateway config validation and fallback, `/v1` stripping, the `X-PostHog-Properties` attribution blob, and the routing guard that selects the plain query in gateway mode so the gateway and traced paths never both emit. The live path (a real PR review through the gateway) needs the secrets populated in CI and has not run yet.

Authored by Claude (Opus 4.8); requires human review.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/.github/workflows/pr-approval-agent.yml
+++ b/.github/workflows/pr-approval-agent.yml
@@ -67,6 +67,10 @@ jobs:
                   ANTHROPIC_API_KEY: ${{ secrets.STAMPHOG_ANTHROPIC_API_KEY }}
                   POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_TOKEN }}
                   GH_TOKEN: ${{ steps.app-token.outputs.token }}
+                  # Set both to route through the ai-gateway (gateway.py overrides
+                  # ANTHROPIC_* at runtime). Unset means direct Anthropic.
+                  AI_GATEWAY_URL: ${{ secrets.STAMPHOG_AI_GATEWAY_URL }}
+                  AI_GATEWAY_API_KEY: ${{ secrets.STAMPHOG_AI_GATEWAY_API_KEY }}
               run: |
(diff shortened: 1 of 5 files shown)


charlesvien (approved):

charlesvien:
exciting!

PR #3360, opened 2026-07-10
fix(canvas): name the canvas before creating its generation task

k11kirky:
## Problem

Every canvas generation task kicked off from the channel composer's Canvas mode (#3348) gets the title `Generate canvas "Untitled canvas"`, and that placeholder also shows as the first prompt bubble before the agent starts. The composer creates the dashboard with the placeholder name at submit time, and auto-naming only ran *after* the task was created — renaming the dashboard but never the task, its backend description, or the generation prompt's header.

## Changes

- `useGenerateFreeformCanvas`: for a still-unnamed canvas, generate the name from the instruction **before** creating the task (concurrently with model resolution, so the common path adds no latency). The real name now feeds the task title/description, the prompt header, the channel filing, and the completion toast; the dashboard rename is fired best-effort. Falls back to the placeholder if name generation fails, as before.
- Removed the now-unused post-task rename block and the tracker store's `updateName`.

## How did you test this?

- `pnpm --filter @posthog/ui typecheck` — clean
- `biome check` on the changed files — clean
- `pnpm --filter @posthog/ui test` — 164 files / 1458 tests passed

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/canvas/hooks/useGenerateFreeformCanvas.ts
+++ b/packages/ui/src/features/canvas/hooks/useGenerateFreeformCanvas.ts
@@ -115,6 +115,17 @@ export function useGenerateFreeformCanvas(args: {
       } = opts;
       setIsStarting(true);
       try {
+        // Auto-name a still-unnamed canvas from its generation prompt BEFORE
+        // creating the task: the name feeds the task title, the backend task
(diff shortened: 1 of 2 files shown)


charlesvien:

PR #3363, opened 2026-07-11
feat: add managed Secure Enclave commit signing

robbie-c:
## Summary
- add an opt-in macOS Secure Enclave signing broker and bundled `ssh-keygen` compatibility wrapper
- add **Settings → Commit signing** to enable managed signing, display the generated public key, copy it, and open GitHub's SSH key settings
- persist enablement, configure future local Claude/Codex sessions, and cleanly restore the prior Git/SSH environment when disabled
- sign the native helper with `APPLE_CODESIGN_IDENTITY` in local builds and show a development-only local code-signing prerequisite when entitlements are unavailable
- document setup, behavior, security boundaries, and migration from Secretive

## Security boundary
- pass the broker control secret over stdin instead of process arguments
- accept control requests only from the workspace-server parent process using the Unix peer PID
- authorize SSH signing only for registered agent processes and their descendants
- stop newly spawned agents until registration completes, resume them on success or failure, and revoke access after process exit
- retain one random broker-lifetime SSH socket path to avoid concurrent-session environment races

## Test plan
- [x] `pnpm --filter @posthog/workspace-server exec vitest run src/services/signing-access/service.test.ts src/services/agent/agent.test.ts` (36 tests)
- [x] `pnpm --filter @posthog/workspace-server typecheck`
- [x] `pnpm --filter @posthog/code typecheck`
- [x] native helper compilation with the production framework flags
- [x] `pnpm --filter @posthog/code compile`
- [x] `node scripts/check-host-boundaries.mjs`
- [x] pre-commit `pnpm typecheck`

## QA
- found and fixed an immediate-exit race that could register a PID after its revoke request
- found and fixed a failed-`SIGSTOP` path that could authorize an unstopped process
- verified both cases with focused regression tests

## Local development
- Secure Enclave key persistence requires configured Apple code signing; ad-hoc signatures fail with OSStatus `-34018`

## Notes
- add the displayed public key to GitHub twice: as an **Authentication Key** and a **Signing Key**
- remote skills downloads emit non-fatal offline warnings during local Electron builds when GitHub cannot be resolved

---


charlesvien:

PR #3370, opened 2026-07-12
feat(canvas): context onboarding with a slack-style channel intro

adamleithp:
## Problem

Contexts had no guided creation flow and their feed was task cards only — no lifecycle announcements, no sense of who created a context or how to give it a CONTEXT.md.

## Changes

**Slack-style channel intro** (`ChannelIntro`, new)
- Every public context's feed opens with a pinned intro: the context name, "@Ann created this context today. This is the very beginning of the mobile context." (creator name rendered as a mention chip, inert for now), derived entirely from the backend channel row — so it renders identically for every teammate and for pre-existing contexts.
- Below it, a context.md card with two states: a pressable "Create a context.md" call to action (lucide `file-plus-corner`) that opens the describe dialog, flipping to an inert `tone=success` "Created context.md" card (`file-check-corner`) once instructions exist.
- A "Learn more about contexts" card sits alongside (placeholder — not wired to a destination yet).

**Create dialog split into two modes** (`CreateChannelModal`)
- Create mode is name-only: creates the context and lands you in its feed. No session starts at creation; onboarding continues from the intro card.
- Describe mode ("Create your context.md", opened from the intro card or the CONTEXT.md empty state's "Build with agent") is a single textarea that seeds the plan-mode session which co-builds CONTEXT.md and publishes via the PostHog MCP.

**Feed opener + system rows** (`ChannelFeedView`, `useChannelFeedMessages`)
- The derived opener row is now the Slack-style "joined mobile", rendered as the creator (initials avatar + name) instead of the robot "PostHog / Agent" chrome; agent lifecycle rows ("is building CONTEXT.md…") keep the agent chrome.
- `ChannelFeedView` gains an `intro` slot rendered above the first entry; when set, the feed renders even with zero entries, so new public contexts show intro + Today marker + joined row instead of the suggestions empty state (which remains for the personal "me" context).
- "Ann is building CONTEXT.md for X" is posted as a durable feed message against the backend channel. Requires posthog#70320 server-side; silently no-ops until it deploys.
- Server-emitted `channel_created`/`context_created` rows are filtered out so nothing duplicates once the backend lands.

**Rename** — channels are now "contexts" across the UI copy (sidebar, dialogs, menus).

<details>
<summary>Screenshots from earlier iterations of this PR (create dialog has since dropped the description field; the feed now opens with the intro)</summary>

image
image
image
image

</details>

Known follow-ups:
- "Learn more about contexts" card needs a destination.
- The context.md card doesn't yet reflect an in-flight plan session (a second click can start a duplicate session).

## How did you test this?

- Manual: created a context in the desktop app; verified the name-only dialog, the intro (name, mention-chip creation line, Today marker, joined row), the create-card → describe dialog → plan session flow, and the task card appearing in the feed.
- `pnpm --filter @posthog/ui typecheck` and Biome clean; lint-staged (biome + typecheck) passed on commit.

## Automatic notifications

- [ ] Publish to changelog?


packages/ui/src/features/canvas/hooks/useGenerateContext.ts
```diff
@@ -1,67 +1,159 @@
...
-            // test a local build of these features before merging.
+            content: buildContextGenerationPrompt({
+              channelName,
+              channelId,
+              description,
+            }),
+            taskDescription: `Build CONTEXT.md for ${channelName}`,
             workspaceMode,
+            adapter: adapter ?? "claude",
+            model: currentModel,
```
adamleithp: **[correctness, CONFIRMED]** `currentModel` is `undefined` until `usePreviewConfig`'s async tRPC fetch resolves — and permanently when no cloud region is set — and `canSubmit` in the dialog never gates on it. A user clicking "Create & plan" in that window submits a cloud task with no model, which the cloud runtime rejects: the context is created but orphaned, with only a generic toast.

`useGenerateFreeformCanvas` already solves exactly this: it falls back to `modelResolver.resolveDefaultModel` and hard-stops with a "No model is configured for cloud runs" toast when none resolves. Do the same here.

packages/ui/src/features/canvas/hooks/useChannelFeedMessages.ts
```diff
@@ -0,0 +1,93 @@
...
+export function useChannelFeedMessages(channelId: string | undefined): {
+  messages: ChannelFeedSystemMessage[];
+  isLoading: boolean;
+} {
+  const query = useAuthenticatedQuery<ChannelFeedMessage[]>(
+    channelFeedMessagesQueryKey(channelId),
+    (client) => client.getChannelFeed(channelId as string),
+    {
+      enabled: !!channelId,
+      refetchInterval: CHANNEL_FEED_MESSAGES_POLL_INTERVAL_MS,
```
adamleithp: **[efficiency, CONFIRMED]** Against prod this endpoint 404s today (posthog#70320 not deployed), `getChannelFeed` throws on non-ok, and this is a numeric `refetchInterval` with TanStack's default `retry: 3` (the app QueryClient only overrides `staleTime`/`refetchOnWindowFocus`). TanStack keeps polling numeric intervals while the query is in error state — so every open channel view fires ~4 requests of 404 every 5s, forever.

Pass `retry: false` and make the interval a function that stops on error, e.g. `refetchInterval: (q) => (q.state.status === "error" ? false : CHANNEL_FEED_MESSAGES_POLL_INTERVAL_MS)`, or gate polling until the first success.

packages/ui/src/features/canvas/components/WebsiteContext.tsx
```diff
@@ -415,144 +328,54 @@ function EmptyState({
...
           <QuillButton variant="primary" size="default" onClick={onCreate}>
-            Create
+            Write it myself
           </QuillButton>
-          <GenerateWithAgent
-            channelId={channelId}
-            channelName={channelName}
-            regenerate={!!stoppedTaskId}
-          />
+          <GenerateWithAgent channelId={channelId} channelName={channelName} />
```
adamleithp, line 342: **[correctness/regression, CONFIRMED]** Two behaviors the deleted `useFolderGenerationTask` flow provided are not re-established here:

1. **No live refresh**: the old code polled the published file every 5s while a generation ran; `useFolderInstructions` has no `refetchInterval`, so a user sitting on this tab never sees the agent's published CONTEXT.md until they navigate away and back.
2. **No double-start guard**: the old tab rendered a `GeneratingState` instead of the button while a shared server-side generation task ran ("any user sees an in-progress generation and won't double-start it"). Now "Build with agent" is unconditional, so a teammate (or a second window) can start a competing plan session for the same context.

The durable `context_md_building` feed row only shows in the channel-home feed, not on this tab — it could double as the guard signal here.

packages/ui/src/features/canvas/components/CreateChannelModal.tsx
```diff
@@ -1,152 +1,252 @@
...
   const [wasOpen, setWasOpen] = useState(open);
   if (open !== wasOpen) {
     setWasOpen(open);
-    if (open) setName("");
+    if (open) {
+      setName("");
+      // Restore a description stashed by a prior failed launch (create mode
+      // only), otherwise start blank. Consuming clears the stash.
+      setDescription(isDescribeMode ? "" : stashedDescription);
+      stashedDescription = "";
```
adamleithp: **[correctness, CONFIRMED]** The stash's only write path is the create-mode launch failure, which navigates to the context home — where the natural retry is the CONTEXT.md empty state's "Build with agent", which reopens this component in **describe mode**. This open branch then blanks the field (`isDescribeMode ? ""`) *and* clears `stashedDescription` without consuming it, so the advertised restore never survives the natural flow. And because the variable is module-global and keyed to nothing, a stash from failed context A can pre-fill a later, unrelated create dialog for context B.

Consider only consuming/clearing the stash in create mode, and keying it (or better: pre-filling the describe-mode dialog with it, since that's where the user actually retries).

packages/ui/src/features/canvas/components/ChannelFeedView.tsx
```diff
@@ -472,31 +473,97 @@ function FeedRow({
...
+        task,
+      })),
+      ...(systemMessages ?? []).map((message) => ({
+        kind: "system" as const,
+        id: message.id,
+        createdAt: message.createdAt,
+        message,
+      })),
+    ];
+    merged.sort((a, b) => (a.createdAt < b.createdAt ? -1 : 1));
```
adamleithp: **[correctness, PLAUSIBLE]** This comparator is inconsistent: equal timestamps return `1` both ways (`compare(a,b) === compare(b,a) === 1`), which defeats the ES2019 stable-sort guarantee for ties — and ties are realistic here (the -1ms offset in `useGenerateContext` exists precisely because a system row and its task can collide; server second-precision truncation too). Tied entries get an arbitrary rather than insertion-order arrangement.

`merged.sort((a, b) => a.createdAt.localeCompare(b.createdAt))` is both consistent and chronological for ISO strings.

packages/ui/src/features/canvas/hooks/useGenerateContext.ts
```diff
@@ -1,67 +1,159 @@
...
+  description: string;
+  workspaceMode?: WorkspaceMode;
+}
+
+// The task's backend channel is what the context feed queries (getTasks by
+// channel), while the folder id drives Recents/Artifacts. Resolve the folder's
+// display name onto its backend channel the way useBackendChannel does: the "me"
+// folder maps to the personal channel; any other name resolve-or-creates its
+// public channel. Without this the task only appears in Recents, not the feed.
+async function resolveBackendChannelId(
```
adamleithp: **[reuse/altitude, CONFIRMED]** This re-implements the folder-name → backend-channel mapping that `useBackendChannel` already encodes (its own comment admits it: "the way useBackendChannel does"). Two copies of the `me`-vs-public branch + normalization will drift on the next mapping change. Worth extracting a shared imperative resolver next to `useBackendChannel`.

Deeper issue both copies share: the mapping is keyed on **display name**, so renaming a context orphans its feed — the next generate resolve-or-creates a fresh empty backend channel while all prior tasks/announcements stay on the old one. Fine to punt, but worth a tracking note; the channels-architecture plan's workstream A (unified channel identity) is the real fix.

packages/ui/src/features/canvas/hooks/useGenerateContext.ts
```diff
@@ -1,67 +1,159 @@
...
 
+  // Resolve adapter + model the same way the composer does. A cloud run rejects
+  // a task with no model ("required when selecting a cloud runtime"), so this
+  // headless launch must supply one rather than relying on a composer's picker.
+  const adapter = useSettingsStore((s) => s.lastUsedAdapter);
+  const { modelOption } = usePreviewConfig(adapter);
+  const currentModel =
+    modelOption?.type === "select" ? modelOption.currentValue : undefined;
+
   const generate = useCallback(
```
adamleithp, line 68: **[conventions, CONFIRMED]** Root `CLAUDE.md` rule 5: "Hooks wrap exactly one query, mutation, subscription, or store selector. Multi-source orchestration belongs in a service method." `generate` chains channel resolution → `createTask` → `fileTask` → feed-message post → three cache invalidations. The task-creation orchestration (resolve channel, create task, announce) belongs in a `@posthog/core` service method; the hook should wrap that one call plus the query invalidations.

packages/ui/src/features/canvas/components/WebsiteChannelHome.tsx
```diff
@@ -131,12 +152,17 @@ export function WebsiteChannelHome({ channelId }: { channelId: string }) {
...
-      <div className="text-center">
-        <h1 className="font-semibold text-2xl text-gray-12 tracking-tight">
-          {channelName ? `Welcome to #${channelName}` : "Welcome"}
-        </h1>
-        <Text className="mt-2 block text-[13px] text-gray-10">
-          Every message kicks off a task the whole channel can follow.
+      <div className="flex flex-col items-center gap-2 text-center">
+        <Heading className="font-bold text-2xl">
+          {channelName === PERSONAL_CHANNEL_NAME
+            ? "Welcome to your own personal context"
```
adamleithp, line 199: **[conventions, CONFIRMED]** Root `CLAUDE.md`: "Empty/placeholder/loading screens (canvas and elsewhere) are a `@posthog/quill` `<Empty>` (`EmptyHeader` → `EmptyMedia variant=\"icon\"` → `EmptyTitle` → `EmptyDescription`, then `EmptyContent` for CTAs). Don't hand-roll the centered Flex + dashed icon box." This empty state is a hand-rolled centered flex with `Heading`/`Text` + custom suggestion cards.

packages/shared/src/domain-types.ts
```diff
@@ -79,6 +79,27 @@ export interface TaskChannel {
   created_by?: UserBasic | null;
 }
 
+/** Lifecycle events a channel-feed system message can announce. */
+export type ChannelFeedMessageEvent = "context_created" | "context_md_building";
```
adamleithp: **[simplification, CONFIRMED]** `context_created` is never posted anywhere in this repo (only `context_md_building` is), and `useChannelFeedMessages` explicitly filters it out via `CREATION_EVENTS`. Advertising it in the union invites a developer to post an event that gets silently dropped from the feed. Narrow the union to `"context_md_building"` until a second event ships.

raquelmsmith (approved):
I've throughly reviewed and found nothing that needed addressing

PR #3378, opened 2026-07-13
fix: compact boundary marker showing "undefined · ~NaNK tokens" on codex sessions

k11kirky:
## Problem

A user on a codex model (gpt-5.6-sol) saw the compaction marker render as **"Conversation compacted · undefined · ~NaNK tokens"**. The codex app-server adapter emits `_posthog/compact_boundary` with only `sessionId`, but the UI assumed `trigger` and `preTokens` (which only the Claude adapter sends) were always present, so it interpolated `undefined` and `NaN` straight into the marker.

## Why

Codex sessions compact regularly, so every codex user hitting auto-compaction sees this broken marker in the transcript.

## Changes

- `CompactBoundaryView` (and the item types feeding it) now treat `trigger`/`preTokens` as optional and drop the missing segments instead of rendering `undefined`/`NaN`.
- The codex adapter now reports `trigger: "auto"` (codex only self-compacts) and `preTokens` from its usage tracker's last context-occupancy reading, so the marker shows real data going forward.

## How did you test this?

- `vitest run codex-app-server-agent` in `packages/agent` (75 passed), including an updated boundary test asserting `trigger`/`preTokens`.
- `vitest run buildConversationItems` in `packages/ui` (32 passed).
- `tsc --noEmit` on `@posthog/agent` and `@posthog/ui`; biome check on changed files (only pre-existing warnings).

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/agent/src/adapters/codex-app-server/codex-app-server-agent.test.ts
+++ b/packages/agent/src/adapters/codex-app-server/codex-app-server-agent.test.ts
@@ -1810,6 +1810,9 @@ describe("CodexAppServerAgent", () => {
     } as unknown as NewSessionRequest);
 
     // The compaction item brackets it: started → in progress, completed → boundary.
+    stub.emit("thread/tokenUsage/updated", {
+      tokenUsage: { last: { inputTokens: 90000, totalTokens: 120000 } },
+    });
     stub.emit("item/started", {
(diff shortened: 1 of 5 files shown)


charlesvien:

PR #3384, opened 2026-07-13
fix(sessions): keep auto task titles pinned to the original prompt

k11kirky:
## Problem

Tasks sometimes get renamed based on the most recent messages in the conversation instead of keeping the name generated from the original prompt.

The only writer that rewrites a task title after creation is `useChatTitleGenerator`: its interval and catch-up fires (`REGENERATE_INTERVAL`) generate from the last 7 prompts and write the result as the task title whenever the title isn't locked (`title_manually_set`). Tasks created server-side (cloud creation, warm-run activation, Slack-origin tasks) get an auto-generated title with `title_manually_set = false`, so they are exactly the ones exposed. #3315 removed the remount-driven refires, but any surviving prompt-window fire still renames these tasks. Task `description` is never rewritten anywhere; the rename is the whole symptom.

## Changes

- Add `canApplyTitleFromPrompts` in `chatTitle.ts`: prompt-based fires past the first prompt may no longer write the title, unless the current title is still the raw-description placeholder (so the retry path for failed initial generations keeps working).
- Gate the title write in `useChatTitleGenerator` with it; summary generation is unchanged, so conversation summaries (used for commit/PR context) still refresh at every interval.

## How did you test this?

Red-green: added a failing test reproducing the rename (unlocked task with a real title at prompt count 8 gets retitled from recent prompts), then implemented the gate and confirmed it passes. Also added coverage for the placeholder-retry path. Ran the full sessions suites in `@posthog/core` (274 tests) and `@posthog/ui` (365 tests), `biome lint`, and `tsc --noEmit` on both packages — all green.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/sessions/chatTitle.test.ts
+++ b/packages/core/src/sessions/chatTitle.test.ts
@@ -1,6 +1,7 @@
 import type { Task } from "@posthog/shared/domain-types";
 import { describe, expect, it } from "vitest";
 import {
+  canApplyTitleFromPrompts,
   decideTitleGeneration,
   formatPromptsForTitleInput,
   getFallbackTaskTitle,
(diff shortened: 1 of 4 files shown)


tatoalo (approved):
thanks!

PR #3396, opened 2026-07-13
fix(code): never lose a local task's initial prompt

frankh:
## Problem

When creating a **local** task, the initial prompt was only ever held in memory. If the agent hadn't produced its first response yet and the user looked away — the app was backgrounded, reloaded, crashed, or a transient connect/`sendPrompt` failure silently exhausted its retries — the prompt was lost and the task never started.

Root cause (`packages/core/src/sessions/sessionService.ts`):
- `createNewLocalSession` stored the prompt only on the in-memory zustand session and delivered it via `sendPrompt`; the session store has no persistence.
- `client.createTaskRun` was called with no options, so the deliverable prompt never reached the server either.
- On reload the store is wiped; `task.latest_run.id` still exists, so the resume path (`routeLocalConnect` → `resume-existing` → `reconnectToLocalSession`) never re-sent the prompt.

## Fix

Persist the prompt durably in the workspace-server `task_metadata` SQLite table, keyed by task id, and wire re-send/clear into the local session flow:

- **Schema**: nullable `pendingInitialPrompt` column on `task_metadata` + migration `0017`.
- **workspace-server / host-router**: repo getter + patch field, `WorkspaceMetadataService` set/get/clear passthroughs, Zod schemas, and one-line tRPC forwards.
- **core `sessionService`**:
  - Write the prompt (JSON `ContentBlock[]`) right after the task run is created, before spawning the agent — never aborting creation on a persistence failure.
  - On resume, `resendPendingPromptIfNeeded` re-sends the stored prompt **unless** the replayed log already contains the `session/prompt` echo (the agent's "received it" signal, preventing double-send).
  - Clear the durable copy on the prompt echo in `handleSessionEvent`, independent of in-memory state.
  - `clearSessionError` falls back to the durable copy so **Retry** and the silent auto-retry path recover after a reload.

Cloud tasks are unaffected — they self-fetch their prompt server-side and never enter `createNewLocalSession`.

## Testing

- `pnpm typecheck` (core, workspace-server, host-router): clean.
- Biome lint on all three packages: clean, zero `noRestrictedImports`.
- Full suites pass (**2380 tests**), including new `task-metadata-repository.test.ts` (round-trip / clear / non-clobbering) and `initialPromptPersistence.test.ts` (echo-detection, resend-once, clear-on-echo, durable retry fallback).

---

### DIFF
--- a/packages/core/src/sessions/initialPromptPersistence.test.ts
+++ b/packages/core/src/sessions/initialPromptPersistence.test.ts
@@ -1,0 +1,208 @@
+import type { ContentBlock } from "@agentclientprotocol/sdk";
+import type { AcpMessage, AgentSession } from "@posthog/shared";
+import { describe, expect, it, vi } from "vitest";
+import { SessionService, type SessionServiceDeps } from "./sessionService";
+
+const TASK_ID = "task-1";
+const RUN_ID = "run-1";
+
+const PROMPT: ContentBlock[] = [{ type: "text", text: "do the thing" }];
+
+function promptEcho(): AcpMessage {
+  return {
+    type: "acp_message",
+    ts: 0,
+    message: {
+      jsonrpc: "2.0",
+      id: 1,
+      method: "session/prompt",
+      params: {},
+    },
+  } as unknown as AcpMessage;
+}
+
+function createHarness(
+  overrides: {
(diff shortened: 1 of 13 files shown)


charlesvien:

PR #3404, opened 2026-07-13
fix(code): make postinstall cross-platform

trippyogi:
﻿## Summary

Replaces the Code app's Bash-based postinstall script with a cross-platform Node implementation so `pnpm install` works on Windows without WSL or Git Bash.

The new script preserves the existing postinstall behavior:

- repairs a missing Electron distribution
- rebuilds `better-sqlite3` against Electron's ABI
- restores the `node-pty` spawn-helper execute bit on POSIX systems
- patches the Electron development app name on macOS
- downloads the bundled agent binaries

It also adds Windows development notes and troubleshooting guidance.

**Behavior note:** the old shell script always invoked `patch-electron-name.sh`, which then no-oped when the macOS plist was absent. The Node script only invokes it when `process.platform === "darwin"`. Intentional correction, not a regression.

## Why

On a fresh Windows 11 checkout, `pnpm install` currently invokes:

```bash
bash scripts/postinstall.sh
```

Windows may resolve `bash` to the System32 WSL launcher even when no WSL distribution is installed, causing installation to fail with:

```text
execvpe(/bin/bash) failed: No such file or directory
```

Most of the postinstall workflow already invokes Node scripts, so moving the orchestration itself to Node removes the unnecessary shell dependency.

## Scope

This addresses the **installation** failure from PostHog/posthog#76200.

The default `pnpm dev` command still depends on phrocs, for which no Windows binary is currently published. Docs now recommend:

- Prefer: `pnpm dev:agent` and `pnpm dev:code` in separate terminals
- Alternative: `pnpm dev:mprocs`

Partially addresses PostHog/posthog#76200. Leaving the issue open for the remaining `pnpm dev` / phrocs Windows work unless maintainers decide the documented workaround is enough.

## Validation

Tested on Windows 11 (build 26200), Node 22.12.0, pnpm 10.23.0:

- [x] `node apps/code/scripts/postinstall.mjs` completes on an existing checkout (rebuilds `better-sqlite3` for Electron, downloads/skips agent binaries) without Git Bash / WSL
- [x] `apps/code` `postinstall` script points at `node scripts/postinstall.mjs`
- [x] Biome check on changed files; `pnpm lint` exits 0 (pre-existing warnings elsewhere)
- [x] Confirmed monorepo `pnpm typecheck` already fails on tip of `upstream/main` in `@posthog/agent` (unrelated to this change)
- [x] Clean install on this branch: deleted all workspace `node_modules`, then `pnpm install` (exit 0 in ~2m). `PATH` still prefers `C:\Windows\System32\bash.exe` (WSL stub); postinstall ran `node scripts/postinstall.mjs` and never invoked bash
- [x] Electron missing-dist self-heal during that clean install (`Electron binary missing at .../electron/dist — downloading...`), then sqlite rebuild + binary download; postinstall completed
- [ ] Packaged Windows app rebuilt against this branch
- [ ] macOS postinstall locally (expects CI / maintainer smoke)
- [ ] Linux postinstall locally (expects CI / maintainer smoke)

## Test plan

- [x] Fresh Windows tree: `pnpm install` after wiping `node_modules`, without relying on Git Bash/WSL
- [ ] Re-run `pnpm install` on an already-installed tree
- [ ] macOS: `pnpm install` still applies development Electron display name
- [ ] Linux: `pnpm install` still rebuilds native modules and downloads binaries
- [ ] Confirm `pnpm dev` / phrocs remains unchanged (out of scope)


charlesvien:

PR #2361, opened 2026-05-25
Add keyboard navigation and jump-to-message modal for long conversations

Basit-Balogun10:
> **Note:** This PR is stacked on #2321. The diff shows +2k lines because GitHub uses `main` as the base - cross-fork branches can't be set as PR bases (tried using Graphite to track the stack too, same limitation). The actual changes in this PR are ~1 commit on top of #2321. Please review with that context, or we wait for #2321 to be merged first.

## Problem

Long conversations become hard to navigate once they grow beyond a few turns. Search (`Ctrl+F`) can find text, but it cannot move between messages or jump to a specific turn in the conversation thread.

Closes #2360. Builds on #2321 to make the shortcuts configurable

## Changes

- Added conversation navigation shortcuts so users can move through user messages with `Alt/Option+Up` and `Alt/Option+Down`.
- Added a message jump picker to jump directly to a specific user message.
- Kept the selected message in view while navigating so the keyboard flow stays anchored in the thread.
- Exposed the new bindings in the shortcuts sheet so the mac equivalent is visible as `⌥↑` and `⌥↓`.
-  Added a date filter to the jump picker — a funnel icon trigger inside the search input opens a dropdown of presets (Today, Yesterday, Last 7 days, Last 14 days, This month, Last month, Last 30 days, Last 90 days, Last 6 months, This year, Last year) plus a "Custom range…" option that reveals datetime-local inputs. The footer message count shows the active filter in human-readable form, e.g. "3 messages (last 7 days)".

## How did you test this?

Validated the touched renderer files with targeted static checks (`get_errors`).

Manual verification flow:
1. Open a task with a long conversation thread.
2. On macOS, press `Option+Up` and `Option+Down` to move backward and forward through user messages.
3. On Windows/Linux, press `Alt+Up` and `Alt+Down` to verify the same navigation path.
4. Press `Cmd/Ctrl+J` to open the message jump picker.
5. Type part of a message, select a result, and confirm the conversation scrolls to that turn.
6. Open the keyboard shortcuts sheet and confirm the new bindings are listed with the correct mac glyphs.
7. Repeat the navigation while focus is inside the conversation editor to confirm the shortcuts still work in-context.
8. Open the jump picker on a conversation spanning multiple days, set a From/To date range, and confirm only messages within that window appear.
9. Confirm the clear (×) button resets the filter and the full list is restored.

## Demo

[...]

PR #2321, opened 2026-05-23
feat: configurable keyboard shortcuts

Basit-Balogun10:
## Problem

Users cannot remap keyboard shortcuts to suit their workflow or resolve conflicts with OS/app bindings on their machine. Closes PostHog/posthog#76329

### Core store (`keybindingsStore.ts`)

Persisted Zustand store backed by `electronStorage` — bindings survive app restarts. Stores custom overrides as `Partial<Record<ConfigurableShortcutId, string[]>>`.

- `addKeybinding` / `updateKeybinding` / `removeKeybinding` — per-binding CRUD, max 2 custom bindings per shortcut
- `resetShortcut` / `resetAll` — revert one or all shortcuts to defaults
- `resolveKey` — falls back to the static default when no custom entry exists
- `splitBindings` — splits comma-separated binding strings while correctly handling `mod+,` (the Settings shortcut key contains a literal comma — uses negative lookbehind regex `(?<!\+),` to distinguish separator commas from key commas)
- `findConflict` — checks the candidate key against all other configurable shortcuts (with live overrides applied) and all non-configurable static shortcuts; returns the conflicting description and whether it is a fixed shortcut

### Definitions (`keyboard-shortcuts.ts`)

- 20 of 28 shortcuts marked `configurable: true` with an explicit `CONFIGURABLE_SHORTCUT_IDS` array and `DEFAULT_KEYBINDINGS` record
- `eventToCombo` — converts a `KeyboardEvent` to a normalised combo string (`mod+shift+v`); requires at least one non-shift modifier; normalises `ArrowUp → up` etc.
- `tiptapEventToCombo` — like `eventToCombo` but also accepts shift-only combos, needed for `prompt-history-prev/next` whose defaults are `shift+up` / `shift+down`
- Platform-aware display formatting throughout: Mac shows `⌘⇧⌥↩↑↓`, Windows shows `Ctrl+Shift+Alt+Enter+↑+↓`. This also fixed a bug where hardcoded shortcut labels in the editor (e.g. the paste-as-file hint toast) were showing the Mac `⌘` symbol on Windows instead of `Ctrl`

### Shortcut recorder UI (`ShortcutRecorder.tsx`)

`ShortcutRecordingModal` — a fullscreen capture overlay:
- Window-level `keydown` listener (no element focus required)
- Live conflict detection with amber highlight and description message
- Enter to confirm (blocked when conflict exists), Escape to cancel, click-backdrop to dismiss
[...]

PR #3406, opened 2026-07-13
feat(sessions): keyboard navigation and jump-to-message picker

### LINKED ISSUE #2360: Add keyboard navigation and jump-to-message modal for long conversations
Long conversations are hard to navigate once they grow past a few turns. Search (Ctrl+F) can find text, but it cannot move between messages or jump to a specific turn in the conversation thread.

This feature should add conversation navigation shortcuts so users can move through messages quickly, including the mac equivalent of Alt+Up/Down (Option+Up/Down) and a jump picker for direct navigation. That makes it much easier to skim, revisit, and compare earlier parts of a chat without losing conversational context.

Google AI Studio implements this using custom scrollbars with clickable markers (for each user message), but displayed markers are only limited to a few messages at a time (updates dynamically on scroll)

Image

GitHub Copilot Chat also allows ArrowUp/Down keyboard navigation to jump across every user message in the conversation history (when the chat area is focused). 

Would be cool if we have something similar in PostHog Code - I'm proposing a jump to message modal showing an instant overview of all user messages + keyboard navigation (see https://github.com/PostHog/code/pull/2361)

charlesvien:
Recreates #2361, which GitHub refuses to reopen because the head branch was force-pushed after it was closed. Same branch and author (@Basit-Balogun10), now including the two commits that port the feature onto the new ChatThread (8ff1896, f733e63).

Note: still stacked on #2321, so the diff includes its 21 commits (cross-fork branches can't be PR bases). The actual change here is the last two commits.

Adds Alt/Option+Up/Down navigation between user messages, a Cmd/Ctrl+J jump-to-message picker with date filtering and the matching entries in the shortcuts sheet. See #2361 for the full description and demo videos.

Closes #2360.


charlesvien:
Superseded by #3407, same work squashed into a single commit and rebased onto main.

PR #3412, opened 2026-07-14
feat(sidebar): Add More tab, customization and click tracking

charlesvien:
## Problem

The sidebar nav keeps growing and every item competes for attention, with no way to tune it per user. We also have no data on which nav items people actually use. (so I added event tracking as well)

BTW- we can make certain items like the Inbox non-customizable/fixed so they are always there. I'm open to feedback here or throwing this out entirely if this is not _it_.

## Changes

- A collapsible More row now sits at the bottom of the nav. Search lives under it by default; Inbox, Agents, Skills, MCP servers, Command Center, Contexts and Activity start at the top level. While More is collapsed and one of its items is the active page, that item takes over the More row so the current page stays visible.
- A "Customize sidebar" entry inside More opens a dialog where all eight of those items can be promoted or demoted. Choices persist as per-item overrides in the sidebar store, so items we add later keep their intended default even for users with saved preferences, and corrupt persisted state degrades to defaults instead of crashing rehydration.
- Every nav row fires a typed "Sidebar nav item clicked" event with an `item` property (and `in_more` for clicks inside the expanded section), and customize toggles fire "Sidebar customized", so per-item usage is measurable.

| Default | More expanded |
| --- | --- |
| Default sidebar | More expanded |

| Customize dialog | Search promoted, Inbox demoted |
| --- | --- |
| Customize dialog | Customized sidebar |

## How did you test this?

- Drove the real desktop app (dev build) end to end: More expand and collapse, the customize dialog, promoting and demoting items and persistence across reload. Screenshots above are from that session.
- Unit tests for the per-item visibility overrides, defaults and corrupt-storage rehydration, plus component tests for the nav section (hidden items moving under More, active-item takeover, click tracking, contexts toggle) and the customize dialog. Typecheck and biome clean.

## Automatic notifications

- [ ] Publish to changelog?


adboio (approved):

charlesvien:
* **#3670** <a href="https://app.graphite.com/github/pr/PostHog/code/3670?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3412** <a href="https://app.graphite.com/github/pr/PostHog/code/3412?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/3412?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3418, opened 2026-07-14
feat(mobile): stop a running cloud run (port #3382)

Gilbert09:
## Problem

Mobile had no way to stop an in-flight cloud run — the only control was the turn-level cancel in the composer. Desktop shipped a "stop run" control in #3382; this ports the same user-facing capability to the React Native app.

**Why:** requested to bring the desktop stop-run feature to mobile so a user viewing a running cloud task can cancel the whole run, not just the current turn.

## Changes

- `cancelRun(taskId, runId, reason?)` in `features/tasks/api.ts` — POSTs to `.../tasks/{taskId}/runs/{runId}/cancel/` with the same auth/header pattern as the sibling run-scoped calls, surfacing the backend `error` message on failure.
- `stopRun` action in the task session store — optimistically marks the session `stopRequested` (and clears the pending turn) so the Stop control hides immediately, reverting on failure.
- A Stop button in the task header (`StopRunButton`), shown only for a non-terminal cloud run, with a confirmation dialog (`confirmStopRun`) mirroring the existing archive-running-task confirm.
- Fires a `Task run stopped` analytics event via the mobile analytics helper, mirroring the desktop event name/properties.
- Extracted a small `countUserMessages` selector so the screen doesn't reach into session-event internals.

## How did you test this?

- `pnpm --filter mobile test` — 409 passing, including new `cancelRun` success/error/HTTP-failure tests and `countUserMessages` tests.
- `pnpm --filter mobile lint` — clean.

---

### DIFF
--- a/apps/mobile/src/app/task/[id].tsx
+++ b/apps/mobile/src/app/task/[id].tsx
@@ -18,6 +18,7 @@ import { getTask, runTaskInCloud } from "@/features/tasks/api";
 import { FloatingTaskHeader } from "@/features/tasks/components/FloatingTaskHeader";
 import { PrDiffStatsBadge } from "@/features/tasks/components/PrDiffStatsBadge";
 import { PrStatusBadge } from "@/features/tasks/components/PrStatusBadge";
+import { StopRunButton } from "@/features/tasks/components/StopRunButton";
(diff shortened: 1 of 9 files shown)


danielcarletti (approved):
LGTM

PR #3421, opened 2026-07-14
feat(task-input): preselect server-side team/user default model

VojtechBartos:
## Problem

PostHog now supports team-level and per-user default AI run preferences for task runs, resolved server-side (PostHog/posthog#70671, UI in PostHog/posthog#70683). PostHog Code ignores them: the task input picker preselects the device-local `lastUsedModel`, so a project-wide default never reaches runs started from the desktop app, and preferences don't follow the user across devices.

## Changes

- `PostHogAPIClient.getTaskRunDefaults()` fetches `GET /api/projects/:id/tasks/my_config/` and returns the `resolved_ai_run_defaults` triple (user preference over project default). Returns null on any failure, missing endpoint (older servers), or when nothing is configured - defaults are a nicety and must never block task creation.
- `usePreviewConfig` fetches the defaults alongside the gateway options. Preselection order for the model picker is now: server default (when it targets the current adapter and the gateway still offers the model) > device-local last-used model > gateway default. The default's reasoning effort is applied with it when the model supports it. Changing the picker still affects only the current composition, so a one-off pick never overrides the server-side preference.

Not in this PR (follow-ups): passing the server default through the one-click inbox flows (`resolveDefaultModel` / `ReportModelResolver`), and omitting the triple from cloud-run creation so the backend resolves it at run time (the saga currently must send `runtime_adapter` for `initial_permission_mode`; with this PR the picker already carries the server default into the run explicitly, and the backend applies defaults for any caller that omits them).

## How did you test this?

- `vitest packages/api-client/src/posthog-client.test.ts` - 70 passed, including 3 new cases for `getTaskRunDefaults` (snake_case → client shape mapping, null on 404 from older servers, null when no default is configured).
- `turbo test --filter=@posthog/api-client --filter=@posthog/ui` and `turbo typecheck` for both packages - green.
- Not manually tested against a backend running the new endpoint yet.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/api-client/src/posthog-client.test.ts
+++ b/packages/api-client/src/posthog-client.test.ts
@@ -1695,4 +1695,71 @@ describe("PostHogAPIClient", () => {
       });
     });
   });
+
+  describe("getTaskRunDefaults", () => {
+    function makeClient(fetch: ReturnType<typeof vi.fn>) {
+      const client = new PostHogAPIClient(
+        "http://localhost:8000",
+        async () => "token",
+        async () => "token",
+        123,
+      );
+      (
+        client as unknown as {
+          api: { baseUrl: string; fetcher: { fetch: typeof fetch } };
+        }
+      ).api = { baseUrl: "http://localhost:8000", fetcher: { fetch } };
+      return client;
+    }
+
+    it("maps the resolved defaults into the client shape", async () => {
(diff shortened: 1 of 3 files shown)


charlesvien:

PR #3353, opened 2026-07-10
feat(web): cloud-tasks web host bring-up

gantoine:
## Summary

Brings up **`apps/web`** as a cloud-tasks-only browser host for PostHog Code — driving PostHog-cloud-hosted agent runs from a plain browser tab, with **no local git/terminal/workspace backend**. An authenticated user can reach the cloud task list, create a cloud task, stream the SSE chat, review diffs, attach files, run `/skill` commands, and use Channels/Canvas, Agents, and Inbox — entirely client-side.

This is also a portability smoke test for the layered architecture: `@posthog/core` and `@posthog/ui` run **unchanged** on the web host; everything web-specific lives in the `apps/web` composition root.

## Approach

Per the architecture rules (logic is portable; hosts are thin):

- **Shared packages untouched where possible.** The post-login shell (`@posthog/ui`) renders as-is. Web-specific behaviour is bound in `apps/web/src/web-container.ts` and served through an in-process tRPC router slice (`web-host-router.ts`) via `unstable_localLink` — the same `HostRouter` shape the desktop app serves over Electron IPC.
- **Real host-agnostic core services** back auth, cloud tasks, analytics, title generation, skills, and artifacts. Node/Electron-only capabilities are replaced with browser-native adapters (Web Notifications, `navigator.onLine`, `crypto.subtle`, localStorage, a `@pierre/diffs` worker, an fflate skill bundler).
- **New `HostCapabilities` primitive** (`@posthog/platform`) with `{ localWorkspaces: boolean }` (false on web) gates local-only UI (terminal, worktrees, folders, "continue locally", back/forward chrome, local settings) without forking components.
- **Procedures the cloud path never needs are thin stubs** in `web-host-router.ts`; anything unlisted `NOT_FOUND`s at call time, which is a deliberate signal of where the host is still thin.

## What works (verified in a logged-in browser)

- **Auth** — full PKCE OAuth in a popup; token exchange + session restore. The refresh token is **encrypted at rest** with AES-GCM under a non-extractable Web Crypto key held in IndexedDB, so a stolen localStorage dump can't be decrypted offline (`webAuthTokenCipher`).
- **Cloud tasks** — list, create (`environment: "cloud"`), SSE chat streaming, chat diffs / review page, log-history hydration.
- **Sidebar** — cloud tasks created in this browser persist (localStorage workspace store) and back-fill via `SYNC_CLOUD_TASKS_FLAG`.
- **Task title generation** via the portable LLM gateway.
[...]

PR #3447, opened 2026-07-14
feat(web): cloud-tasks browser host

gantoine:
## Summary

Brings up **`apps/web`** as a cloud-tasks-only browser host for PostHog Code — the same `@posthog/ui` shell and `@posthog/core` services as the desktop app, running in a plain browser tab over web platform adapters (no Electron, no local workspace-server). An authenticated user can reach the cloud task list, create a cloud task, stream the SSE chat, review diffs, attach files, run `/skill` commands, and use Channels/Canvas — entirely client-side.

Also a portability smoke test for the layered architecture: `@posthog/core` and `@posthog/ui` run **unchanged**; everything web-specific lives in the `apps/web` composition root.

## What's here

- **Auth** — browser PKCE OAuth (popup → `/callback` → BroadcastChannel relay), refresh token encrypted at rest with a non-extractable Web Crypto key in IndexedDB.
- **Cloud tasks** — real `AuthService` / `CloudTaskService` (host-agnostic core) over an in-process host-router slice via tRPC `unstable_localLink`; SSE chat, diffs, log history, sidebar.
- **Canvas / Channels** — the four canvas routers over the host-agnostic canvas services (PostHog `desktop_file_system` API).
- **Attachments** (drag/paste/picker), **`/skill` commands** (fflate bundler), **posthog-js** analytics/error-tracking/flags, **browser notifications**, **per-device stores** (localStorage).
- **Build** — vendor + route code-splitting (TanStack Router `autoCodeSplitting`).
- **Tests** — a hermetic Playwright happy-path e2e (boot → onboarding → sign-in → `/callback`), wired into CI alongside the desktop suite; `mprocs` procs for local dev.

## ⚠️ Depends on (merge first)

This is the additive web-host layer split out of #3353. It builds on three small enabling PRs and is **red until they land**, then a rebase turns it green:

1. #3444 — `refactor(auth): make the token cipher async`
2. #3446 — `feat(platform): host capabilities seam + boot-time binding guard`
3. #3445 — `refactor(shared): move the SKILL.md serializer into @posthog/shared`

Verified locally: with all three applied on top of current `main`, `@posthog/web` and `apps/code` typecheck clean and the web e2e passes. The standalone red is exclusively unresolved symbols from those three (async cipher, `@posthog/platform/host-capabilities` + `@posthog/di/hostCapabilities` + `REQUIRED_HOST_CAPABILITIES`, `serializeSkillMarkdown`).

## Notes

`apps/web` is a new app; nothing in `main` imports it, so its blast radius on the desktop host is nil. The shared-package/desktop changes that *do* carry risk were peeled into the three enabler PRs above.

---


apps/web/src/web-archive-store.ts
jonathanlab, line 1: Would it make more sense to use indexedDB here? These are quite complex shapes to store, but maybe it's overkill for our usecase.

jonathanlab (approved):
This looks really good: I noticed we're using a mix of local storage and indexed DB, I think it would make sense to choose 1 source of truth for handling browser storage instead of having this split across 2 mechanisms.

I think it would also be good if we updated our `AGENT.md` or relevant agent instructions with pointers on how to build for web.

PR #3451, opened 2026-07-14
fix(agent): let MCP publish channel CONTEXT.md and canvases

jakesciotto:
## Problem

A channel's **CONTEXT.md** and its **freeform canvases** are not files on disk — they live in PostHog, and the agent publishes them via the PostHog MCP sub-tools `desktop-file-system-instructions-partial-update` and `desktop-file-system-canvas-partial-update`.

The destructive-sub-tool permission gate classified every `*-partial-update` as destructive, so after the user already approved the generation plan, publishing hit a **second, redundant** "modify live PostHog data" approval (and behaved inconsistently in cloud runs). That gate exists to protect the user's PostHog **product data** (flags, experiments, notebooks) — not the app's own artifacts, which are the whole point of the generation task.

The hard part: the write must stay **bound to the artifact the task actually owns**. A blanket exemption by sub-tool name would let the agent publish to *any* id — silently overwriting an arbitrary channel's CONTEXT.md or any canvas. And the target id can't be threaded through structured session metadata without PostHog backend changes, because these tasks default to cloud runs whose session metadata is rebuilt server-side.

## Changes

- The generation prompts embed a machine-readable **authorized-write marker** (`sub-tool` + `id`) via a shared helper (`@posthog/shared`, rendered as an invisible HTML comment). This is the one trusted, in-repo source of the target id that works on both local and cloud runs.
- The agent captures those ids **only from user messages** (which it cannot forge) onto the session.
- The permission gate skips the destructive prompt for a sanctioned publish **only when the call's `id` matches an authorized id** — and never in plan mode, so CONTEXT.md co-design is preserved. Any other id still prompts. `isPostHogDestructiveSubTool` stays a pure classifier (these publishes are destructive; the id-scoped exemption lives in the handler).
- Fails safe: an unrecognised/malformed id, or a resumed session that never re-received the marker, falls back to the normal approval prompt.

## How did you test this?

- `posthog-exec-gate.test.ts` — id extraction from the exec command, the sanctioned-sub-tool predicate, and destructive classification (**18 passed**).
- `permission-handlers.test.ts` — matching id auto-allows without a prompt; a **different** id still prompts; plan mode never auto-allows (**30 passed**).
- `authorized-write-marker.test.ts` (shared) — build/parse round-trip, multiple markers, malformed markers (**4 passed**).
- `pnpm --filter @posthog/{shared,agent,ui} typecheck` and Biome on all changed files — clean.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/agent/src/adapters/claude/claude-agent.ts
+++ b/packages/agent/src/adapters/claude/claude-agent.ts
@@ -44,7 +44,7 @@ import {
   type SDKUserMessage,
   type SlashCommand,
 } from "@anthropic-ai/claude-agent-sdk";
-import { serializeError } from "@posthog/shared";
+import { parseAuthorizedWriteMarkers, serializeError } from "@posthog/shared";
 import { v7 as uuidv7 } from "uuid";
 import packageJson from "../../../package.json" with { type: "json" };
 import {
@@ -106,6 +106,7 @@ import {
   setMcpToolApprovalStates,
 } from "./mcp/tool-metadata";
 import { canUseTool } from "./permissions/permission-handlers";
+import { isSanctionedFirstPartyWriteSubTool } from "./permissions/posthog-exec-gate";
 import { getAvailableSlashCommands } from "./session/commands";
 import { parseMcpServers } from "./session/mcp-config";
 import {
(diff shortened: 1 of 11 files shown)


charlesvien:

PR #3478, opened 2026-07-15
feat(agent): export cloud run telemetry over OTLP

tatoalo:
## Problem

Cloud-task execution details were available only through per-run session logs, which made it difficult to query run health, timing, usage, and failures across tasks and users. The previous OTel path was not wired to a valid ingest endpoint, and terminal cloud-run hydration could also lose persisted state or redirect on transient task-detail failures.

## Changes

- Add opt-in OTLP logs and traces for cloud agent runs under `service.name=posthog-code-agent`.
- Emit a fail-closed allowlist of lifecycle metadata. Prompts, agent text, tool arguments and output, and raw error messages are excluded.
- Build one trace per run with a `task_run` root, per-turn spans, and per-tool-kind spans. Logs carry the active trace and span IDs for correlation.
- Export completed spans incrementally and close the root at an in-process terminal point, including background completion and fatal failures.
- Isolate exporter failures and cap shutdown time so telemetry cannot break or delay the agent run.
- Harden terminal cloud-run state and hydration so persisted permissions and conversation history survive reload and real resume paths.
- Document the telemetry configuration and behavior in the agent package.
- Companion backend configuration: https://github.com/PostHog/posthog/pull/71100


REPORT.md
jonathanlab, line 1: this probably should go before merging

jonathanlab (approved):
-

REPORT.md
tatoalo: Removed in 18df0f64d (and its posthog-side counterpart in PostHog/posthog@63d88b5b6).

PR #3489, opened 2026-07-16
fix(cloud-task): Load run history from presigned log URLs

charlesvien:
## Problem

Opening a cloud task with a large history can fail permanently with "Failed to load cloud run history". The bootstrap pages through the `session_logs/` API, where every page makes the server re-read and re-parse the run's entire log chain from S3. On big runs a single page exceeds the 30s default `authenticatedFetch` timeout, and since the bootstrap is all-or-nothing and Retry restarts at offset 0, the thread becomes unopenable while the run itself is still healthy.

## Changes

- Prefer downloading history straight from the presigned resume-chain `log_urls` that the run detail API now returns (PostHog/posthog#71394), parsing the JSONL client-side. This removes Django from the history byte path entirely and works for any history size. Falls back to the paginated API on old servers or when a download fails.
- Paginated fallback hardening: each page gets an explicit 120s timeout instead of inheriting the 30s auth default, transient page failures retry with backoff, and fetched pages survive watcher retries so Retry resumes from the last good offset instead of restarting at 0.
- A bootstrap generation counter discards superseded bootstrap work, since history fetches can now outlive a user-triggered retry.
- Mobile mirrors the same strategy: chain URLs first, then the paginated API, then the legacy single `log_url` fallback, with the page timeout raised from 10s to 120s.

## How did you test this?

Added Vitest cases in `cloud-task.test.ts`: bootstrap from chain URLs (including a final line without trailing newline) without touching `session_logs`, fallback to the paginated API when a chain download fails, and resume from the last fetched page after a watcher retry. Ran the full `@posthog/core` suite (2283 tests) and the mobile suite (429 tests), plus `pnpm --filter @posthog/core typecheck` and Biome on the changed packages. I did not manually test against a live run; the server-side PR needs to be deployed for the chain path to activate.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/apps/mobile/src/features/tasks/api.ts
+++ b/apps/mobile/src/features/tasks/api.ts
@@ -719,9 +719,11 @@ export async function fetchSessionLogs(
         offset: String(options.offset ?? 0),
       });
 
+      // Big runs can take the server a long time per page (it re-reads the whole
+      // log chain each request), so this needs far more than the default budget.
       const response = await authedFetch(
         `${baseUrl}/api/projects/${projectId}/tasks/${taskId}/runs/${runId}/session_logs/?${params}`,
-        { signal: createTimeoutSignal(10_000) },
+        { signal: createTimeoutSignal(120_000) },
       );
 
       if (!response.ok) {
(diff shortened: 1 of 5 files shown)


charlesvien:
@greptileai

PR #3534, opened 2026-07-17
feat(channels): stream agent turns into threads, read Activity per mention

adamleithp:
Agent turns and permission asks now land in a channel thread as messages you can read in place, and Activity treats a mention as read when you open its thread — not when you glance at the page.

> **Rebased 2026-07-20:** #3380 merged after this branch was cut and already ships the agent-row rendering (`author_kind`/`event`/`payload` on `TaskThreadMessage`, robot avatar + "Agent" label). Those pieces were dropped from this PR in ef4444b; what remains below is net-new.

## Thread messages

- **`turn_complete` / `permission_request` are typed.** Added a `TaskThreadMessageEvent` union and pointed `TaskThreadMessage.event` at it, so agent rows render from a stable key rather than a freeform string.
- **No duplicate turn.** The backend posts every agent turn as a durable `turn_complete` row keyed by `payload.run_id`. `visibleThreadMessages` drops that copy for a run the viewer is already streaming — the streamed turn is the one they watched arrive, and swapping it for the server's copy at the end made the message jump. A row with no run id is kept: a possible duplicate beats dropping the only copy.
- **Turns break where the work happened.** Text arrives as mid-sentence chunks, so consecutive chunks join raw — but a tool call between two pieces of prose is a paragraph break. Without it, "…before committing." ran straight into "The patch rebased cleanly." Trailing calls add no dangling break, and a run of calls adds only one.
- **Long bodies get a framed window** (`ThreadScrollBody`, built on quill's `ChatStream`): capped height, scrolled for the rest, so one essay doesn't push the next row off screen. It follows live output while streaming, and opens on the last line once finished. The border lives on the wrapper, not the stream — `ChatStream`'s mask would otherwise eat a border on its own element exactly when there's more to read.

## Activity

- **A mention opens its thread beside the list**, instead of navigating away to the full task. Mentions whose channel folder doesn't resolve still fall through to the task view.
- **Read is per mention, earned by opening its thread.** Opening the Activity page marked everything seen, which is backwards: the page is a list of things you haven't dealt with, not a receipt for them. `isMentionUnread`/`countUnreadMentions` replace the global watermark; `lastSeenAt` stays as a read-only historical watermark so the switch doesn't resurface old mentions as unread.
- **The open mention's row is highlighted** (`bg-fill-selected`). Keyed on the mention, not the task — the panel tracks a task, and a task can be mentioned in several rows, so keying on it alone lit all of them for one click.
- **Read ids are bounded** (500, oldest out first) and survive persistence — `Set` doesn't round-trip through JSON, so it's stored as an array and rebuilt on load.

## Notifications

- **A task whose thread panel is open counts as "being viewed".** The active target was route-only, so a task filling half the screen in a thread panel still got a "needs your input" toast. `activeNotificationTarget` now considers the panel — a collapsed panel is a rail with nothing legible in it, so it doesn't count. The host's `getActiveTarget` is thin glue over it.

## Notes

- `@posthog/quill` `0.3.0-beta.24` → `0.3.0-beta.25` for `ChatStream`.
- Known gap: `permission_request` rows render as text; their Approve/Deny actions aren't wired yet.
- Activity can't yet grey out superseded permission asks — `TaskMentionDTO` carries no `event`/`payload`/`author_kind` (#3380 only added them to `TaskThreadMessage`), which is also why an agent's mention still shows as "Unknown" in the list. Needs a `../posthog` change; not in this PR.
- Pre-existing failure on `main`, unrelated and untouched here: `@posthog/git` `detectDefaultBranch > prefers remote HEAD over local detection`.

## Testing

- `pnpm typecheck`, `pnpm lint`, `node scripts/check-host-boundaries.mjs` (no new violations) all clean.
- `pnpm test`: green except the pre-existing `@posthog/git` failure above.
- New unit tests: `visibleThreadMessages` (threadTimeline), `isMentionUnread`/`countUnreadMentions` (mentionActivity), tool-call turn breaks (threadAgentTurns), `activeNotificationTarget`.
- Driven live over CDP against the real app: agent turns streaming into a thread, the scroll window opening on its last line, and the Activity row highlighting only the mention whose thread is open.
- Re-ran typecheck, lint, boundaries, and core/ui tests after the rebase + trim.

### DIFF
--- a/apps/code/src/renderer/desktop-services.ts
+++ b/apps/code/src/renderer/desktop-services.ts
@@ -70,6 +70,7 @@ import {
   type IAuthSideEffects,
 } from "@posthog/ui/features/auth/identifiers";
 import { authKeys } from "@posthog/ui/features/auth/useCurrentUser";
+import { useThreadPanelStore } from "@posthog/ui/features/canvas/stores/threadPanelStore";
 import {
   FEATURE_FLAGS,
   type FeatureFlags,
@@ -85,6 +86,7 @@ import {
 } from "@posthog/ui/features/integrations/integrationsClientImpl";
 import { NAVIGATION_TASK_BINDER } from "@posthog/ui/features/navigation/taskBinder";
 import { navigationTaskBinder } from "@posthog/ui/features/navigation/taskBinderImpl";
+import { activeNotificationTarget } from "@posthog/ui/features/notifications/activeTarget";
 import {
   ACTIVE_VIEW_PROVIDER,
   type IActiveView,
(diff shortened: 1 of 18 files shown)


adboio:
Rebased onto `main` and trimmed the parts #3380 (merged after this branch was cut) already ships:

- **Dropped**: the `TaskThreadMessage` field additions (`author_kind`/`event`/`payload` — now on main) and the separate `AgentThreadRow` + `isAgentThreadMessage` rendering path (main's `ThreadMessageRow` already handles agent rows identically). See ef4444b.
- **Kept**: the `TaskThreadMessageEvent` union, `visibleThreadMessages` streaming dedup, tool-call paragraph breaks, `ThreadScrollBody` + quill beta.25, and the entire Activity/notifications half — all still net-new vs main.

Typecheck, biome, host boundaries, and core/ui tests all green after the trim.

PR #3552, opened 2026-07-17
fix(auth): load projects for project-scoped OAuth

dmarticus:
## Problem

Project-scoped OAuth tokens expose authorized project IDs in `scoped_teams`, but PostHog Code discarded that field and tried to load projects through the organization endpoint. When the backend rejects project-scoped tokens on organization endpoints, the session remains authenticated with “No project selected.”

This follows [#3279](https://github.com/PostHog/code/pull/3279), merged on July 8, 2026. That fix correctly handled empty `scoped_organizations` by falling back to the user’s current organization, but it depended on the organization endpoint accepting the team-scoped token. The direct scoped-project path removes that dependency.

**Why:** Local development should work when a developer grants project-level access instead of organization-level access during OAuth authorization.

## Changes

- Preserve `scoped_teams` from OAuth token responses and hydrate those projects directly through project-scoped endpoints.
- Keep the existing organization-scoped and legacy fallback paths, including transient recovery behavior.
- Cover both runtime token parsing and the flow where organization access is forbidden.

## How did you test this?

- `pnpm build:deps`
- `pnpm --filter @posthog/core exec vitest run src/auth/auth.test.ts src/auth/oauth.schemas.test.ts`
- `pnpm exec biome lint packages/core`

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/auth/auth.test.ts
+++ b/packages/core/src/auth/auth.test.ts
@@ -81,6 +81,7 @@ function mockTokenResponse(
     accessToken?: string;
     refreshToken?: string;
     scopedOrgs?: string[];
+    scopedTeams?: number[];
   } = {},
 ) {
   return {
@@ -91,6 +92,7 @@ function mockTokenResponse(
       expires_in: 3600,
       token_type: "Bearer",
       scope: "",
+      scoped_teams: overrides.scopedTeams,
       scoped_organizations: overrides.scopedOrgs ?? ["org-1"],
     },
   };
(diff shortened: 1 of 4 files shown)


dmarticus:
CI follow-up: all code-owned checks on this PR passed (build, unit/integration/E2E, typecheck, quality, and security scans). The remaining `Trunk Impacted Targets / upload` failure is repository-wide: `TRUNK_API_TOKEN` is empty and the Trunk API returns HTTP 401. Recent unrelated PRs are failing the same workflow. I updated this branch from `main`, and the fresh run reproduced the same infrastructure failure; no workflow or secret changes are included in this PR.

PR #3585, opened 2026-07-20
feat(mcp-servers): resolve store icons from icon_domain via the logo.dev proxy

cvolzer3:
## Problem

I changed the MCP store to use a catalog-as-code approach to adding MCP servers. As part of that changed, I moved us away from bundled assets to logo.dev icons. The existing MCP store componetry uses the old model structure, so it needs to be updated.

## Changes

- Resolve brand domain from server URLs for custom installations without templates.
- Render icons through a proxy with the active theme, caching loads and retrying on theme flips
- Migrate mobile implementation to align with these changes, removing bundled assets

## How did you test this?

Tested the desktop version locally. Not sure how to test mobile, so I did not check that.


tatoalo (approved):

cvolzer3:
@tatoalo fastest stamp this side of the mississippi

PR #3596, opened 2026-07-20
fix(mobile): prewarm the selected sandbox image (port #3511)

Gilbert09:
## Problem

Mobile prewarms cloud sandboxes, but the warm request dropped the sandbox
environment / custom base image, so a warmed sandbox could be built on a
different image than the task would actually run on — and then get reused
anyway. This ports desktop PR #3511 to mobile.

**Why:** keep the warm path correct so a reused sandbox always matches the
selected image, matching desktop behaviour.

## Changes

- `warmTask()` now optionally sends `sandbox_environment_id` / `custom_image_id`
  in the warm request body (only when set).
- `useWarmTask` folds both ids into the debounce/dedupe key, so changing the
  selection triggers a re-warm, and forwards them to the warm call.
- `runTaskInCloud` carries the ids on the run so a reused warm sandbox matches
  the selection instead of a mismatched default.

Mobile has no sandbox-environment / custom-image picker today (the ids are only
surfaced read-only via the custom-image badge), so these values are threaded as
optional seams — inert until a selection exists, no new picker UI added.

## How did you test this?

- Added unit tests (`api.warm.test.ts`, `api.test.ts`, `useWarmTask.test.tsx`)
  covering: ids forwarded when set, omitted when unset, and the warm key
  re-warming when the image selection changes.
- Ran the mobile tasks-feature test suite (`vitest run src/features/tasks/`,
  196 passing), `tsc --noEmit`, and `biome check` on the touched files.

## Automatic notifications

- [ ] Publish to changelog?

---


MarconLP (approved):

Gilbert09:
/trunk merge

<sub>🦉 via [talyn.dev](https://talyn.dev)</sub>

PR #3611, opened 2026-07-21
fix(agent): treat an emptied sandbox env file as GitHub logout

VojtechBartos:
On a GitHub actor transition (multiplayer follow-ups on the PostHog side), the backend logs the sandbox out by emptying the token vars in `/tmp/agent-github-env`. This agent-server was resurrecting the previous actor's token anyway.

```mermaid
flowchart TD
    R([resolveGithubToken]) --> F{ /tmp/agent-github-env present? }
    F -- "no (local / desktop)" --> P[use process.env]
    F -- yes --> T{ token var non-empty? }
    T -- yes --> U[use the file token]
    T -- "no (emptied = logout)" --> E["return '' (no token)"]
```

## Changes

- **`resolveGithubToken`**: a present-but-empty file is an explicit logout (return `""`), not a fallback to `process.env`. Only an absent/unmanaged file defers to the process env. (`"" ?? processEnv === ""`, so an emptied file no longer resurrects the frozen launch-time token.)
- **`gh` attribution/whoami** (`fetchPrAttribution`, `fetchGhLogin`): run as the current actor via the live file token instead of inheriting `process.env`.

## Companion PR (required together)

PostHog/posthog#71941 (backend): removes the frozen process-env token and adds the rebind/logout gate.

`github-token.test.ts` (13 pass), including the assertion that an emptied file does not resurrect the process-env token. Validated live in a running sandbox (old resolver returns the previous actor's token; new returns `""`), and end-to-end on a dedicated GitHub org where both users have access to the same repo: no-access follow-up blocked, with-access follow-up rebinds.

Tested locally together with posthog/posthog changes.

## Agent context

Model: Claude Opus 4.8.

### DIFF
--- a/packages/agent/src/utils/github-token.ts
+++ b/packages/agent/src/utils/github-token.ts
@@ -18,10 +18,15 @@ export function readGithubTokenFromSandboxEnvFile(
   let raw: string;
   try {
     raw = readFileSync(envFilePath, "utf8");
-  } catch {
-    // No env file (local/desktop or test) — signal "unmanaged" so the caller
-    // falls back to the process env.
-    return undefined;
+  } catch (err) {
+    // A genuinely absent file (local/desktop or test) is unmanaged: signal that so
+    // the caller falls back to the process env. But an existing-yet-unreadable file
+    // during an actor transition must NOT resurrect the frozen process token, so
+    // treat any other read error as an explicit logout (fail closed).
+    if ((err as NodeJS.ErrnoException).code === "ENOENT") {
+      return undefined;
+    }
+    return "";
   }
   const env: Record<string, string> = {};
   for (const entry of raw.split("\0")) {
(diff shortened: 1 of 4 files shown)


packages/agent/src/utils/github-token.ts
```diff
@@ -12,19 +15,38 @@ const SANDBOX_GITHUB_ENV_FILE = "/tmp/agent-github-env";
...
+  // The file is the backend's live credential channel. If it carries the token
+  // vars but they are emptied, that is an explicit logout on an actor
+  // transition — return "" so the caller does NOT resurrect the previous
+  // actor's token from the frozen launch-time process env. Only a file with no
+  // token vars at all is "unmanaged" and defers to the process env.
+  if (GITHUB_TOKEN_ENV_VARS.some((name) => name in env)) {
+    return "";
   }
   return undefined;
 }
```
tatoalo, line 59: The BE clears this file to zero bytes, which returns `undefined` here and lets consumers fall back to the previous actor’s cached token. Can we tsreat any successfully read empty credential file as logout and add a zero-byte regression test just to make this a bit cleaner perhaps?

tatoalo (approved):

PR #3612, opened 2026-07-21
Group task configuration under settings

puemos:
## Problem

The primary navigation is becoming crowded with setup and configuration destinations that are used much less often than core workflows.

The navigation should prioritize daily actions—creating tasks, searching, reviewing the inbox, and using the command center. Configuration that is typically completed once or revisited occasionally should live in Settings, where it remains discoverable without competing with frequent work.

**Why:** Separate doing work from configuring how work gets done. This keeps the main navigation focused while giving task and agent configuration a clear home.

## Changes

- Replace the separate Agents, Skills, and MCP servers navigation entries with one **Configure** entry that opens Settings with **Agents** selected.
- Group Settings by concern: **Account**, **Workspace**, **Configure**, **Experience**, **Integrations**, and **Application**.
- Place Agents, Skills, MCP servers, Claude Code, and Self-driving together under **Configure**.
- Make Settings the canonical home for Skills and MCP servers while preserving legacy URLs through redirects.
- Preserve existing configuration behavior and full-height pages, and consistently fall back from categories unavailable on the current host.

The information architecture follows one rule:

- **Navigation:** destinations and actions used regularly
- **Settings:** setup, customization, integrations, and lower-frequency configuration

## Screenshots

Grouped Settings navigation
Configure settings

## How did you test this?

- Focused settings visibility tests
- Biome checks on touched files
- `git diff --check`

## Automatic notifications

- [ ] Publish to changelog?

---


k11kirky (approved):

packages/ui/src/features/sidebar/components/items/ConfigureItem.tsx
```diff
@@ -0,0 +1,17 @@
+import { SlidersHorizontal } from "@phosphor-icons/react";
+import { SidebarItem } from "../SidebarItem";
+
+interface ConfigureItemProps {
+  onClick: () => void;
+}
+
+export function ConfigureItem({ onClick }: ConfigureItemProps) {
```
k11kirky, line 8: `ConfigureItem` never receives an `isActive` prop, so this sidebar entry never highlights while Settings is open (settings renders as a route inside the shell with the sidebar still visible). The three deleted items (`AgentsItem`/`SkillsItem`/`McpServersItem`) all derived `isActive` from `view.type`; `useIsSettingsOpen()` in `useOpenSettings.ts` already exists and could drive this here.

packages/ui/src/features/sidebar/components/SidebarNavSection.tsx
```diff
@@ -158,15 +145,7 @@ export function SidebarNavSection({
...
-        <AgentsItem isActive={isAgentsActive} onClick={navigateToAgents} />
-      </Box>
-
-      <Box>
-        <SkillsItem isActive={isSkillsActive} onClick={goSkills} />
-      </Box>
-
-      <Box>
-        <McpServersItem isActive={isMcpServersActive} onClick={goMcpServers} />
+        <ConfigureItem onClick={() => openSettings("agents")} />
```
k11kirky, line 148: This `onClick` has no `inChannels` branching, so clicking Configure while in the Channels (`/website`) space navigates out to the top-level `/settings/$category` route — unlike Home/Command Center (which branch via `goHome`/`goCommandCenter`) and unlike the removed `navigateToWebsiteSkills`/`navigateToWebsiteMcpServers`, which kept Skills/MCP servers in-space. If leaving the space is intentional, worth calling that out in the comment above the way the Inbox/New task jump-out is.

packages/ui/src/features/sidebar/components/SidebarNavSection.tsx
```diff
@@ -28,14 +24,12 @@ import { useCommandMenuStore } from "@posthog/ui/shell/commandMenuStore";
 import { Box, Flex } from "@radix-ui/themes";
 import { useRouterState } from "@tanstack/react-router";
 import { ActivityItem } from "./items/ActivityItem";
-import { AgentsItem } from "./items/AgentsItem";
 import { CommandCenterItem } from "./items/CommandCenterItem";
+import { ConfigureItem } from "./items/ConfigureItem";
```
k11kirky, line 28: After this PR, `navigateToAgents`, `navigateToSkills`, `navigateToMcpServers`, `navigateToWebsiteSkills`, and `navigateToWebsiteMcpServers` in `navigationBridge.ts` have zero remaining callers repo-wide. Worth deleting them, and deciding whether the still-wired standalone `/agents`, `/skills`, `/mcp-servers` (and `/website` mirror) routes should redirect into Settings now that they're a second live UI for the same views.

packages/ui/src/features/settings/components/SettingsPanel.tsx
```diff
@@ -193,6 +210,10 @@ export function SettingsPanel({
     !localWorkspaces && LOCAL_ONLY_CATEGORIES.has(activeCategory)
```
k11kirky: This guard only checks `LOCAL_ONLY_CATEGORIES`, not the `hiddenCategories` computed just above (which also hides `plan-usage` when billing and spend analysis are disabled). A deep link or `openSettings("plan-usage")` on such a host still renders `PlanUsageSettings` even though the sidebar hides it. Pre-existing gap, but `hiddenCategories.has(activeCategory)` is a behavior-equivalent superset that would close it now that the set already exists.

k11kirky (commented):
Automated review — findings inline.

packages/ui/src/features/settings/components/SettingsPageContent.tsx
```diff
@@ -0,0 +1,181 @@
...
+import { TerminalSettings } from "@posthog/ui/features/settings/sections/TerminalSettings";
+import { UpdatesSettings } from "@posthog/ui/features/settings/sections/UpdatesSettings";
+import { WorkspacesSettings } from "@posthog/ui/features/settings/sections/WorkspacesSettings";
+import { WorktreesSettings } from "@posthog/ui/features/settings/sections/worktrees/WorktreesSettings";
+import type { SettingsCategory } from "@posthog/ui/features/settings/types";
+import { SkillsView } from "@posthog/ui/features/skills/SkillsView";
+import { Box, Flex, ScrollArea, Text } from "@radix-ui/themes";
+import type { ComponentType, ReactNode } from "react";
+
+const SETTINGS_PAGE_LAYOUT = {
```
k11kirky, line 24: This binary layout choice (contained vs full-bleed) goes through a const object + mapped type + `defineSettingsPage` factory default + a layout→component `Record` — four pieces of indirection, none referenced outside this file, for 2 of 20 pages that use `FULL_BLEED`. A `fullBleed?: boolean` field on `SettingsPageDefinition` plus an inline ternary would read the same with less machinery.

packages/ui/src/features/settings/components/SettingsPanel.tsx
```diff
@@ -246,21 +263,25 @@ export function SettingsPanel({
...
-                  key={item.id}
-                  item={item}
-                  isActive={isActive}
-                  onClick={() => setCategory(item.id)}
-                />
-              );
-            })}
+          <div className="flex flex-col gap-3 py-2">
+            {sidebarGroups.map((group) => (
+              <div key={group.label}>
```
k11kirky, line 234: This hand-rolls the uppercase/tracked group-label styling instead of reusing quill's `MenuLabel`, which `TaskListView.tsx` already uses for the same grouped-nav-label purpose (`.quill-menu-label`: 10px, uppercase, tracked, weight 600 — a close match, overridable via `className` as `TaskListView` does). Consolidating avoids the two label styles drifting apart on a future restyle.

packages/ui/src/features/sidebar/components/items/ConfigureItem.tsx
puemos: Settings is a full-page route that unmounts the standard app sidebar, so Configure is not visible while Settings is open and has no active state to display. No change needed.

packages/ui/src/features/settings/components/SettingsPageContent.tsx
puemos: Keeping the typed layout registry intentionally. Page title, component, and layout metadata live together, and the typed layout constant makes supported variants exhaustive. This avoids returning to parallel metadata structures or putting layout branching back in the render path.

packages/ui/src/features/sidebar/components/SidebarNavSection.tsx
puemos: This is intentional. Configure is the single entry into the top-level Settings experience from either Code or Channels, consistent with Settings being the canonical home for configuration.

packages/ui/src/features/settings/components/SettingsPanel.tsx
puemos: Addressed in commit 7438f5ab: category resolution now uses the complete hiddenCategories set, so Plan & Usage and host-specific hidden categories fall back to General consistently. Added focused visibility coverage.

packages/ui/src/features/settings/components/SettingsPageContent.tsx
```diff
@@ -0,0 +1,181 @@
...
+  notifications: defineSettingsPage("Notifications", NotificationsSettings),
+  "plan-usage": defineSettingsPage("Plan & usage", PlanUsageSettings),
+  workspaces: defineSettingsPage("Workspaces", WorkspacesSettings),
+  worktrees: defineSettingsPage("Worktrees", WorktreesSettings),
+  environments: defineSettingsPage("Environments", EnvironmentsSettings),
+  "cloud-environments": defineSettingsPage(
+    "Environments",
+    EnvironmentsSettings,
+  ),
+  agents: defineSettingsPage("Agents", AgentsSettings),
```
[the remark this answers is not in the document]
puemos: Confirmed: the pre-router approval screen can open the shared Settings panel, and this PR newly exposes the Agents setup action there. This commit does not address that approval-state behavior, so I am leaving the thread open rather than marking a valid concern resolved.

packages/ui/src/features/settings/components/SettingsPanel.tsx
puemos: Addressed in commit 7438f5ab: Settings group headings now reuse Quill MenuLabel with only local spacing and color overrides.

packages/ui/src/features/sidebar/components/SidebarNavSection.tsx
puemos: Addressed in commit 7438f5ab: removed the five unused navigation helpers, redirected legacy Skills/MCP routes—including Channels mirrors—to Settings, and updated remaining internal links to target Settings directly. The /code/agents route remains because it is the separate agent-applications experience.

PR #3650, opened 2026-07-21
feat(setup): require hosting-provider env var checklist in setup runs

MattBro:
## Problem

Cloud wizard setup runs generate a PR that defines env vars in `.env.example`, but nothing tells the user those vars also need to exist in their hosting provider (Vercel, Netlify, etc.). Users who merge the PR and deploy get a production site that silently sends no events — a recurring theme in wizard feedback.

## Changes

- Adds `DEPLOYMENT_ENV_VAR_PROMPT` to `packages/core/src/setup/prompts.ts`: a checklist instruction requiring the agent to detect the hosting provider and list, in the PR body (or final message for local runs), exactly which env vars the user still needs to set there.
- Appends it to both live setup suggestion prompts in `suggestions.ts` (fresh install and SDK-already-present variants), with phrasing that works for both PR-based cloud runs and local runs.
- Removes `WIZARD_PROMPT`, which has been dead code since its last consumer was deleted (`eae3b11b4`, May 2026).

## How did you test this?

- `packages/core`: 2,488 tests pass (`vitest run`), including new tests asserting both suggestion variants include the deployment checklist.
- `pnpm run lint` — no errors.
- Note: `tsc --noEmit` in `packages/core` currently fails on two `pi-runtime` files with a module-resolution error against `@posthog/agent/pi/types`; these files are untouched by this PR and the error exists independently of it.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/core/src/setup/prompts.ts
+++ b/packages/core/src/setup/prompts.ts
@@ -1,6 +1,6 @@
 import { BASE_CATEGORY_ENUM } from "@posthog/core/setup/types";
 
-// Appended to every prompt that instruments an integration. Writing env vars to
+// Appended to every prompt that performs the initial PostHog SDK install/init. Writing env vars to
 // .env files does nothing for the deployed app, and no run can read the hosting
 // provider's env, so the checklist is unconditional. Where it gets delivered is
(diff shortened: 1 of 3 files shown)


charlesvien:

PR #3651, opened 2026-07-21
refactor(web): route browser persistence through one validated seam

gantoine:
## Why

The web host uses localStorage as its single persistence layer, but the `getItem → JSON.parse → guard → JSON.stringify → setItem` boilerplate was hand-rolled across five per-device stores. This consolidates all browser-storage access into one seam and adds Zod-validated versioning so future shape changes don't require hand-written migrations.

## What

**One seam — `apps/web/src/web-local-store.ts`.** All localStorage access routes through it (grep confirms zero direct `window.localStorage` elsewhere):
- `createRecordStore(key, entrySchema)` — the `Record<string, Entry>` registries (workspaces, archive, task-metadata). Validates **per entry** on load and drops only stale rows.
- `readValidated(key, schema, fallback)` — the browser-tabs snapshot (single object; drops to empty + re-seeds if incompatible).
- Raw `readJson` / `writeJson` / `removeKey` — auth session/preferences (record types owned by core, no local schema).
- `rawLocalStorage` — the zustand persist backend.

**Versioning via Zod, not migrations.** Every persisted store is a discardable per-device cache validated on read; invalid data is shed and rebuilt from the server. Evolving a shape is now a schema edit. Reuses the canonical `workspaceSchema` / `tabsSnapshotSchema` from `@posthog/shared` (no drift from the desktop service), and derives the two web-local types (`WebArchivedTask`, `TaskMetadata`) from their schemas via `z.infer`.

**IndexedDB left untouched.** It holds only the non-extractable AES-GCM auth cipher key (`web-auth-adapters.ts`), which localStorage physically cannot store without exposing its raw bytes — the property that keeps a stolen token dump undecryptable offline. This is documented as a deliberate exemption, not a second app-state store.

**Docs.** Adds a "Web Host" section to `AGENTS.md` (build story + storage policy) and fixes a stale `web-container.ts` comment that referenced a nonexistent test.

## Behavior / risk

No behavioral change for valid data. On first load after this lands, any persisted cache that fails validation is dropped and rebuilt (these are per-device sidebar/tab caches, not durable data). No async ripple into host-agnostic `core`/`ui` — reads stay synchronous.

## Testing

- `pnpm --filter @posthog/web typecheck` ✅
- `biome check` ✅ (also enforced by pre-commit)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -126,6 +126,19 @@ node scripts/check-host-boundaries.mjs --prune
 
 Do not use `--init` to baseline new violations.
 
+## Web Host
+
+`apps/web` is a real host and the portability smoke test: if a `@posthog/core`/`@posthog/ui` change compiles and boots on desktop but not here, it leaked a host dependency. It is **cloud-only** — no local filesystem, git, pty, or worktrees — so it binds `HOST_CAPABILITIES` to `{ localWorkspaces: false }` and stubs local-only host clients to reject at call time.
+
+Building a feature for web means: the portable core/UI already runs unchanged; you only supply web adapters and bind them.
+
+- **Composition root** is `apps/web/src/web-container.ts` (`WebBindings` + `TypedContainer<WebBindings>`). It loads the same core/UI feature modules `apps/code`'s renderer does, then binds web adapters (the `web-*.ts` files) for each platform/host capability. Unlike desktop, module-loading and adapter-binding live in this one file, not split into `desktop-contributions.ts` / `desktop-services.ts`.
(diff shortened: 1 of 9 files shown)


gantoine:
Addressed the review feedback in c9428ab90. All three comments were the same root issue — the seam's best-effort *swallow* was applied where the original direct writes let errors **propagate**, turning failures into silent successes. Fixed by splitting the seam into two tiers:

- **Auth session/preferences + logout** (veria-ai, greptile P1 security / P2): `saveCurrent`, the preferences `write`, and `clearCurrent` now use strict variants (`writeJsonStrict` / `removeKeyStrict`) that propagate. A swallowed `clearCurrent()` would have reported logout complete while the session stayed in localStorage, recoverable on reload — that's the regression, now gone.
- **Zustand backend** (greptile P1): `rawLocalStorage.setItem`/`removeItem` let storage errors throw again (matching the original `web-storage.ts`), so the renderer persistence layer's await/log path sees real failures instead of dropped drafts/settings/layout writes looking successful.

The best-effort `writeJson`/`removeKey` stay only for the rebuildable per-device caches (workspaces, archive, task-metadata, browser-tabs), where a dropped write costs cross-reload persistence of a cache that re-derives — not correctness. The two tiers are documented at the top of `web-local-store.ts`.

jonathanlab (approved):

PR #3659, opened 2026-07-21
feat(agent): route selected products to the ai-gateway

brandonleung:
## Problem

Model calls from code sandboxes all go through the legacy Python `llm-gateway`. We're migrating products to the new Go `ai-gateway` one at a time, so the agent-server needs to pick a gateway per request. The two gateways also take attribution differently: the old one wants a product slug in the URL plus per-property headers, the new one wants a single JSON header.

## Changes

Routing to the new gateway is **opt-in via two env vars**:

- **`AI_GATEWAY_URL`** is the new gateway's address; **`AI_GATEWAY_PRODUCTS`** lists the `ai_product` values that may use it. Both must be set. Unlisted products keep using the Python gateway.
- Requests on the new path carry `ai_product`, `ai_stage`, and `team_id` in one `X-PostHog-Properties` header, built by a new `buildPosthogPropertiesBlob` helper in `@posthog/shared` (enforces the gateway's 8KB / ASCII limits).
- Signals sandbox stages get **per-stage `ai_product` tags** (`signals_scout`, ...), matching the non-sandbox signals stages.

## How did you test this?

Unit tests cover the routing choice and the header builder (cap trimming, sanitization, unlisted products falling back). `tsc` and `biome` clean.

**Inert until** the sandbox chart sets both env vars (PostHog/posthog#72770 plus a charts PR). Rollback: clear either one.


packages/agent/src/utils/gateway.ts
```diff
@@ -79,16 +81,123 @@ export function resolveLlmGatewayUrl(
...
   product: GatewayProduct = "posthog_code",
 ): string {
   return `${getGatewayBaseUrl(posthogHost)}/v1/usage/${product}`;
 }
+
+export interface GatewayTarget {
+  /** Base URL: slugless for the Go gateway, product-slugged for the Python one. */
+  baseUrl: string;
+  /** True when the Go gateway is selected, which changes the header format. */
+  slugless: boolean;
```
richardsolomou: nit but could we name this `isAiGateway` instead? `slugless` also controls the header protocol, so it hides the actual distinction here.

richardsolomou (approved):
would be worth testing exact 8,192-byte and 8,193-byte blobs so an off-by-one at the gateway limit cant slip through

PR #3484, opened 2026-07-15
fix(sidebar): keep the peeked sidebar open while the ProjectSwitcher menu is open

aleexzxzxzx:
## Problem

Closes #3483

The collapsed-sidebar hover-peek (#3367) closes the peek on any directional exit,
including while the ProjectSwitcher menu is open. That menu renders in a portal
anchored to its trigger inside the sidebar, so when the peek collapses the menu is
left floating over the content, chasing its vanished anchor. Its submenus
(`Project`, `Organization`) fly out to the right, past `PEEK_CLOSE_MARGIN`, so
hovering them triggers the same close.

## Changes

Add a "hold" to `sidebarPeekStore`: while a sidebar-spawned menu is open,
`endSidebarPeek` is a no-op, so a pointer that leaves the panel can't collapse the
peek and strand the open menu's anchor.

- `holdSidebarPeek()` / `releaseSidebarPeek()` guard a module-level `held` flag;
  `endSidebarPeek` early-returns while held, and `cancelSidebarPeek` clears it.
- `ProjectSwitcher` holds on open and releases on close (and on unmount) via
  `onOpenChange`.
- Hover behavior is unchanged when no menu is open; once the menu closes, the peek
  collapses on the next directional exit as before.

## Before / after

https://github.com/user-attachments/assets/e39a23b4-3fae-4c1a-9f3b-02ed8c9528fb

## How did you test this?

- New unit test `sidebarPeekStore.test.ts` (4 cases: hold blocks `endSidebarPeek`,
  release restores it, hold cancels a pending hide, `cancelSidebarPeek` clears the
  hold). `pnpm --filter @posthog/ui test` passes.
- `pnpm --filter @posthog/ui typecheck` and Biome pass.
- Manually in the dev app: peek the collapsed sidebar, open the ProjectSwitcher, move
  the cursor into the content and onto the submenus (stays open); close the menu, then
  move the cursor out (collapses normally).

---

PR #3666, opened 2026-07-22
fix(sidebar): keep peeked sidebar open while the task Filter menu is open

aleexzxzxzx:
## Problem

With sidebar hover-peek active, opening the tasks-header **Filter** dropdown and then moving the pointer off the panel collapsed the sidebar while the menu (rendered in a portal anchored to its trigger) stayed open, leaving it floating over the content, chasing its vanished anchor.

Same failure mode as #3483, but for the task Filter menu instead of the ProjectSwitcher.

Closes #3665

## Changes

- `TaskFilterMenu` now holds the sidebar peek on open and releases it on close (and on unmount), reusing `holdSidebarPeek`/`releaseSidebarPeek` from #3484.
- Hover behaviour is unchanged when no menu is open.

## Before / after

https://github.com/user-attachments/assets/02652562-99ea-4e27-855c-d3cb83dda609

## How did you test this?

- Sidebar unit tests pass (46/46), including the existing peek-hold store tests.
- Manually: collapsed sidebar → hover to peek → open Filter → move pointer out of the panel → sidebar stays; close menu → hover-collapse resumes.

### DIFF
--- a/packages/ui/src/features/sidebar/components/ProjectSwitcher.tsx
+++ b/packages/ui/src/features/sidebar/components/ProjectSwitcher.tsx
@@ -49,34 +49,24 @@ import {
 import { useCurrentUser } from "@posthog/ui/features/auth/useCurrentUser";
 import { useProjects } from "@posthog/ui/features/projects/useProjects";
 import { openSettings } from "@posthog/ui/features/settings/hooks/useOpenSettings";
-import {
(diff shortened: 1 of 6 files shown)


charlesvien:
Thanks for this @aleexzxzxzx!

aleexzxzxzx:
@charlesvien  happy to help!

PR #3671, opened 2026-07-22
feat(canvas): task-centric activity feed

raquelmsmith:
## Problem

Channels needs one Activity view for tasks that require attention. The unread badge and list must update when an agent finishes or requests permission, regardless of whether the task started from the new-task page or a channel.

Opening a channel must not mark its visible task previews read. Opening the specific task or thread should clear only that task while keeping the activity row in the list.

## Changes

Adds a task-centric Activity view backed by the companion endpoint in [PostHog/posthog #72793](https://github.com/PostHog/posthog/pull/72793).

- Shows one row per task with completed, awaiting-input, reply, mention, and creation activity.
- Displays unread rows and the aggregate unread count in the Activity navigation item.
- Loads the full feed with cursor pagination.
- Registers newly created cloud runs with the existing `SessionService` SSE watcher, including tasks started inside channels.
- Projects completion and permission events into the activity cache immediately, using the same typed notification events that drive toasts.
- Removes activity polling and keeps cross-feature coordination in a boot contribution rather than React effects.
- Marks only the opened task read. Channel feed previews remain read-only, and opening a task from its card, thread, or Activity marks that task.
- Keeps read rows in the Activity list and protects newer events with per-task read timestamps.
- Uses agent attribution for agent activity and formats personal and named channels with a `#` prefix.

The feature remains behind the existing `project-bluebird` flag and depends on the backend endpoint.

## How did you test this?

Automated regression coverage verifies:

- channel-created cloud tasks attach to the existing event stream
- normal and warm-created cloud runs are watched before they are surfaced
- completion and permission events immediately create unread activity
- thread previews do not mark activity read
- opening a task clears only that task
- read activity remains visible
- newer activity is not cleared by an older marker
- pagination loads every page

Ran focused Vitest suites, Biome, package typechecks, the full 20-package monorepo typecheck, and host-boundary checks. CI unit and integration tests pass.

Manual testing by the human reviewer covered channel-created tasks, toasts, permission and completion activity, unread badges, channel navigation, and opening tasks through both Activity and channel cards.

## Automatic notifications

- [ ] Publish to changelog?

---


packages/ui/src/features/canvas/hooks/useTaskActivity.ts
```diff
@@ -0,0 +1,61 @@
...
+    },
+    initialPageParam: undefined as
+      | { before: string; beforeId: string }
+      | undefined,
+    getNextPageParam: (page: TaskActivityPage) =>
+      page.next_before && page.next_before_id
+        ? { before: page.next_before, beforeId: page.next_before_id }
+        : undefined,
+    enabled: !!client && (options?.enabled ?? true),
+    staleTime: Number.POSITIVE_INFINITY,
```
tatoalo, line 43: this permanently caches the feed right? do we want this as such?

tatoalo (approved):

PR #3677, opened 2026-07-22
feat(sessions): Add clickable minimap of user messages

thmsobrmlr:
## TL;DR

Long chats are hard to move around in. This adds a row of small dots down the right edge of the chat — one per message you sent. Click a dot to jump to that message; hover to see what it said. The dot for the message you're reading right now is highlighted.

## Problem

In long agent conversations there is no quick way to see where your own messages sit or to jump back to one. The Cmd/Ctrl+J picker and Alt+Up/Down exist, but nothing gives a spatial, always-visible overview of the thread.

## Changes

- New `ConversationMinimap`: a vertical rail docked left of the scrollbar with one clickable marker per user message, hover tooltips with a one-line preview, and an accent on the message currently on screen. Hidden when the conversation has fewer than two user messages and in compact embeds (command center, canvas panel).
- Wired into both thread renderers behind the `useNewChatThread` setting: the new `ChatThread` uses the scroller engine's `currentAnchorId` and `scrollToMessage`; the legacy `ConversationView` reuses its jump-to-message path and tracks the current message through a new `onFirstVisibleRowChange` callback on `VirtualizedList` (fires only when the topmost row changes, so scrolling does not re-render the thread).

## How did you test this?

- Added `ConversationMinimap.test.tsx` (6 tests: marker filtering/preview truncation, hidden under two messages, click-to-jump, aria-current on the active marker) — passing.
- `pnpm --filter @posthog/ui test src/features/sessions` (478 tests) and `pnpm --filter @posthog/ui typecheck` — both clean; Biome clean on changed files.
- Could not run the Electron app in this sandbox (no display), so no screenshot; the rail follows the same overlay conventions as the existing scroll-to-bottom button and sticky header.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/sessions/components/ConversationMinimap.test.tsx
+++ b/packages/ui/src/features/sessions/components/ConversationMinimap.test.tsx
@@ -1,0 +1,94 @@
+import { fireEvent, render, screen, within } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import type { ConversationItem } from "./buildConversationItems";
+import {
+  ConversationMinimap,
+  type MinimapMessage,
+  toMinimapMessages,
+} from "./ConversationMinimap";
+
+function userMessage(id: string, content: string): ConversationItem {
+  return { type: "user_message", id, content, timestamp: 0 };
+}
+
+describe("toMinimapMessages", () => {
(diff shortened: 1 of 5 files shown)


charlesvien:

PR #3682, opened 2026-07-22
fix(sessions): notify cloud turn completions in either log order

charlesvien:
## Problem

Since #3572, desktop notifications for cloud task completions only fire when `_posthog/turn_complete` happens to land before the `session/prompt` response in the log stream. The two writes race in the agent's log stream, so roughly half of all cloud completions never notify at all: no banner, no sound, not even later.

The `completedActiveTurn` gate from #3572 treats a turn as already notified once `currentPromptId` is cleared, but the prompt response also clears it. When the response lands first, the turn is disarmed before `turn_complete` arrives and the notification is suppressed. The old comment claiming "Cloud sessions never see that response" is wrong: the agent taps both directions of the ACP stream into the log, and a survey of 33 completed turns in real session logs found 14 with the response first and 19 with `turn_complete` first.

Note: this is one of two issues behind the "notifs don't fire until I refocus the app" report. The other (notifications bursting on refocus instead of arriving while the app is backgrounded) is a delivery stall around App Nap and the cloud watcher's permanent give-up, and needs a separate fix.

## Changes

Cloud sessions no longer disarm the turn on the `session/prompt` response; only `_posthog/turn_complete` does. The response still clears `isPromptPending`, so spinner behavior is unchanged, and local sessions are untouched. The `turn_complete` handler now reads the entry's real `stopReason` instead of hardcoding `end_turn`, so cancelled or failed turns no longer ring or speak "done". Keeping the turn start time intact until `turn_complete` also fixes completion sound duration scaling for cloud tasks, which the response previously wiped.

## How did you test this?

- New parameterised cases in `cloudTaskUpdateNotifications.test.ts`: response before and after `turn_complete`, duplicate `turn_complete` after a response-first turn, multi-turn with mixed orderings, a cancelled turn and a duration assertion.
- Red/green verified: the response-first cases fail on main's `sessionService.ts` (4 failures) and pass with the fix.
- `@posthog/core`: full suite (2490 tests), typecheck and `biome lint` clean. `@posthog/ui`: sessions and notifications suites (542 tests) pass after updating one hydration test that encoded the old disarm behavior; workspace typecheck clean.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/core/src/sessions/cloudTaskUpdateNotifications.test.ts
+++ b/packages/core/src/sessions/cloudTaskUpdateNotifications.test.ts
@@ -6,14 +6,32 @@ import { SessionService, type SessionServiceDeps } from "./sessionService";
 const TASK_ID = "task-1";
 const RUN_ID = "run-1";
 
-function turnComplete(timestamp?: string): StoredLogEntry {
+function turnComplete(
+  timestamp?: string,
+  stopReason = "end_turn",
+): StoredLogEntry {
   return {
     type: "notification",
     timestamp,
     notification: {
       method: "_posthog/turn_complete",
-      params: { sessionId: RUN_ID, stopReason: "end_turn" },
+      params: { sessionId: RUN_ID, stopReason },
     },
+  };
+}
+
+// The agent's JSON-RPC response to `session/prompt`. Real logs carry it next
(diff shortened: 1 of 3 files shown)


richardsolomou (approved):
my b

richardsolomou:
meep

PR #3704, opened 2026-07-22
Relay background-turn responses to Slack

VojtechBartos:
## Problem

When the cloud agent starts a background process and ends its turn to wait for it ("I'll wait for the completion notification"), the wake-up turn that eventually produces the answer never reaches Slack. A background (task-notification) turn resolves no tracked `prompt()` call, and `relayAgentResponse` only fires at the three prompt-resolution sites, so the relay for that turn simply never happens. The agent's transcript contains the answer, the user's thread doesn't, and the agent has no way to know the reply was dropped.

Origin: https://posthog.slack.com/archives/C09SK2PAGKF/p1784716116964029

## Changes

- `agent-server.ts`: the ACP transport tap now triggers `relayAgentResponse` when a clean `_posthog/background_turn_complete` (stop reason `end_turn`) passes through. Refusal/error background turns stay silent, matching the tracked path's `end_turn` gating.
- `session-log-writer.ts`: relays now consume prose through a new `takeUnrelayedAgentResponseParts`, which returns only the text blocks no prior relay has delivered. Background turns append to the same turn buffer as the tracked turn before them (only a new prompt resets it), so an unscoped read would re-post already-delivered prose, and a silent wake-up (e.g. the agent just reschedules) would re-post the previous answer verbatim.
- Tracked-turn relays behave exactly as before: the buffer resets at prompt start, so the first take returns the full turn.

Companion backend PR (accepts late relays from completed runs): PostHog/posthog#72897

## How did you test this?

Automated only, all in `packages/agent`:

- New: `agent-server.test.ts` "background turn relay" — relays prose on a clean `background_turn_complete`, skips non-`end_turn` stop reasons, skips silent wake-ups.
- New: `session-log-writer.test.ts` — take-cursor semantics (only prose since previous take; reset starts over).
- Updated existing relay stubs (`question-relay.test.ts`, relay-echo tests) to the consuming API.
- Full runs: `agent-server.test.ts` (150 passed), `session-log-writer.test.ts` + `question-relay.test.ts` (87 passed), `claude-agent.task-notification.test.ts` (2 passed), `tsc --noEmit`, `biome check`.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/agent/src/server/agent-server.test.ts
+++ b/packages/agent/src/server/agent-server.test.ts
@@ -479,6 +479,83 @@ describe("AgentServer HTTP Mode", () => {
     testServer.session = null;
   });
 
+  describe("background turn relay", () => {
+    // A background (task-notification) turn resolves no tracked prompt, so
+    // the prompt-resolution relay sites never fire — the transport tap must
+    // relay its prose when _posthog/background_turn_complete passes through.
+    const stubRelaySession = (parts: string[] | undefined) => {
+      const testServer = createServer() as unknown as {
+        session: unknown;
+        posthogAPI: PostHogAPIClient;
+        handleAcpTransportMessage(message: unknown): void;
+      };
+      const relaySpy = vi
+        .spyOn(testServer.posthogAPI, "relayMessage")
+        .mockResolvedValue(undefined);
+      testServer.session = {
+        payload: { task_id: "test-task-id", run_id: "test-run-id" },
+        sseController: null,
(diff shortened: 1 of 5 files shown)


charlesvien:
Hey @VojtechBartos, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3719, opened 2026-07-22
feat: add durable cloud Pi runtime sessions

jonathanlab:
## Summary

- add a native Pi runtime path alongside ACP while adapting both runtimes to the shared `AgentConversationEvent` / ChatThread UI
- add browser-safe native Pi RPC transport, local pooled sessions, cloud session routing, model/thinking/queue controls, retries, cancellation, compaction, bash, and tool rendering
- use Django-owned task runtime selection and the shared task-creation saga for ACP and Pi
- restore cloud Pi sessions through Pi's native `SessionManager`/RPC flow and wait for `pi_run_started` before startup commands
- persist translated Pi conversation events for UI replay while keeping native TaskSession JSONL authoritative for runtime restoration
- resume completed cloud Pi runs before reconnecting and clear/recreate bound session transports
- share cloud task/domain types and host-router implementations across desktop and web
- upgrade the Pi packages to 0.81.0

Backend dependency: https://github.com/PostHog/posthog/pull/72928

## Architecture

Pi does not impersonate ACP. ACP and Pi independently adapt to `AgentConversationEvent`; ChatThread renders that runtime-neutral model. Native Pi RPC remains canonical, while TaskSession JSONL is restored by Pi itself rather than parsed or rewritten by product code.

## Diff distribution

The current `+5,027 / -1,762` diff is split between:

- **2,606 production-code additions (52%)**
- **2,341 test additions (47%)**
- **80 build/package configuration additions (1%)**

## Testing

- full workspace typecheck after rebasing onto current `main`
- agent focused tests, build, and package verification
- core Pi session/provider/controller and task-creation tests
- workspace-server and host-router focused tests/typechecks
- full core and harness suites during development
- Biome and `git diff --check`
- real Electron + local Django/Temporal/Docker E2E:
  - fresh cloud Pi task
  - model response and follow-up
  - replacement-sandbox native JSONL restore
  - full app restart and conversation restore
  - follow-up after restart
  - durable event projection and TaskSession revision promotion

## Rollout

Merge/deploy the backend dependency first, then publish the updated `@posthog/agent` and rebuild the production sandbox image before enabling the client path behind `pi-harness`.


packages/ui/src/features/task-detail/components/TaskDetail.tsx
```diff
@@ -51,6 +51,17 @@ export function TaskDetail({
...
+  const selectedTaskRunRef = useRef({
+    taskId,
+    taskRunId: initialTask.latest_run?.id,
+  });
+  if (selectedTaskRunRef.current.taskId !== taskId) {
+    selectedTaskRunRef.current = {
+      taskId,
+      taskRunId: initialTask.latest_run?.id,
+    };
+  }
```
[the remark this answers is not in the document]
jonathanlab: Fixed in c98e238b9f. `PiSessionView` now follows the live `task.latest_run.id`, so a resumed cloud Pi run switches away from the terminal run.

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,651 @@
...
+
+  private syncTaskSession(): Promise<void> {
+    const sync = this.sessionSyncQueue.then(async () => {
+      if (!this.sessionFile) {
+        return;
+      }
+
+      let content: string;
+      try {
+        content = await readFile(this.sessionFile, "utf8");
```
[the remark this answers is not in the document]
jonathanlab: Fixed in c98e238b9f. Removed periodic reads of the live JSONL; synchronization now happens at Pi’s `agent_settled` boundary, plus initialization and controlled shutdown/fatal paths.

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,651 @@
...
+        this.evictCompletedUserMessageDeliveries();
+      }
+      return result;
+    } catch (error) {
+      if (this.userMessageDeliveries.get(messageId) === delivery) {
+        this.userMessageDeliveries.delete(messageId);
+      }
+      throw error;
+    }
+  }
```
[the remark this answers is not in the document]
jonathanlab: Fixed in 419c5403f6 by binding every connection token to the server’s configured task, run, and team. Same-run SSE reconnects reuse the existing session, while a resumed run gets a new sandbox/server, so an authenticated request can no longer replace the session underneath an in-flight delivery.

packages/agent/src/pi/remote-rpc-client.ts
```diff
@@ -0,0 +1,164 @@
...
+    return parsePiRpcResponse(await this.transport.request(command));
+  }
+
+  private data<T>(response: RpcResponse): T {
+    if (!response.success) {
+      throw new Error(response.error);
+    }
+    return (response as unknown as { data: T }).data;
+  }
+}
```
jonathanlab, line 164: This seems like unnecessary plumbing. Why do we need to redefine every operation here as a wrapped function?

packages/agent/src/pi/rpc-client.ts
```diff
@@ -76,6 +57,7 @@ class SecurePiRpcClient extends RpcClient {
   constructor(
     private readonly secureOptions: RpcClientOptions,
     private readonly providerOptions: PiRpcProviderOptions,
+    private readonly sessionDir?: string,
```
jonathanlab: Why do we need to plumb through `sessionDir` in the `rpc-client` and other parts?

packages/agent/src/pi/rpc-transport.ts
```diff
@@ -0,0 +1,61 @@
...
+      command: z.string(),
+      success: z.literal(false),
+      error: z.string(),
+    })
+    .loose(),
+]);
+
+export function parsePiRpcResponse(value: unknown): RpcResponse {
+  return piRpcResponseSchema.parse(value) as RpcResponse;
+}
```
jonathanlab, line 42: Where should we handle invalid Pi RPC responses? currently we're not handling the respones of `.parse` at all

packages/agent/src/pi/types.ts
```diff
@@ -32,9 +32,7 @@ export type PiNativeModelInfo = Awaited<
   ReturnType<RpcClient["getAvailableModels"]>
 >[number];
 
-export type PiModelOption = PiNativeModelInfo & {
-  thinkingLevels: PiThinkingLevel[];
-};
+export type PiModelOption = PiNativeModelInfo;
```
jonathanlab, line 33: We do not need this anymore now that it's just an alias

packages/agent/src/server/bin.ts
```diff
@@ -33,6 +37,8 @@ const envSchema = z.object({
     })
     .regex(/^\d+$/, "POSTHOG_PROJECT_ID must be a numeric string")
     .transform((val) => parseInt(val, 10)),
+  POSTHOG_AGENT_PROTOCOL: z.enum(["acp", "pi"]).optional(),
```
jonathanlab: Since when did we move to `AGENT_PROTOCOL` instead of `runtime`? Shouldn't this be something like `AGENT_RUNTIME`? Our Task column also has a `runtime` column which can be either `acp` or `pi`. this is confusing

packages/agent/src/server/bin.ts
```diff
@@ -33,6 +37,8 @@ const envSchema = z.object({
...
+  POSTHOG_SANDBOX_ID: z.string().min(1).optional(),
```
jonathanlab, line 40: Question: Why are we plumbing the `POSTHOG_SANDBOX_ID` here/why do we need to know about this ID at all?

packages/agent/src/server/bin.ts
```diff
@@ -241,10 +248,19 @@ program
       baseBranch: options.baseBranch,
       claudeCode,
       allowedDomains,
+      protocol: env.POSTHOG_AGENT_PROTOCOL,
+      piRpcHostPath: resolve(
```
jonathanlab: This seems duplicated from what we're doing in `rpc-client.ts`.. or am I wrong?

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,659 @@
...
+}
+
+interface PiCloudSession {
+  payload: JwtPayload;
+  runtime: PiRuntime;
+  sseController: SseController | null;
+  unsubscribe: () => void;
+}
+
+const COMPLETED_USER_MESSAGE_DELIVERY_LIMIT = 500;
```
jonathanlab: What is this limit? Why do we need it..?

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,659 @@
...
+
+const userMessageCommandSchema = z.object({
+  content: z.string().min(1),
+  messageId: z.string().min(1).optional(),
+});
+
+const commandSchemas = {
+  user_message: userMessageCommandSchema,
+  cancel: emptySchema,
+  "pi/rpc": z.object({ command: piRpcCommandSchema }),
```
jonathanlab, line 58: `pi/rpc` makes sense to me. What's the use of `cancel` and `user_message`..? These are provided by the pi RPC already right?

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,659 @@
...
+      mode: this.config.mode,
+    };
+    await this.initializeSession(payload, null);
+  }
+
+  async stop(): Promise<void> {
+    const session = this.session;
+    if (session) {
+      await session.runtime.client.abort().catch(() => undefined);
+      await session.runtime.client.waitForIdle(5_000).catch(() => undefined);
```
jonathanlab: Why are we not logging any errors here

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,659 @@
...
+    );
+    await this.eventStreamSender?.stop();
+    this.server?.close();
+    this.server = null;
+  }
+
+  async reportFatalError(error: unknown): Promise<void> {
+    const message = error instanceof Error ? error.message : String(error);
+    this.broadcast({
+      type: "pi_event",
```
jonathanlab, line 197: What is a `pi_event`? Do we use it anywhere else? Seems strange to me

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,659 @@
...
+      this.flushConversationLog(),
+    ]).catch((syncError) =>
+      this.logger.error("Failed to persist crashed Pi session", syncError),
+    );
+    await this.posthogAPI
+      .updateTaskRun(this.config.taskId, this.config.runId, {
+        status: "failed",
+        error_message: `Pi agent server crashed: ${message}`,
+      })
+      .catch(() => undefined);
```
jonathanlab: Why are we not logging anything here

packages/agent/src/server/pi-agent-server.ts
jonathanlab, line 1: General question: How much code here is duplicated exactly from the other server and is it worth deduplicating

packages/agent/src/posthog-api.ts
```diff
@@ -210,10 +223,115 @@ export class PostHogAPIClient {
...
+    );
+  }
+
+  async downloadTaskSession(access: TaskSessionStorageAccess): Promise<string> {
+    const response = await fetch(access.download_url, {
+      signal: AbortSignal.timeout(30_000),
+    });
+    if (response.status === 404) {
+      return "";
+    }
```
jonathanlab, line 241: We should not just return `""` here right, but throw a proper error

packages/agent/tsup.config.ts
```diff
@@ -167,9 +182,12 @@ export default defineConfig([
...
   {
     entry: { "server/bin": "src/server/bin.ts" },
-    format: ["cjs"],
+    format: ["esm"],
     dts: false,
     clean: false,
+    banner: {
+      js: 'import { createRequire as __createRequire } from "node:module"; const require = __createRequire(import.meta.url);',
+    },
     ...sharedOptions,
```
jonathanlab: This seems hacky

packages/core/src/pi-runtime/cloudPiSessionClient.ts
```diff
@@ -0,0 +1,407 @@
...
+} from "../cloud-task/schemas";
+import type { PiSession } from "./piSessionController";
+
+const readinessCommands = new Set<RpcCommand["type"]>([
+  "get_state",
+  "get_entries",
+  "get_available_models",
+  "get_available_thinking_levels",
+  "get_commands",
+]);
```
jonathanlab: What are `readinessCommands`? This seems hacky and unnecessary.. We can determine cloud session liveliness in other ways now.

packages/core/src/pi-runtime/cloudPiSessionClient.ts
```diff
@@ -0,0 +1,407 @@
...
+    } else if (command.type === "get_commands") {
+      data = { commands: [] };
+    } else if (command.type === "get_entries" && this.hasSnapshot) {
+      data = { entries: [] };
+    } else {
+      throw new Error(
+        `Cloud task run ${this.context.runId} is ${this.runStatus}`,
+      );
+    }
+
```
jonathanlab: Same issue here, this is hacky

packages/core/src/pi-runtime/piSessionProvider.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (
+        isTerminalStatus(run.status) ||
+        typeof run.state?.sandbox_url === "string"
+      ) {
+        return run.status;
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, delayMs));
+      delayMs = Math.min(delayMs * 2, 5_000);
+    }
```
jonathanlab: Should this even be responsible for waiting for cloud run readiness?

packages/core/src/pi-runtime/piSessionController.ts
```diff
@@ -150,40 +164,38 @@ export class PiSessionController {
...
+      const command = parseCommandLine(message);
+      const customInstructions = command?.args?.trim() || undefined;
+      await session.client.compact(customInstructions);
+      await this.refreshConversation(taskId);
+    } else if (action === "prompt") {
+      await session.client.prompt(message);
+    } else if (action === "steer") {
+      await session.client.steer(message);
+    } else {
+      await session.client.followUp(message);
```
jonathanlab: what are these for..? isn't this an rpc concern

jonathanlab: **Agent response:** Pi RPC exposes separate `prompt`, `steer`, and `follow_up` operations, but it cannot choose the product action intended by the current composer mode. The controller owns that decision using streaming state and the selected messaging mode, then invokes the corresponding native RPC operation. The RPC layer remains responsible only for transporting the selected command.

packages/agent/src/server/pi-agent-server.ts
jonathanlab: **Agent response:** This bounds the successful message-ID deduplication cache. Temporal can redeliver a `user_message`, and retaining its ID prevents the same prompt from being submitted twice; without a limit, a long-running server would retain IDs indefinitely. `500` provides a large recent retry window while keeping memory bounded.

packages/agent/src/pi/remote-rpc-client.ts
jonathanlab: **Agent response:** The upstream `RpcClient` is a concrete Node client that owns and spawns a child process, so it cannot be instantiated in the browser or supplied with the cloud transport. `RemotePiRpcClient` preserves Pi’s native typed method interface while mapping those methods to `RpcCommand`s over an injected transport and validating/unwrapping the responses. This adapter can go away if upstream exposes a transport-neutral client, but removing it now would move the same mapping into callers and lose method-level typing.

packages/agent/src/pi/rpc-transport.ts
jonathanlab: **Agent response:** `piRpcResponseSchema.parse()` is the validation boundary: it returns a valid response or throws a `ZodError`, and that rejection propagates through `RemotePiRpcClient` into the session/controller error handling. Invalid responses therefore are not accepted or ignored. We could wrap the Zod error with a more domain-specific message, but there is no missing failure path here.

packages/agent/src/server/bin.ts
jonathanlab: **Agent response:** `POSTHOG_SANDBOX_ID` is already supplied by Django as an environment variable. `bin.ts` is the executable boundary: it validates the environment once and passes the typed value into `PiAgentServer`, rather than having runtime code read mutable global `process.env` state. The ID is then included in every TaskSession promotion so Django can reject writes from a stale/replaced sandbox.

packages/agent/src/server/pi-agent-server.ts
jonathanlab: **Agent response:** The Twig Pi client is migrated to `pi/rpc`, but backend Temporal follow-up, pending-message, and cancellation paths still call the shared `user_message` and `cancel` endpoints. Those endpoints remain for those callers and for message-ID deduplication; they can be removed after the backend paths are migrated too.

packages/agent/src/server/pi-agent-server.ts
jonathanlab: **Agent response:** The overlap is limited mainly to the HTTP shell: Hono setup, JWT authentication, SSE handling, JSON-RPC parsing, and shutdown plumbing. The ACP and Pi session/runtime logic is materially different and should remain separate. A future cleanup could extract narrow stable transport/auth helpers, but a shared base server would couple two independent runtimes and is not warranted in this change.

packages/agent/src/pi/rpc-client.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. I verified that `SessionManager.create(cwd)` chooses a native directory and exposes the generated path through `get_state.sessionFile`, while `SessionManager.open(sessionFile)` derives its directory from the restored file. The redundant `sessionDir` option and fd-3 plumbing have been removed.

packages/agent/src/pi/types.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. Removed the redundant `PiModelOption` alias and now use `PiNativeModelInfo` directly.

packages/agent/src/server/bin.ts
jonathanlab: **Agent response:** Fixed across both repositories in `cedae819f5` and `PostHog/posthog@1016b0031e8`. The runtime selector is now `POSTHOG_AGENT_RUNTIME` / `agent_runtime`, matching `Task.runtime`; the unused server config field was removed.

packages/agent/src/server/bin.ts
jonathanlab: **Agent response:** Simplified in `cedae819f5`. The packaged Pi host path now resolves directly from the ESM entrypoint with `new URL("../pi/rpc-host.js", import.meta.url)`, removing the `process.argv`/`realpathSync` path reconstruction. The explicit override remains necessary because the bundled server entry and normal `rpc-client` module have different output directories.

packages/agent/src/server/pi-agent-server.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. Best-effort abort and idle-wait failures during shutdown are now logged at debug level rather than silently discarded.

packages/agent/src/server/pi-agent-server.ts
jonathanlab: **Agent response:** A `pi_event` is the durable TaskRun-log projection of one translated `AgentConversationEvent`; `CloudPiSessionClient` consumes these entries to render history after the sandbox is gone. It contains selected conversation/runtime events, not raw Pi RPC commands or responses. Native TaskSession JSONL remains the authoritative restoration format.

packages/agent/src/server/pi-agent-server.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. Failure to mark a crashed Pi run as failed is still nonfatal to the crash handler, but is now logged with the original error.

packages/agent/src/posthog-api.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. A 404 is treated as an empty session only for revision 0, where no object exists yet. A missing promoted revision now throws instead of silently starting a new session; both cases have regression tests.

packages/agent/tsup.config.ts
jonathanlab: **Agent response:** Simplified in `cedae819f5`. `server/bin.ts` now participates in the existing ESM build entry that already owns the required CommonJS interop shim, so the duplicate tsup build block and banner were removed.

packages/core/src/pi-runtime/cloudPiSessionClient.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. Removed `readinessCommands`, command-specific readiness retries, and the task-state polling fallback. Cloud RPC now waits for the native current-run `pi_run_started` event before sending any command, without automatically replaying mutating commands.

packages/core/src/pi-runtime/cloudPiSessionClient.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. Terminal runs now use an explicit read-only terminal Pi client instead of fabricating successful Pi RPC response envelopes inside the transport. Opening terminal history remains read-only and does not provision a sandbox.

packages/core/src/pi-runtime/piSessionProvider.ts
jonathanlab: **Agent response:** Fixed in `cedae819f5`. `RoutingPiSessionProvider` now only resolves local versus cloud context and constructs the session; cloud-readiness polling was removed from the provider, and the cloud client uses the native readiness event.

tatoalo:
Stress-tested this a bit in cloud runs (replicating this also on the backed PR):

* attachments are not working
* during task log hydration and cold reload I saw that the renderer combined duplicate copies of persisted and live Pi events even though BE storage contained one (correct)
* reconnect snapshots leave RPC calls hanging
* user-message idem is not durable, they seem to stay in the process memory
* every `pi/rpc` command is passed directly to `sendPiRpcCommand` so no durable request identity is enforced

tatoalo (approved):

packages/agent/src/server/bin.ts
```diff
@@ -33,13 +36,16 @@ const envSchema = z.object({
...
+  POSTHOG_SANDBOX_ID: z.string().min(1).optional(),
   POSTHOG_CODE_RUNTIME_ADAPTER: z.enum(["claude", "codex"]).optional(),
   POSTHOG_CODE_MODEL: z.string().optional(),
   POSTHOG_CODE_REASONING_EFFORT: z
-    .enum(["low", "medium", "high", "xhigh", "max"])
+    .enum(["off", "minimal", "low", "medium", "high", "xhigh", "max"])
     .optional(),
   POSTHOG_AGENT_STATE_DIR: z.string().startsWith("/").optional(),
   POSTHOG_TASK_RUN_EVENT_INGEST_TOKEN: z.string().min(1).optional(),
+  POSTHOG_TASK_RUN_SESSION_TOKEN: z.string().min(1).optional(),
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. The validated session token is copied into server configuration and then removed from `process.env`, matching the existing OTEL credential handling.

packages/agent/src/pi/conversation/translatePiConversation.ts
```diff
@@ -102,6 +111,76 @@ export function createPiConversationTranslator(): PiConversationTranslator {
...
+  function completeDirectBash(
+    result: PiDirectBashResult,
+  ): AgentConversationEvent[] {
+    const failed = result.cancelled || (result.exitCode ?? 0) !== 0;
+    return finishDirectBash(failed ? "failed" : "completed", result.output);
+  }
+
+  function failDirectBash(message: string): AgentConversationEvent[] {
+    return finishDirectBash("failed", message);
+  }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. A transport failure now combines already-streamed bash output with the terminal error instead of replacing the output.

packages/agent/src/pi/conversation/translatePiConversation.ts
```diff
@@ -322,14 +438,26 @@ export function createPiConversationTranslator(): PiConversationTranslator {
...
       ];
+      if (event.result?.summary) {
+        events.push({
+          type: "assistant_message_chunk",
+          timestamp,
+          content: { type: "text", text: event.result.summary },
+        });
+      }
+
+      return events;
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. The synthesized summary timestamp now advances `latestConversationTimestamp`, preserving monotonic ordering for subsequent events.

packages/core/src/pi-runtime/cloudPiSessionClient.ts
```diff
@@ -0,0 +1,483 @@
...
+    onCloudStatus?: (status: TaskRunStatus) => void,
+  ): void {
+    const snapshotCanProveReadiness =
+      update.kind === "snapshot" && update.status === "in_progress";
+    const hasCurrentReadinessEvent =
+      (update.kind === "logs" || snapshotCanProveReadiness) &&
+      update.newEntries.some((entry) => entry.type === "pi_run_started");
+    if (hasCurrentReadinessEvent) {
+      this.markRuntimeReady();
+    }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Any run-scoped snapshot containing `pi_run_started` now proves readiness; it no longer depends on the snapshot status already being `in_progress`.

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,741 @@
...
+      await session.runtime.client.stop();
+    }
+    this.session = null;
+    await this.flushConversationLog().catch((error) =>
+      this.logger.error("Failed to persist Pi events during shutdown", error),
+    );
+    await this.eventStreamSender?.stop();
+    this.server?.close();
+    this.server = null;
+  }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Shutdown logs and contains a Pi client stop failure so transcript flushing, event-stream shutdown, and server closure still run.

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,741 @@
...
+        ),
+      )
+      .then(() => undefined)
+      .catch((error) => {
+        this.pendingLogEntries = [...entries, ...this.pendingLogEntries];
+        throw error;
+      });
+    this.logFlushQueue = flush.catch(() => undefined);
+    return flush;
+  }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. SSE replay and pending log entries now have explicit caps, long turns flush logs in bounded batches, and only one log drain can be active at a time.

packages/core/src/pi-runtime/piSessionController.ts
```diff
@@ -293,44 +648,429 @@ export class PiSessionController {
...
+
+    for (const message of messages) {
+      if (message.mode === "steer") {
+        await session.client.steer(message.content);
+      } else {
+        await session.client.followUp(message.content);
+      }
+    }
+    this.queuesToRestore.delete(taskId);
   }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. The captured restore snapshot is consumed before delivery, so a later queue-command failure cannot replay a prompt that was already accepted.

packages/core/src/pi-runtime/piSessionController.ts
```diff
@@ -1,78 +1,134 @@
...
 @injectable()
 export class PiSessionController {
   readonly store: PiSessionStore = createPiSessionStore();
 
+  private readonly sessions = new Map<string, Promise<PiSession>>();
   private readonly subscriptions = new Map<string, () => void>();
   private readonly liveEvents = new Map<string, AgentConversationEvent[]>();
   private readonly connections = new Map<string, Promise<void>>();
   private readonly readiness = new Map<string, Promise<void>>();
+  private readonly sessionVersions = new Map<string, number>();
```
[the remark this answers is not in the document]
jonathanlab: invalid as a bug — deleting a per-task version during disconnect can let a still-settling operation compare equal after a reconnect resets to the same version, reintroducing the stale-session race the generation protects against. The retained UUID→number entry is tiny; making cleanup collision-safe would add disproportionate complexity.

packages/workspace-server/src/services/pi-session/pi-session.ts
```diff
@@ -161,227 +170,35 @@ export class PiSessionService extends TypedEventEmitter<PiSessionEvents> {
...
-      prompt,
-      images,
-      timeout,
-    );
-  }
-
-  stderr(taskId: string): string {
-    return this.requireSession(taskId).client.getStderr();
+  clearQueue(taskId: string): Promise<PiQueueSnapshot> {
+    return this.requireSession(taskId).client.clearQueue();
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Generic RPC, queue reads, and queue clears now share one active-request wrapper, including LRU touch and post-request pool enforcement.

packages/ui/src/features/pi-sessions/PiSessionView.tsx
```diff
@@ -98,6 +124,16 @@ export function PiSessionView({ taskId }: PiSessionViewProps) {
...
+  const handleControllerError = useCallback(
+    (error: unknown, fallback: string) => {
+      if (error instanceof PiOperationError) {
+        return;
+      }
+      toast.error(fallback);
+    },
+    [],
+  );
+
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Every controller callback failure is now written through the scoped UI logger before structured-operation handling or fallback toast handling.

packages/ui/src/features/pi-sessions/PiSessionView.tsx
```diff
@@ -163,96 +200,259 @@ export function PiSessionView({ taskId }: PiSessionViewProps) {
...
+      .catch((error) =>
+        handleControllerError(error, "Failed to reconnect to Pi"),
+      );
+  }, [handleControllerError, piSessionController, taskId]);
+
+  const restart = useCallback(() => {
+    void piSessionController
+      .restart(taskId)
+      .catch((error) => handleControllerError(error, "Failed to restart Pi"));
+  }, [handleControllerError, piSessionController, taskId]);
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Retry and Restart now set/check the existing connecting state before their first await, preventing duplicate cloud recovery calls.

packages/ui/src/features/task-detail/components/TaskInput.tsx
```diff
@@ -237,6 +252,23 @@ export function TaskInput({
...
+    () => (cloudRegion ? getCloudUrlFromRegion(cloudRegion) : null),
+    [cloudRegion],
+  );
+  const { data: piModelCatalog = [], isPending: isPiConfigLoading } = useQuery({
+    ...trpc.agent.getPiModelCatalog.queryOptions({
+      apiHost: piApiHost ?? "",
+      region: cloudRegion ?? "us",
+    }),
+    enabled: runtime === "pi" && piApiHost !== null,
+  });
```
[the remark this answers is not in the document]
jonathanlab: invalid as a bug — this is a potential cache-tuning optimization, not a correctness defect or a demonstrated meaningful performance regression. TanStack Query still deduplicates concurrent requests and retains the result in its normal cache.

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,741 @@
...
+      cwd,
+      sessionFile: restoredSessionFile,
+      providerOptions: {
+        apiKey: this.config.apiKey,
+        baseUrl: resolveLlmGatewayUrl(
+          process.env.LLM_GATEWAY_URL,
+          this.config.apiUrl,
+        ),
+      },
+    });
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. `PiAgentServer` now passes the configured task model into `createPiRpcClient`, which starts the PostHog provider with that model.

packages/agent/src/pi/runtime.ts
```diff
@@ -23,6 +22,10 @@ export class PiRuntime {
   private readonly conversationListeners = new Set<
     (event: AgentConversationEvent) => void
   >();
+  private readonly pendingUserMessages: Array<{
+    id: string;
+    message: string;
+  }> = [];
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Pending entries now retain their command type, and both cloud and local queue-clear paths remove steer/follow-up IDs while preserving a live prompt ID.

packages/harness/src/extensions/posthog-provider/models.ts
```diff
@@ -239,20 +235,33 @@ async function fetchGatewayModels(
   }
 }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. `fetchPosthogPiModelCatalog` now uses fallback models without network access under `PI_OFFLINE` or `HARNESS_STATIC_MODELS`, with regression coverage.

packages/agent/src/pi/conversation/translatePiMessage.ts
```diff
@@ -43,11 +44,27 @@ function isPiToolName(name: string): name is PiToolName {
...
+    .filter((block) => block.type === "text")
+    .map((block) => block.text)
+    .join("");
+
+  if (!text) {
+    return undefined;
+  }
+
+  return [{ type: "content", content: { type: "text", text } }];
+}
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Generic tool-result rendering now preserves image blocks while retaining the prior adjacent-text coalescing behavior.

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,741 @@
...
+        status: "failed",
+        error_message: `Pi agent server crashed: ${message}`,
+      })
+      .catch((updateError) =>
+        this.logger.error(
+          "Failed to mark crashed Pi run as failed",
+          updateError,
+        ),
+      );
+    await this.eventStreamSender?.stop();
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Crash reporting now drains settled persistence and syncs the native TaskSession before flushing the crash event and marking the run failed.

packages/workspace-server/src/services/pi-session/pi-session.ts
```diff
@@ -161,227 +170,35 @@ export class PiSessionService extends TypedEventEmitter<PiSessionEvents> {
...
+        await this.persistSessionState(taskId);
+      }
 
-    if (!result.cancelled) {
-      await this.persistSessionState(taskId);
+      return response;
+    } finally {
+      session.activeRequestCount -= 1;
+      void this.enforceHotPoolLimit();
     }
```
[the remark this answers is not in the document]
jonathanlab: invalid as a bug — the cancelled operations do not rebind Pi, and `persistSessionState` rereads the unchanged live `sessionFile`; the result is one redundant idempotent metadata write, not incorrect state, data loss, or corruption.

packages/ui/src/features/task-detail/components/TaskInput.tsx
```diff
@@ -980,6 +1034,30 @@ export function TaskInput({
...
+    [sessionId, setLastUsedAgentRuntime],
+  );
+
+  const handlePiModelChange = useCallback(
+    (model: PiModelSelection) => {
+      setSelectedPiModelId(model.id);
+      setLastUsedModel(model.id);
+    },
+    [setLastUsedModel],
+  );
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Settings now persist `lastUsedPiModel` separately, so choosing a Pi model no longer overwrites the ACP model preference.

packages/agent/src/pi/conversation/translatePiConversation.ts
```diff
@@ -102,6 +111,76 @@ export function createPiConversationTranslator(): PiConversationTranslator {
...
+                {
+                  type: "content",
+                  content: { type: "text", text: output },
+                },
+              ]
+            : [],
+        },
+      },
+    ];
+  }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Small output remains smooth, while updates above 4 KiB are emitted at geometrically increasing thresholds; total retained/transmitted cumulative payload is now linear rather than quadratic.

packages/agent/src/pi/conversation/translatePiConversation.ts
```diff
@@ -102,6 +111,76 @@ export function createPiConversationTranslator(): PiConversationTranslator {
...
+        toolCall: {
+          id: toolCallId,
+          title: command,
+          kind: "execute",
+          status: "in_progress",
+          rawInput: { command },
+        },
+      },
+    ];
+  }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. `PiRuntime` rejects a second direct bash command while one is active and releases the guard in `finally`.

packages/agent/src/server/pi-agent-server.ts
```diff
@@ -0,0 +1,741 @@
...
+        } as (typeof images)[number]);
+        continue;
+      }
+
+      await mkdir(attachmentDirectory, { recursive: true });
+      const fileName = `${artifact.id}-${basename(artifact.name)}`;
+      const filePath = join(attachmentDirectory, fileName);
+      await writeFile(filePath, Buffer.from(data));
+      filePaths.push(filePath);
+    }
```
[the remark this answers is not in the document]
jonathanlab: invalid as a bug — parallel downloads could improve latency for multi-attachment prompts, but this is a low-impact optimization with no functional failure or demonstrated meaningful regression. The sequential path is correct and keeps failure behavior deterministic.

packages/core/src/pi-runtime/cloudPiSessionClient.ts
```diff
@@ -0,0 +1,483 @@
...
+  private async waitForRuntimeReady(): Promise<void> {
+    if (this.runtimeReady || isTerminalStatus(this.runStatus)) {
+      return;
+    }
+
+    await Promise.race([
+      this.runtimeReadyReceived,
+      this.terminalStatusReceived,
+    ]);
+  }
```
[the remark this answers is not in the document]
jonathanlab: invalid — the absence of a fixed frontend deadline is intentional. Cloud provisioning and backend lifecycle state are authoritative; the previous fixed timeout caused false failures for legitimate slow sandboxes. Subscription errors and terminal backend statuses already settle these waits.

packages/workspace-server/src/services/pi-session/pi-session.ts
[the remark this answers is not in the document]
jonathanlab: invalid as a bug — explicit Stop is intended to terminate the child immediately, including active work. Pi rejects pending RPC promises on process exit, so this does not hang or corrupt state; waiting for active commands would make user cancellation ineffective.

packages/ui/src/shell/GlobalEventHandlers.tsx
```diff
@@ -282,10 +287,11 @@ export function GlobalEventHandlers({
     const handleFocus = () => {
       loadFolders();
       sessionService.retryUnhealthyCloudSessions();
+      piSessionController?.retryUnhealthyCloudSessions();
     };
     window.addEventListener("focus", handleFocus);
     return () => window.removeEventListener("focus", handleFocus);
-  }, [loadFolders, sessionService]);
+  }, [loadFolders, piSessionController, sessionService]);
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. The controller now tracks tasks with mounted/active consumers and focus recovery skips sessions after their view disconnects, while preserving their transcript state for remount.

packages/ui/src/features/pi-sessions/PiSessionView.tsx
```diff
@@ -163,96 +200,259 @@ export function PiSessionView({ taskId }: PiSessionViewProps) {
...
+    if (!failure || failure.scope !== "operation") {
+      return;
+    }
+    if (failure.recoveryPrompt) {
+      draftActions.setPendingContent(
+        taskId,
+        xmlToContent(failure.recoveryPrompt),
+      );
+      draftActions.requestFocus(taskId);
+    }
```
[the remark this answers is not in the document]
jonathanlab: valid — fixed in 3890211965. Recovery text is restored only when the current draft is empty, matching the ACP safeguard and preserving text typed during authentication restoration.

jonathanlab:
ReviewHog outside-diff finding — **ChatThread hardcodes `footerEvents=[]` for Pi**: valid — fixed in `3890211965`. The Pi thread now passes its already-computed `lastTurnInfo`, `isCompacting`, and `completedToolCallCount` through to `ChatThreadFooter`, while ACP continues deriving those values from ACP events.

PR #3728, opened 2026-07-22
Add a persisted "mode after plan approval" setting

haacked:
## Problem

Approving a plan pre-selects a continue-mode that's either a sticky "last used" choice or a hardcoded "auto" fallback — there's no way to pin a specific mode as the permanent default. Pick a different mode once (e.g. for one sensitive plan), and every future plan approval silently defaults to that mode too.

Screenshot 2026-07-22 at 4 16 23 PM

## Changes

- Adds `defaultPlanApprovalMode` to the settings store (`packages/ui/src/features/settings/settingsStore.ts`), default `"last_used"` (fully backward compatible).
- Exposes it as "Mode after plan approval" in Settings → General, right below "Initial task mode" (`GeneralSettings.tsx`).
- `PlanApprovalSelector` now resolves its pre-selected mode from this pinned setting first, falling back to the existing remembered/last-used chain when it's `"last_used"` or the pinned mode isn't offered.
- Since both the Claude and Codex adapters route their plan-approval prompt through this same shared UI component, the change covers both without touching `packages/agent` or `packages/core`.

## How did you test this?

- `pnpm --filter @posthog/ui typecheck` and `pnpm --filter @posthog/ui test` (added cases to `settingsStore.test.ts` and `PlanApprovalSelector.test.tsx` covering persistence round-trip and the pinned-mode-overrides-remembered-mode / pinned-mode-not-offered branches) — all passing.
- `biome check` on the changed files.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/permissions/PlanApprovalSelector.test.tsx
+++ b/packages/ui/src/features/permissions/PlanApprovalSelector.test.tsx
@@ -53,27 +53,49 @@ function renderSelector(options: PermissionOption[]) {
 describe("PlanApprovalSelector", () => {
   beforeEach(() => {
     // Reset the remembered choice so tests don't leak through persistence.
-    useSettingsStore.setState({ lastPlanApprovalMode: null });
+    useSettingsStore.setState({
+      lastPlanApprovalMode: null,
+      defaultPlanApprovalMode: "last_used",
+    });
   });
 
   it.each([
     {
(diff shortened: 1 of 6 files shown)


haacked:
I didn't realize the current behavior is supposed to be last used by default. I think there's a bug in that. If we fix that bug, I don't think we need this PR.

PR #3733, opened 2026-07-22
Add /btw command for quick side questions

haacked:
## Problem

Users mid-conversation with Code often want a quick answer about something in the current session (what changed, what a function does) without derailing the active turn or polluting the conversation history. There was no way to do this without either interrupting the running agent or starting a whole new session.

## Changes

Adds a `/btw` command that answers a one-shot side question by forking the live session's transcript, similar to how Claude Code's own `/btw` works.

The fork is a single turn with no tools: `tools: []`, `allowedTools: []`, `mcpServers: {}`, and an always-deny `canUseTool` independently block tool use, and it gets its own session id and abort controller so nothing on the live session is touched. The answer renders as a dismissible card above the composer and never enters the session transcript, so it works both mid-turn and while idle.

The Claude adapter advertises the capability during `initialize()`, the same way it advertises `steering`. Workspace-server, the host-router, and `SessionService` all thread the capability through, and the UI gates the command on `sessionSupportsSideQuestion`, which is true only for local (non-cloud) Claude sessions.

The last commit addresses review findings from an internal pass: fixes a doc comment on `AgentService.sideQuestion` that claimed it doesn't touch the session's idle lifecycle when it actually resets the idle-kill timer, adds an aria-live region so the answer is announced to screen readers, adds a hover tooltip for a truncated question, and adds test coverage for the in-flight guard, the stale-answer guard in the store, the card's render states, and the capability gating.

Here's a screenshot of it in action:

Screenshot 2026-07-22 at 4 06 18 PM

## How did you test this?

Ran the full test suites for the touched packages (`@posthog/agent`, `@posthog/core`, `@posthog/workspace-server`, `@posthog/ui`) and `pnpm typecheck` on each; everything passes. Ran an eight-dimension code review (security, performance, correctness, maintainability, testing, compatibility, architecture, frontend) against the diff and applied the fixes it surfaced.

I haven't manually exercised the `/btw` command in the running Electron app.

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/apps/web/src/web-host-router.ts
+++ b/apps/web/src/web-host-router.ts
@@ -134,6 +134,13 @@ const agentStubRouter = router({
   }),
   // Called by resetSessionService() on logout/project switch.
   resetAll: publicProcedure.mutation(() => undefined),
+  // Unreachable in practice: the UI gates "/btw" on sessionSupportsSideQuestion,
+  // which is false for cloud sessions. Present so SessionTrpc stays satisfied.
+  sideQuestion: publicProcedure
+    .input(z.object({ sessionId: z.string(), question: z.string() }))
+    .mutation(() => {
+      throw new Error("Side questions require a local session");
+    }),
   // Model/mode/effort options for the task-input preview + cloud run creation
   // (a cloud run requires a model). Real: fetched from the CORS-open PostHog LLM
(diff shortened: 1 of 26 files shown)


charlesvien:
Remade in the monorepo as PostHog/posthog#76461 (PostHog/code main is frozen after the desktop import).

PR #3736, opened 2026-07-22
fix(code-review): order diff review files to match the changes file tree

adboio:
## Problem

In the diff review panel, files scroll past in a different order than they appear in the changes file tree beside it, so the two views disagree as you navigate.

## Why

The review list rendered files in raw git-diff order (byte/ASCII sort — uppercase paths first), while the changes tree sorts with `localeCompare` and groups a directory's contents together. Untracked files also dumped at the end of "Changes" in the review instead of interleaving by path like the tree does. So `Beta.txt`/`ZETA_CAPS.txt` jumped to the top in the review but sat mid-list in the tree.

## Changes

- Extract the changes tree's build / compact / flatten logic into a shared pure helper in `@posthog/core/git-interaction/changeTree` (`buildChangeTree`, `compactChangeTree`, `flattenChangeTree`, plus `orderPathsLikeChangeTree` and a generic `sortByChangeTreeOrder`).
- `ChangesTreeView` now imports the shared tree builders instead of defining its own, so the tree and the review share one source of truth for ordering.
- The review list (local staged group, the merged unstaged+untracked "Changes" group, and the remote/branch/PR list) is now sorted to mirror the tree's leaf order by construction.

## How did you test this?

- New unit tests for the shared helper (14) covering case-insensitive sort, directory grouping, untracked interleave, and single-child compaction.
- New unit tests for the review ordering wrappers (6).
- Existing code-review (45) and task-detail (67) suites still green.

---

### DIFF
--- a/packages/core/src/git-interaction/changeTree.test.ts
+++ b/packages/core/src/git-interaction/changeTree.test.ts
@@ -1,0 +1,173 @@
+import type { ChangedFile } from "@posthog/shared/domain-types";
+import { describe, expect, it } from "vitest";
+import {
+  buildChangeTree,
+  compactChangeTree,
+  flattenChangeTree,
+  orderPathsLikeChangeTree,
+  sortByChangeTreeOrder,
+} from "./changeTree";
+
+const file = (path: string, over: Partial<ChangedFile> = {}): ChangedFile => ({
+  path,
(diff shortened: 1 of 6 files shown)


charlesvien:

Shout if you get stuck.

PR #3738, opened 2026-07-23
feat(loops): Add stop button to loop run rows

charlesvien:
## Problem

An in-progress loop run can only be stopped by opening its task detail page. The loop detail page shows run history and lets you view a run, but there is no way to stop a run or shut down its sandbox from there.

## Changes

Active cloud runs (queued or in progress) in the run history list now show a red "Stop run" button to the left of "View run". It reuses the sessions `StopCloudRunDialog`, so stopping goes through `sessionService.stopCloudRun` with the run's task and run ids, ends the cloud session and shuts down its sandbox. On success the runs list refetches immediately instead of waiting for the 15s poll. Terminal, local-environment and not-started runs show no button.

## How did you test this?

`pnpm --filter @posthog/ui typecheck` and Biome lint on the loops feature, both clean. Not exercised against a live in-progress loop run.

## Automatic notifications

- [ ] Publish to changelog?
- [ ] Alert Sales and Marketing teams?

PR #3739, opened 2026-07-23
feat(loops): Clean up the loop wizard

charlesvien:
## Problem

The create/edit loop wizard was cramped and confusing. Dropdowns were tiny, "Add trigger" always inserted a schedule you then had to re-type via a small select, trigger cards blended into the modal background, and the Context, Repository and Notifications sections didn't explain what they are, where output goes or how access works.

## Changes

- "Add trigger" is now a type picker menu (Schedule, GitHub event, API) with a description per type. Trigger cards get a header (icon, type, subtitle, enable switch, delete), labeled fields, specific inline validation, a darker background than the modal and an explicit empty state.
- GitHub event checkboxes are relabeled with a description of what fires them, matching the backend's action handling (any action on the subscribed event).
- Selects across the wizard go from 24px to 32px with 13px text via a new opt-in `size` prop on `SettingsOptionSelect` and `TimezonePicker`, plus wiring up the previously dead `size` prop on `GitHubRepoPicker`. Existing call sites keep the old size.
- Clearer copy: Context explains it's a channel from the sidebar (and the whole section hides when the channels alpha is off, unless the loop is already attached), Repository becomes "Base repository", Notifications state where each channel delivers and that full output lives on the loop page.

## How did you test this?

Drove the real Electron app over CDP through all four wizard steps: trigger type menu, all three card types, inline validation, empty state, and the channels toggle off/on hiding Context on both Options and Review. Ran `pnpm --filter @posthog/ui typecheck`, Biome lint on changed files and the full `@posthog/ui` Vitest suite (1977 passing).

## Automatic notifications

- [ ] Publish to changelog?

### DIFF
--- a/packages/ui/src/features/folder-picker/GitHubRepoPicker.tsx
+++ b/packages/ui/src/features/folder-picker/GitHubRepoPicker.tsx
@@ -43,6 +43,7 @@ export function GitHubRepoPicker({
   repositories,
   isLoading,
   placeholder = "Select repository...",
+  size = "1",
   disabled = false,
   anchor,
   showSearchInput = true,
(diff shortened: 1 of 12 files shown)


charlesvien:
* **#3739** <a href="https://app.graphite.com/github/pr/PostHog/code/3739?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/code/3739?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3738** <a href="https://app.graphite.com/github/pr/PostHog/code/3738?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #3772, opened 2026-07-23
feat: honest spend attribution and context-size lens in usage display

pauldambra:
## Problem

The usage banner's top-tool headline literally says "`<tool>` drives X% of your spend," computed from `share_of_scoped` in `packages/core/src/billing/spendSuggestions.ts`. That field is a co-occurrence artifact: a single turn can call several tools, and the backend attributes the full turn cost to *every* tool it touched, so whichever tool is most ubiquitous (Bash) looks artificially dominant, and the numbers don't reconcile to total spend.

This is the client half of a companion change to the posthog monorepo's `/api/llm_analytics/@me/spend/` endpoint, which adds honest attribution fields plus a genuinely new lever: spend broken down by input-context size.

## Changes

- **Types** (`packages/api-client/src/spend-analysis.ts` and `packages/core/src/billing/spendAnalysisTypes.ts`, kept in sync): added optional `cost_attributed_usd` / `share_attributed` to `SpendAnalysisToolRow` (fractional cost split across a turn's tools, reconciling to scoped spend), a new `SpendAnalysisInputSizeRow` type, and an optional `by_input_size?` breakdown on `SpendAnalysisResponse`.
- **Suggestions** (`packages/core/src/billing/spendSuggestions.ts`): the top-tool suggestion now uses `share_attributed` and is worded as co-occurrence-aware ("X was involved in N% of your spend when each turn's cost is split across the tools it used") instead of claiming a tool "drives" the spend. A new primary suggestion reads `by_input_size` and recommends trimming context (clearing between unrelated tasks, avoiding re-reads of large files/output) when the largest-context bucket or the top bucket by cost exceeds ~40% of scoped spend. The no-tool suggestion wording was tightened. All of this falls back to the previous behavior when the new fields are absent.
- **UI** (`packages/ui/src/features/usage/components/SpendBreakdownTables.tsx`, `SpendAnalysisSection.tsx`): the tool table gains an "Attributed" cost column when the field is present, and a new "By context size" table renders `by_input_size` when the backend serves it.

All new fields are optional on the client, exactly like the existing `by_day?` field, so this PR is backwards-compatible and can merge before or after the API rollout.

## How did you test this?

- `pnpm --filter @posthog/core typecheck`, `pnpm --filter @posthog/api-client typecheck`, `pnpm --filter @posthog/ui typecheck`, and a full workspace `pnpm typecheck` (24/24 tasks) all pass.
- `pnpm --filter @posthog/core test`: 2502/2502 pass, including a new `spendSuggestions.test.ts` (13 cases) covering the attributed-cost wording, the share_of_scoped/`share_attributed` fallback, the `by_input_size` threshold logic (including the `>300k`-under-threshold-but-another-bucket-over fallback), and the absent-field paths.
- `pnpm --filter @posthog/api-client test`: 130/130 pass.
- `pnpm --filter @posthog/ui test`: 2012/2012 pass.
- `biome check` on all changed files: clean.
- Full `pnpm test` (turbo, whole workspace): the only failures are in `@posthog/workspace-server` (git-network-integration and HogQL-fetch tests unrelated to this change, timing out in this sandbox with no diff in that package) — pre-existing and not touched by this PR.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/api-client/src/spend-analysis.ts
+++ b/packages/api-client/src/spend-analysis.ts
@@ -20,6 +20,9 @@ export interface SpendAnalysisToolRow {
   cost_usd: number;
   share_of_scoped: number;
   avg_input_tokens: number;
+  // Optional until the backend honest-attribution rollout reaches every deployment.
+  cost_attributed_usd?: number;
+  share_attributed?: number;
 }
 
 export interface SpendAnalysisModelRow {
@@ -36,6 +39,15 @@ export interface SpendAnalysisDayRow {
   cost_usd: number;
 }
 
+export interface SpendAnalysisInputSizeRow {
+  bucket: string;
+  min_input_tokens: number;
+  generation_count: number;
+  cost_usd: number;
+  share_of_scoped: number;
+  avg_cost_per_generation: number;
+}
+
 export interface SpendAnalysisBreakdown<TRow> {
   items: TRow[];
   truncated: boolean;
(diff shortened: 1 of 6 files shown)


charlesvien:
Hey @pauldambra, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3784, opened 2026-07-24
feat(channels): channel spaces and a structured task Activity view

puemos:
## Problem

Everything in Code hangs off one flat task list. There's no notion of *where* work lives, so a task, its canvases and its discussion have no shared home, and the task view has nowhere to put anything that isn't the agent transcript.

This reorganises the app around **channels**: the sidebar scopes to the channel you're in, and a task gets a structured Activity panel beside the transcript.

Behind `code-spaces-layout` (plus `project-bluebird`). Flag off is byte-for-byte the current experience — that's asserted by tests, not just by inspection.

## Changes

**Channel-scoped sidebar.** When you're in a channel the sidebar becomes that channel: a row of global destinations (Inbox / Activity / Command Center), a switcher that opens a searchable channel list, the channel's own sections, and its pinned + recent tasks and canvases with search and filters. Creating anything — ⌘N, the command menu, the FAB — files it into the channel you're in.

**Task Activity panel.** The thread panel gains Timeline / Artifacts / Comments. Timeline is the task's full history on a rail; Artifacts collects PRs, canvases, plans and the originating Slack thread across every run, with live PR state.

**Channel switching.** ⌘1–9 jump to starred channels, and a channel can be starred from the switcher or the channel header. Private channels show a lock instead of a hash, as in Slack.

**Under the UI.** The list-merging, run-status and run-artifact logic moved to `@posthog/core` as pure, tested functions. Three copies of the PR-artifact row and two of the thread-conversation logic were consolidated into `usePrArtifact` and `useThreadConversation`, so the new surfaces can't drift from the ones they were copied from.

The last commit is a review pass over the prototype: ten verified defects, the three worst being a cold load provisioning a backend channel literally named "channel", the flag unmounting the only global owner of Cmd+W, and a project switch leaving the next task filed into the previous project's channel.

## How did you test this?

- `pnpm --filter @posthog/ui test` — 2170 passing.
- `pnpm --filter @posthog/core test` — 2542 passing.
- `pnpm typecheck` — 24/24 packages clean.
- Biome clean across the repo.
- New tests cover the parts most likely to break quietly: the flag-off routing path for task creation, channel scoping and self-healing, the starred-channel chords, and the extracted core logic.
- Driven interactively in the running app, flag on and off.

## Automatic notifications

- [ ] Publish to changelog?

---


packages/ui/src/features/canvas/components/ChannelsSidebar.tsx
```diff
@@ -78,15 +84,71 @@ export function ChannelsSidebar() {
...
+  // alone. The latch resets when the layout turns off so a flag that resolves
+  // late (useFeatureFlag re-syncs on every flags payload) can't strand the
+  // sidebar unscoped for the rest of the session.
+  const autoScopedRef = useRef(false);
+  useEffect(() => {
+    if (!channelsLayout) {
+      autoScopedRef.current = false;
+      return;
+    }
+    if (autoScopedRef.current || currentChannelId) return;
```
puemos: [P1] Preserve the route channel before applying the #me default

On a cold load of `/website/<non-me-channel>`, `currentChannelId` is null when both effects are scheduled. The route effect above first stores the URL's channel, but this later effect closes over the same null value and immediately overwrites it with `#me`. On the next render the route effect's dependencies are unchanged, so it does not restore the URL channel; the sidebar remains scoped to `#me`, and global task creation can consequently file into `#me` while the page still shows the other channel. Skip auto-scoping when `routeChannelId` is present (or combine the two decisions so the route wins).

packages/ui/src/features/canvas/components/TaskArtifactsList.tsx
```diff
@@ -0,0 +1,335 @@
...
+          size={14}
+          weight="bold"
+          className="shrink-0"
+          style={{ color: iconColor }}
+        />
+      }
+      title={title}
+      detail={detailParts.join(" · ") || null}
+      onHoverStart={() => setCountsWanted(true)}
+      onOpen={() => setReviewMode(taskId, "split")}
```
puemos: [P2] Open the PR represented by this row

This list intentionally contains PRs from every run, but the click discards `url` and only opens review mode for the task. `CloudReviewPage` then derives its source from `resolveCloudPrUrl(...)` (the task's primary PR), so clicking any non-primary/historical PR row opens the primary PR's diff instead of the row the user selected. Pass the selected URL into review navigation, or open `safeUrl` externally when the review surface cannot select a specific PR.

puemos (commented):
Validated the findings below against the current head and traced each through its call sites.

packages/ui/src/features/canvas/components/ChannelsSidebar.tsx
puemos: Fixed in 2c0c0c5: route-scoped channels now take precedence over the #me fallback, with cold deep-link regression coverage.

packages/ui/src/features/canvas/components/TaskArtifactsList.tsx
puemos: Fixed in 2c0c0c5: artifact rows now carry the selected validated PR URL into the review source, with multi-run interaction coverage.

puemos:
Superseded by the reviewable four-PR stack:

1. [#3812](https://github.com/PostHog/code/pull/3812) — shared thread and PR artifact refactors
2. [#3813](https://github.com/PostHog/code/pull/3813) — dormant channel scope foundations
3. [#3814](https://github.com/PostHog/code/pull/3814) — complete flag-gated spaces UI
4. [#3815](https://github.com/PostHog/code/pull/3815) — structured Activity view

The stack tip is byte-for-byte equivalent to this branch, and every stack commit is GitHub-verified.

PR #3795, opened 2026-07-24
fix(message-editor): preserve comment anchors in pasted GitHub links

cvolzer3:
## Problem

Pasting a GitHub comment link into the composer — e.g.

```
https://github.com/PostHog/posthog/pull/72409#discussion_r3647131256
```

auto-converts it into a PR chip, but `parseGithubIssueUrl` rebuilt the chip URL from owner/repo/number only, silently dropping the `#discussion_r…` anchor. The chip's click-through and the `url="…"` attribute sent to the agent both ended up pointing at the PR itself instead of the specific comment.

## Fix

`parseGithubIssueUrl` now keeps the URL suffix on `normalizedUrl` whenever it contains a non-empty fragment. The whole suffix is preserved (path tail + query + fragment), not just the fragment, because anchors like `#r123` only resolve on the `/files` subpage they were copied from — `/pull/123#r123` would land nowhere.

Behavior without a fragment is unchanged: tab suffixes (`/files`), query strings, and trailing junk are still stripped, and `http://` is still normalized to `https://`.

`apps/mobile/src/lib/githubIssueUrl.ts` is a verbatim copy of the helper (used for PR badges and markdown link chips) and gets the identical fix to stay in sync.

## Testing

- Flipped the old "fragment is stripped" unit test to assert preservation, and added cases for the reported `#discussion_r…` link, a `/files#r…` link, and an empty `#` (still stripped).
- `pnpm --filter @posthog/core exec vitest run src/message-editor/githubIssueUrl.test.ts src/message-editor/content.test.ts` — 42 tests pass.
- `pnpm typecheck` and Biome pass (also enforced by the pre-commit hook).

### DIFF
--- a/apps/mobile/src/lib/githubIssueUrl.ts
+++ b/apps/mobile/src/lib/githubIssueUrl.ts
@@ -9,23 +9,28 @@ export interface ParsedGithubIssueUrl {
 }
 
 const GITHUB_ISSUE_URL_PATTERN =
-  /^https?:\/\/github\.com\/([^/\s]+)\/([^/\s]+)\/(issues|pull)\/(\d+)(?:[/?#].*)?$/;
+  /^https?:\/\/github\.com\/([^/\s]+)\/([^/\s]+)\/(issues|pull)\/(\d+)([/?#].*)?$/;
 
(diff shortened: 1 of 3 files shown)


tatoalo (approved):
Thanks!

PR #3798, opened 2026-07-24
Add headless CLI for one-shot agent runs

haacked:
## Problem

There's no way to run the agent from a script. The desktop app needs a GUI, and cloud tasks need a sandbox; anyone who wants "prompt in, answer on stdout" (CI checks, shell pipelines, cron jobs) has to shell out to a different tool entirely.

## Changes

Adds `packages/cli` (`@posthog/code-cli`, bin `posthog-code-cli`), a headless CLI host that drives the same in-process ACP connection the desktop app uses via `Agent` from `@posthog/agent`:

```
posthog-code-cli "Fix the failing test" --cwd ~/src/myrepo --permission-mode auto
```

- Prompt as a positional arg or piped on stdin; assistant text streams to stdout, diagnostics go to stderr, exit code reflects the stop reason (0 end_turn, 2 refusal, 3 token/turn limits, 130 SIGINT).
- `--permission-mode auto` (default) or `bypassPermissions`. Interactive modes are rejected since there's no UI to answer prompts. Tool permissions that still reach the client are auto-approved (preferring `allow_once` so one-shot runs don't persist allow rules); `AskUserQuestion` calls are parked with the same no-user-available contract cloud background runs use, so the model restates the question as text and ends its turn.
- `--output json` buffers and emits one `{ text, stopReason, usage, sessionId }` document for machine consumption.
- Auth comes from `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` env vars, falling back to a stored claude login credential.

`ClaudeAcpAgent` previously hardcoded a debug console logger, ignoring the logger injected through `createAcpConnection`; it now accepts the injected logger (keeping the console fallback), which lets the CLI route adapter diagnostics off stdout behind `--debug`. Desktop and cloud behavior is unchanged since they don't pass one.

The new package lands in `packages/` rather than the `apps/cli` slot listed in AGENTS.md because repo settings deny writes there; the AGENTS.md hosts table now points at `packages/cli`.

## How did you test this?

- 39 unit tests across arg parsing, the unattended permission policy, and the output sink (`pnpm --filter @posthog/code-cli test`).
- Full `@posthog/agent` suite after the logger change (1625 passed; one pre-existing flake in `settings.test.ts` that passes in isolation).
- Live smoke runs against a scratch repo: read-only question, auto-mode file edit applied without prompting, `--output json` shape, clean stdout (zero stderr bytes without `--debug`, 15 diagnostic lines with it), and exit codes for bad flags/missing prompt.
- Opt-in e2e spec (`pnpm --filter @posthog/code-cli test:e2e`) using the same gateway env contract as `packages/agent/e2e`.

## Automatic notifications

- [ ] Publish to changelog?

---


charlesvien:
Remade in the monorepo as PostHog/posthog#76470 (PostHog/code main is frozen after the desktop import).

PR #3807, opened 2026-07-25
perf: reduce task-time CPU and process leaks

richardsolomou:
## Problem

Running tasks can keep the renderer, GPU, filesystem watcher, and orphaned child processes busy enough to heat the laptop. Live sampling found per-event transcript work and two task servers that survived for over a day at roughly 100% CPU each.

## Changes

I batch streamed `session/update` notifications at 50ms while keeping completion, permission, and runtime state events immediate. Long transcripts derive incrementally: each event is parsed once, and a content change (tool merge, streamed subagent output, progress step, thought settling, or turn completion) re-issues just that row object. Memoized rows re-render exactly where something happened instead of cloning the whole active turn per frame. The chat-thread row grouper reuses cached rows across frames and identity-checks its retained prefix, including late completion of an older turn.

Diff workers are capped at two, persistent timers and watcher invalidations run less often, and narrower subscriptions cut re-renders. Changed-file derivation has its own revision, so partial tool input does not clone and rescan the full tool map. Process cleanup indexes each process snapshot once, terminates nested descendant groups in batches, isolates Windows failures per PID, and revalidates pid, pgid, and start time before delayed signals. It falls back to the tracked process group when a snapshot cannot recover an orphaned tree. Conversation refreshes drop pending batches atomically so events cannot be appended twice during initial load or later refreshes.

## How did you test this?

- Ran the full core (2,512), UI (2,193), and workspace-server (794) suites plus package typechecks and Biome checks after rebasing onto main's chat-thread virtualization.
- Re-ran 99 focused core, UI, and workspace-server tests after review fixes, plus all three package typechecks, the full workspace typecheck commit hook, Biome, and the host-boundary checker.
- Added regressions for chunks arriving during initial load, interleaved text and tool batching, late completion of an older turn, immediate elapsed-time rendering, changed-file revision stability, orphaned process-group fallback, and delayed process identity validation.
- Benchmarks measured against `bf27356b8` (main at the time of measurement; none of the measured code paths changed on main since). A deterministic synthetic 10,000-event transcript uses 9,000 history events and streams 1,000 events through each side's real derivation and grouping pipeline, one frame per event. The 50ms batching is excluded, so these streaming numbers are the PR's worst case. Times are medians of three runs. Process rows spawn a real six-PID tree across three process groups and poll liveness.

| Benchmark | Base | This PR | Change |
| --- | --- | --- | --- |
| Streaming derivation + row grouping (1,000 frames) | 138.5 ms | 19.6 ms | 7.1× faster |
| Initial 9,000-event history build | 1.8 ms | 2.6 ms | comparable |
| Row identities changed per frame | 334 | 1 | 334× fewer |
| React re-render time across the streamed window | 4.25 s | 1.06 s | 4.0× faster |
| Nested process-tree kill | 5 of 6 processes leak | all 6 dead in 32 ms | leak fixed |

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/pi-runtime/piSessionController.ts
+++ b/packages/core/src/pi-runtime/piSessionController.ts
@@ -80,6 +80,8 @@ export type PiSessionProvider = PiSessionFactory;
 
 export type PiSubmitResult = "prompt" | "steer" | "followUp" | "compact";
 
+const STREAM_UPDATE_INTERVAL_MS = 50;
+
 type PiOperation =
   | "prompt"
   | "compact"
@@ -122,6 +124,11 @@ export class PiSessionController {
   private readonly sessions = new Map<string, Promise<PiSession>>();
   private readonly subscriptions = new Map<string, () => void>();
   private readonly liveEvents = new Map<string, AgentConversationEvent[]>();
+  private readonly pendingEvents = new Map<string, AgentConversationEvent[]>();
+  private readonly pendingEventTimers = new Map<
+    string,
+    ReturnType<typeof setTimeout>
+  >();
   private readonly connections = new Map<string, Promise<void>>();
   private readonly readiness = new Map<string, Promise<void>>();
   private readonly sessionVersions = new Map<string, number>();
@@ -201,6 +208,7 @@ export class PiSessionController {
 
   disconnect(taskId: string): void {
     this.cancelAuthRestoration.get(taskId)?.();
+    this.flushPendingEvents(taskId);
     this.resetTransport(taskId);
     this.taskRunIds.delete(taskId);
     this.liveEvents.delete(taskId);
@@ -600,11 +608,15 @@ export class PiSessionController {
       const conversationEvents = events.filter(
         (event) => event.type !== "queue_update",
       );
-      const liveEvents = this.liveEvents.get(taskId) ?? [];
+      const liveEvents = [
+        ...(this.liveEvents.get(taskId) ?? []),
+        ...(this.pendingEvents.get(taskId) ?? []),
+      ];
       const newLiveEvents = this.reconcileLiveEvents(
         conversationEvents,
         liveEvents,
       );
+      this.discardPendingEvents(taskId);
       this.liveEvents.set(taskId, newLiveEvents);
       const historyUserMessageIds = new Set(
         conversationEvents.flatMap((event) =>
@@ -701,6 +713,16 @@ export class PiSessionController {
   }
 
   private handleEvent(taskId: string, event: AgentConversationEvent): void {
+    if (
+      event.type === "assistant_message_chunk" ||
+      event.type === "assistant_thought_chunk" ||
+      event.type === "tool_call_updated"
+    ) {
+      this.queueEvent(taskId, event);
+      return;
+    }
+
+    this.flushPendingEvents(taskId);
     if (event.type === "queue_update") {
       const queue = {
         steering: event.steering,
@@ -742,16 +764,9 @@ export class PiSessionController {
         );
       }
     }
-    const isDirectBashEvent =
-      (event.type === "tool_call_started" ||
-        event.type === "tool_call_updated") &&
-      event.toolCall.origin === "user_shell";
     const hasTurnActivity =
-      !isDirectBashEvent &&
-      (event.type === "assistant_message_chunk" ||
-        event.type === "assistant_thought_chunk" ||
-        event.type === "tool_call_started" ||
-        event.type === "tool_call_updated");
+      event.type === "tool_call_started" &&
+      event.toolCall.origin !== "user_shell";
     if (status && hasTurnActivity) {
       status = { ...status, isStreaming: true };
     }
@@ -794,6 +809,78 @@ export class PiSessionController {
     if (event.type === "turn_completed") {
       void this.refreshStats(taskId);
     }
+  }
+
+  private queueEvent(taskId: string, event: AgentConversationEvent): void {
+    const pending = this.pendingEvents.get(taskId) ?? [];
+    pending.push(event);
+    this.pendingEvents.set(taskId, pending);
+
+    if (this.pendingEventTimers.has(taskId)) return;
+    this.pendingEventTimers.set(
+      taskId,
+      setTimeout(() => {
+        this.pendingEventTimers.delete(taskId);
+        this.flushPendingEvents(taskId);
+      }, STREAM_UPDATE_INTERVAL_MS),
+    );
+  }
+
+  private flushPendingEvents(taskId: string): void {
+    const timer = this.pendingEventTimers.get(taskId);
+    if (timer) {
+      clearTimeout(timer);
+      this.pendingEventTimers.delete(taskId);
+    }
+
+    const pending = this.pendingEvents.get(taskId);
+    if (!pending?.length) return;
+    this.pendingEvents.delete(taskId);
+
+    const session = this.getSession(taskId);
+    const seenSourceIds = new Set(
+      session.events.flatMap((event) =>
+        event.sourceId ? [event.sourceId] : [],
+      ),
+    );
+    const events = pending.filter((event) => {
+      if (!event.sourceId || !seenSourceIds.has(event.sourceId)) {
+        if (event.sourceId) seenSourceIds.add(event.sourceId);
+        return true;
+      }
+      return false;
+    });
+    if (events.length === 0) return;
+
+    this.liveEvents.set(taskId, [
+      ...(this.liveEvents.get(taskId) ?? []),
+      ...events,
+    ]);
+    const hasTurnActivity = events.some(
+      (event) =>
+        event.type !== "tool_call_updated" ||
+        event.toolCall.origin !== "user_shell",
+    );
+    const latestSession = this.getSession(taskId);
+    this.updateSession(taskId, {
+      connectionState: "connected",
+      events: [...latestSession.events, ...events],
+      status:
+        latestSession.status && hasTurnActivity
+          ? { ...latestSession.status, isStreaming: true }
+          : latestSession.status,
+      error:
+        latestSession.error?.scope === "operation"
+          ? latestSession.error
+          : undefined,
+    });
+  }
+
+  private discardPendingEvents(taskId: string): void {
+    const timer = this.pendingEventTimers.get(taskId);
+    if (timer) clearTimeout(timer);
+    this.pendingEventTimers.delete(taskId);
+    this.pendingEvents.delete(taskId);
   }
 
   private async refreshStats(taskId: string): Promise<void> {
@@ -1188,6 +1275,7 @@ export class PiSessionController {
   }
 
   private resetTransport(taskId: string): void {
+    this.discardPendingEvents(taskId);
     this.advanceSessionVersion(taskId);
     this.subscriptions.get(taskId)?.();
     this.subscriptions.delete(taskId);
--- a/packages/core/src/sessions/sessionService.ts
+++ b/packages/core/src/sessions/sessionService.ts
@@ -124,10 +124,10 @@ const MAX_RESPONDED_PERMISSION_REQUEST_IDS = 500;
  * Streamed events are buffered and flushed on this cadence so a burst of tokens
  * coalesces into one processing pass (and roughly one render) instead of one
  * per event. Electron IPC delivers each event as its own task, so a microtask
- * flush wouldn't batch across them — a short timer does. One frame is
- * imperceptible for streamed text.
+ * flush wouldn't batch across them — a short timer does. A 50ms presentation
+ * delay is imperceptible for streamed text and cuts update churn substantially.
  */
-const SESSION_EVENT_FLUSH_MS = 16;
+const SESSION_EVENT_FLUSH_MS = 50;
 /**
  * A backgrounded session's transcript is freed this long after it stops being
  * viewed, and reloaded from disk on return. Only disconnected (idle, no live
@@ -1537,6 +1537,18 @@ function classifyTurnEventKind(
   return "other";
 }
 
+function isPassiveStreamEvent(acpMsg: AcpMessage): boolean {
+  const msg = acpMsg.message;
+  if (!("method" in msg) || msg.method !== "session/update") return false;
+  const sessionUpdate = (
+    msg as { params?: { update?: { sessionUpdate?: string } } }
+  ).params?.update?.sessionUpdate;
+  return (
+    sessionUpdate === "agent_message_chunk" ||
+    sessionUpdate === "agent_thought_chunk"
+  );
+}
+
 export class SessionService {
   private connectingTasks = new Map<string, Promise<void>>();
   private reconcilingTasks = new Set<string>();
@@ -2593,9 +2605,7 @@ export class SessionService {
     const batches = this.pendingSessionEvents;
     this.pendingSessionEvents = new Map();
     for (const [taskRunId, events] of batches) {
-      for (const acpMsg of events) {
-        this.handleSessionEvent(taskRunId, acpMsg);
-      }
+      this.handleSessionEventBatch(taskRunId, events);
     }
   }
 
@@ -2605,9 +2615,36 @@ export class SessionService {
     const events = this.pendingSessionEvents.get(taskRunId);
     if (!events) return;
     this.pendingSessionEvents.delete(taskRunId);
-    for (const acpMsg of events) {
-      this.handleSessionEvent(taskRunId, acpMsg);
+    this.handleSessionEventBatch(taskRunId, events);
+  }
+
+  private handleSessionEventBatch(
+    taskRunId: string,
+    events: AcpMessage[],
+  ): void {
+    let passive: AcpMessage[] = [];
+    const flushPassive = () => {
+      if (passive.length === 0) return;
+      const session = this.d.store.getSessions()[taskRunId];
+      if (session) {
+        if (session.initialPrompt?.length) {
+          this.d.store.updateSession(taskRunId, { initialPrompt: undefined });
+        }
+        this.d.store.appendEvents(taskRunId, passive);
+        this.updatePromptStateFromEvents(taskRunId, passive, { isLive: true });
+      }
+      passive = [];
+    };
+
+    for (const event of events) {
+      if (isPassiveStreamEvent(event)) {
+        passive.push(event);
+      } else {
+        flushPassive();
+        this.handleSessionEvent(taskRunId, event);
+      }
     }
+    flushPassive();
   }
 
   // --- Transcript residency (memory eviction) ---
@@ -2713,7 +2750,13 @@ export class SessionService {
       { taskRunId },
       {
         onData: (payload: unknown) => {
-          this.enqueueSessionEvent(taskRunId, payload as AcpMessage);
+          const event = payload as AcpMessage;
+          if (isPassiveStreamEvent(event)) {
+            this.enqueueSessionEvent(taskRunId, event);
+          } else {
+            this.flushSessionEventsForTask(taskRunId);
+            this.handleSessionEvent(taskRunId, event);
+          }
         },
         onError: (err) => {
           this.d.log.error("Session subscription error", {
--- a/packages/core/src/sessions/sessionService.ts
+++ b/packages/core/src/sessions/sessionService.ts
@@ -1539,7 +1539,10 @@ function classifyTurnEventKind(
 
 function isStreamUpdateEvent(acpMsg: AcpMessage): boolean {
   const msg = acpMsg.message;
-  return "method" in msg && msg.method === "session/update";
+  if (!("method" in msg) || msg.method !== "session/update") return false;
+  const update = (msg as { params?: { update?: { sessionUpdate?: string } } })
+    .params?.update;
+  return update?.sessionUpdate !== "config_option_update";
 }
 
 export class SessionService {
--- a/packages/core/src/sessions/sessionService.ts
+++ b/packages/core/src/sessions/sessionService.ts
@@ -1537,16 +1537,9 @@ function classifyTurnEventKind(
   return "other";
 }
 
-function isPassiveStreamEvent(acpMsg: AcpMessage): boolean {
+function isStreamUpdateEvent(acpMsg: AcpMessage): boolean {
   const msg = acpMsg.message;
-  if (!("method" in msg) || msg.method !== "session/update") return false;
-  const sessionUpdate = (
-    msg as { params?: { update?: { sessionUpdate?: string } } }
-  ).params?.update?.sessionUpdate;
-  return (
-    sessionUpdate === "agent_message_chunk" ||
-    sessionUpdate === "agent_thought_chunk"
-  );
+  return "method" in msg && msg.method === "session/update";
 }
 
 export class SessionService {
@@ -2637,7 +2630,7 @@ export class SessionService {
     };
 
     for (const event of events) {
-      if (isPassiveStreamEvent(event)) {
+      if (isStreamUpdateEvent(event)) {
         passive.push(event);
       } else {
         flushPassive();
@@ -2751,7 +2744,7 @@ export class SessionService {
       {
         onData: (payload: unknown) => {
           const event = payload as AcpMessage;
-          if (isPassiveStreamEvent(event)) {
+          if (isStreamUpdateEvent(event)) {
             this.enqueueSessionEvent(taskRunId, event);
           } else {
             this.flushSessionEventsForTask(taskRunId);
--- a/packages/core/src/sidebar/buildSidebarData.ts
+++ b/packages/core/src/sidebar/buildSidebarData.ts
@@ -1,4 +1,8 @@
-import { readPrUrls, type WorkspaceMode } from "@posthog/shared";
+import {
+  readPrUrls,
+  type SessionStatus,
+  type WorkspaceMode,
+} from "@posthog/shared";
 import type { Task, TaskRunStatus } from "@posthog/shared/domain-types";
 import { getRepositoryInfo } from "./groupTasks";
 import type { TaskData, TaskGroup } from "./sidebarData.types";
@@ -83,7 +87,7 @@ export function filterVisibleTasks(
 }
 
 export interface TaskSession {
-  status?: string;
+  status?: SessionStatus;
   isPromptPending?: boolean;
   pendingPermissions?: { size: number };
   cloudStatus?: TaskRunStatus;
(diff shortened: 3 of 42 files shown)


packages/ui/src/features/sessions/components/chat-thread/chatRowGrouping.ts
```diff
@@ -0,0 +1,138 @@
...
+      const boundaryId = items[rebuildStart]?.id;
+      const cachedBoundaryIndex = boundaryId
+        ? cachedRows.findIndex((row) => row.id === boundaryId)
+        : -1;
+      const prefixRowCount =
+        rebuildStart === 0
+          ? 0
+          : cachedBoundaryIndex >= 0
+            ? cachedBoundaryIndex
+            : cachedRows.length;
```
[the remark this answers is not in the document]
richardsolomou: done, replacements inside the cached range now force a regroup when the confirmed boundary id changes. added a regression test for the optimistic prompt case.

packages/workspace-server/src/services/process-tracking/process-utils.ts
```diff
@@ -14,28 +76,24 @@ export function killProcessTree(pid: number): void {
...
-      // orphaned children in the same group would be missed. The catch blocks
-      // handle ESRCH if everything already exited.
       setTimeout(() => {
-        for (const target of [-pid, pid]) {
-          try {
-            process.kill(target, "SIGKILL");
-          } catch {}
-        }
+        signalTargets(targets, "SIGKILL");
       }, SIGKILL_GRACE_MS).unref();
```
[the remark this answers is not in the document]
richardsolomou: done, the delayed kill now takes a fresh snapshot and matches pid, pgid, and process start time before signalling. cleanup is also batched across tracked processes.

packages/workspace-server/src/services/process-tracking/process-utils.ts
```diff
@@ -1,46 +1,171 @@
...
+  ownPgid: number | undefined,
+  deps: UnixProcessKillerDeps,
+): void {
+  const trees = rootPids
+    .map((pid) => findProcessTree(pid, initialProcesses))
+    .filter((tree) => tree.length > 0);
+  const originalTree = Array.from(
+    new Map(trees.flat().map((entry) => [entry.pid, entry])).values(),
+  );
+  if (originalTree.length === 0) return;
```
k11kirky: **The Unix kill is now gated on the `ps` snapshot, so it sends no signal at all in exactly the orphaned-group case this PR is fixing.**

`if (originalTree.length === 0) return;` guards both the SIGTERM below and the deferred SIGKILL. `findProcessTree` only reaches entries linked by `ppid` from the root, plus the root's own row — so a tracked leader that has already exited (children reparented to pid 1) yields `[]`, and its surviving group is never signalled.

`snapshotUnixProcesses` also returns `[]` from its catch when `ps -axo pid=,ppid=,pgid=,lstart=` is unavailable or unparseable (busybox / a minimal sandbox image), which makes every `kill` / `killByTaskId` / `@preDestroy killAll` a complete no-op.

Pre-PR this was unconditional:

```js
for (const target of [-pid, pid]) {
  try { process.kill(target, "SIGTERM"); sent = true; break; } catch {}
}
```

which reaped the group in both cases. The new test `does not signal raw ids when the root identity is unavailable` locks the regression in. Suggest keeping a group-kill fallback when the snapshot yields nothing for a root pid.

packages/workspace-server/src/services/process-tracking/process-utils.ts
```diff
@@ -1,46 +1,171 @@
...
- * Kill a process and all its children by killing the process group.
- * On Unix, we use process.kill(-pid) to kill the entire process group.
+ * Kill a process and all its descendants, including children that created
+ * their own process groups.
  * On Windows, we use taskkill with /T flag to kill the process tree.
  */
-export function killProcessTree(pid: number): void {
+export function killProcessTrees(pids: readonly number[]): void {
+  if (pids.length === 0) return;
   try {
```
k11kirky: **Windows: the per-pid `try/catch` became one `try` around the loop, so the first failing pid aborts the kill for every remaining pid.**

The only `catch {}` is at line 162, outside the `for`. An already-exited pid makes `taskkill` exit 128, `execSync` throws, the loop unwinds, and pids 2..N are never signalled — yet `killProcesses` still calls `unregister(pid, "killed")` for all of them, so the tracker reports them dead and nothing retries.

Pre-PR each pid went through its own `killProcessTree` call with its own try/catch, so a stale pid couldn't block its siblings. Moving the `try` inside the loop restores that.

k11kirky (commented):
## Review

Broad task-time performance PR: 50ms batch window in three places, incremental conversation-item derivation with row-identity swaps, an extracted `chatRowGrouping.ts` cache, throttled UI timers and watcher flushes, and rewritten process-tree termination with `ps`-snapshot identity revalidation. The measurements look credible for the paths they cover, but there are a few places where the mechanism doesn't hold — including two regressions in the process-kill code this PR exists to fix, and one path where batching was *removed* rather than widened.

12 inline comments below, most severe first.

### Altitude / duplication (no single line to anchor to)

- **Batching is now hand-rolled four times** with four separate constants: `cloud-task.ts` (50ms + `EVENT_BATCH_MAX_SIZE`), `sessionService.ts` (50ms, one global timer + per-key map), `piSessionController.ts` (50ms, one timer per task), and `apps/mobile/src/features/tasks/lib/cloudTaskStream.ts` — still at 16ms, untouched by this PR. They already differ in guarantees (only cloud-task caps batch size; only sessionService drains per key on demand; only the pi controller has to hand-clear its timer to avoid a double-append). A shared `createEventBatcher(flushMs)` would make retuning the cadence one edit and one test suite.
- **`chatRowGrouping.ts` is a second incremental prefix-cache** alongside the live `createIncrementalThreadGrouper` (`new-thread/incrementalThreadGrouping.ts`, used by `ConversationView`) over the same `ConversationItem[]`. Same five steps (input fast-out → pick a cut → identity-verify the prefix → boundary id → cached row index → rebuild suffix and concat), different helpers. The existing one carries `findStablePrefixItemCount` and `groupBoundaryAtOrBefore`; the new one compensates with a full O(n) pointer scan.
- **`mergeConversationItems` may defeat the new cache on cloud runs** (didn't confirm, needs ≥2 user messages + an echo that differs from the pinned seed — personalization folded in, or attachments). When it re-spreads the pinned optimistic item, `items[0]` is fresh every frame, the i=0 identity compare fails, and the grouper regroups the entire thread per frame. Worth a check with a real cloud task.

### Checked and refuted, in case they come up

- The `b.items[existing.index] === existing.item` guard on shell-execute results **cannot fail** — the only `splice` back-walks over trailing `progress_group` rows only, so it always inserts above any `user_shell_execute` row. No re-basing needed (which is why only `progressCards` need it).
- The watcher's `bulkChanged` latch has **identical output** to the old `totalChanges <= BULK_THRESHOLD` guard; the per-path event drop for >100-path windows predates this PR.
- `MAX_WAIT_MS` 1000→3000 is a real 3s worst-case staleness for the diff panel during sustained writes, but consistent with the PR's intent — flagging as a deliberate trade-off, not a defect.
- `flushPassive` **is** a faithful subset of `handleSessionEvent` for the two chunk kinds — I checked specifically for dropped activity timestamps, unread markers, notifications, title generation and residency bookkeeping, and every branch not replicated is guarded by a condition a chunk notification can't satisfy.

*Review by [PostHog Code](https://posthog.com/code?ref=pr)*

packages/core/src/pi-runtime/piSessionController.ts
```diff
@@ -310,6 +331,38 @@ export class PiSessionController {
...
+      this.pendingEventTimers.delete(taskId);
+    }
+
+    const pending = this.pendingEvents.get(taskId);
+    if (!pending?.length) return;
+    this.pendingEvents.delete(taskId);
+
+    const session = this.getSession(taskId);
+    this.updateSession(taskId, {
+      events: [...session.events, ...pending],
```
k11kirky: **`loadSession` doesn't drop the pending buffer, so a chunk that arrives during the conversation fetch is appended twice — here.**

`connect()` installs the subscription synchronously (`ensureSubscription`) *before* awaiting `Promise.all([client.conversation, client.status])`. A chunk arriving in that window goes into `liveEvents` **and** into `pendingEvents` with a 50ms timer. `loadSession` then writes `events: [...events, ...liveEvents]` — chunk present once — and 50ms later this still-armed timer appends the same chunk again, duplicating the streamed text.

`refreshConversation` clears the pending buffer and timer for exactly this reason, with a comment saying so:

```ts
// Pending events are a subset of liveEvents, so each is either in the fetched
// conversation or in remainingEvents; a timer firing after this replacement
// would append it a second time.
```

`loadSession` performs the same transcript replacement and omits the guard. Extracting a `discardPending(taskId)` helper and calling it from both would close it — and there's no test on this path (the existing one only covers `refreshConversation`).

packages/ui/src/features/sessions/components/chat-thread/chatRowGrouping.ts
```diff
@@ -0,0 +1,144 @@
...
+        if (isUserInitiatedConversationItem(items[index])) {
+          rebuildStart = index;
+          break;
+        }
+      }
+
+      // The builder replaces rows in place anywhere in the list (a status
+      // completing, a shell result arriving, a thought settling), so every
+      // retained item must be identity-checked — pointer compares, cheap.
+      for (let i = 0; i < rebuildStart; i++) {
```
k11kirky, line 108: **The retained prefix is validated by item identity only, but turn completion mutates the shared `turnContext` in place — so a turn frozen into the prefix keeps rendering as streaming after it finishes.**

Cloud steer mid-turn appends a tail optimistic `user_message` (`sessionService.ts:3531` → `appendOptimisticItem({ pinToTop: false })`), which is user-initiated, so `rebuildStart` lands on it and the still-streaming turn's rows fall into the retained prefix. `completePromptTurn` then does `turn.context.turnComplete = true` in place (`buildConversationItems.ts:594-597`) without replacing any item — this loop sees no change, `cachedRows.slice(0, prefixRowCount)` returns the same `TurnRow` objects, and `ThreadRow` (memo, shallow-compared on the row object) bails.

The flags are read *inside* that skipped render — `isStreaming={!item.turnContext.turnComplete}` (`ChatThread.tsx:990`), `turnStreaming` (`:422`), `turnComplete` (`:1000`) — so the turn keeps showing streaming markdown, a spinning "Using …" group and no turn timestamp until an unrelated item append forces a full rebuild.

This is the contract stated in `buildConversationItems.ts:214`: *"Replace, never mutate: memoized rows (and the chat-thread row cache) only see the flip if the item's identity changes"* — only `thoughtComplete` currently gets that treatment. Either give `turnComplete`/`turnCancelled` the same replace-don't-mutate handling, or have the prefix check compare turn flags too.

(The virtualized path only partially escapes: `FlatRowView`'s comparator includes `row.turnTimestamp`, so the turn's *trailing* row re-renders but non-trailing rows stay stale.)

packages/core/src/sessions/sessionService.ts
```diff
@@ -2661,7 +2698,13 @@ export class SessionService {
       { taskRunId },
       {
         onData: (payload: unknown) => {
-          this.enqueueSessionEvent(taskRunId, payload as AcpMessage);
+          const event = payload as AcpMessage;
+          if (isPassiveStreamEvent(event)) {
```
k11kirky: **Narrowing batching to the two chunk kinds removes coalescing that previously covered `tool_call_update` — the highest-volume stream event — so tool-heavy turns regress versus pre-PR.**

`onData` used to enqueue *every* event into the 16ms batcher. It now enqueues only if `isPassiveStreamEvent(event)` and otherwise calls `flushSessionEventsForTask` + `handleSessionEvent` synchronously.

`inputJsonDeltaToAcpNotifications` emits one `tool_call_update` per `input_json_delta` (`sdk-to-acp.ts:689`), and `sessionEvents.ts:248` documents *"thousands of near-identical growing snapshots… one 9k-update tool run"* — which is why `collapseSupersededToolCallUpdates` exists. So a tool-heavy turn now does one store commit and one render **per event**, with zero coalescing where the old 16ms timer batched them all.

Interleaved prose + tool traffic is worse: each `tool_call_update` drains the 1–2 buffered chunks, fragmenting text into many tiny `appendEvents` calls. The batching test only passes because it uses a pure-text burst.

`PiSessionController.handleEvent` gets this right 1200 lines away — it queues `tool_call_updated`. Suggest inverting the predicate: batch every `session/update` and keep only the genuinely stateful/terminal messages (responses, permission requests, `runtime_status`, `_posthog/*`) on the immediate path, so a newly added streaming update kind defaults to batched instead of silently regressing.

packages/ui/src/features/sessions/components/incrementalConversationItems.ts
```diff
@@ -141,10 +143,18 @@ export function createIncrementalConversationBuilder() {
       turn.context.turnComplete = true;
     }
 
-    markThoughtCompletion(builder.items);
+    markThoughtCompletion(
```
k11kirky: **Bounding the scan at `min(thoughtScanStart, currentTurnStartIndex)` skips an older turn that completes late, leaving its thought row spinning until the next idle rebuild.**

Steer/cancel then immediately send (`sessionService.ts:3531-3551`; `cancelPrompt` doesn't await the agent's cancelled response) lets turn A's JSON-RPC response land after turn B has started — a case the codebase explicitly handles elsewhere (*"A late response from a previously cancelled turn must not…"*, `sessionService.ts:2911`).

`handlePromptRequest` advances `currentTurnStartIndex` past A's rows and nothing lowers it on completion, so when `completePromptTurn(A)` fires, `markThoughtCompletion` starts *above* A's trailing `agent_thought_chunk` row. `thoughtComplete` stays `false`, the row identity never changes, and `<ThoughtView isLoading={!thoughtComplete} />` (`SessionUpdateView.tsx:94`) spins forever — the turn flags can't rescue it because the spinner reads the row field, not `turnContext`. Only `finalizeBuilder`'s index-0 rescan clears it.

Tracking the lowest index touched by any completed turn in the batch (rather than assuming completions only affect the active turn) would cover it.

packages/core/src/task-detail/cloudToolChanges.ts
```diff
@@ -189,48 +231,38 @@ export function buildCloudEventSummary(
...
   }
 
-  return { toolCalls };
+  let projectedState: TrackerState | undefined;
+  let projectedRevision = -1;
+  let projectedResult: CloudEventSummary = { toolCalls: new Map() };
+
+  return createAppendOnlyTracker<TrackerState, CloudEventSummary>({
+    init: () => ({ toolCalls: new Map(), revision: 0 }),
+    processEvent: (state, event) => {
```
k11kirky, line 289: **The revision bumps for *any* tool-call event, so the tracker clones the whole Map and re-arms an O(toolCalls) downstream memo on every streamed flush.**

`applyCloudEvent` returns true for every `tool_call` / `tool_call_update` carrying a `toolCallId` — no filter on whether a diff-relevant field (`locations`, `content`, terminal `status`) actually changed. `getResult` then returns a fresh `{ toolCalls: new Map(state.toolCalls) }`, and `useCloudRunState.ts:42` keys on that identity:

```ts
const fallbackFiles = useMemo(() => extractCloudToolChangedFiles(summary.toolCalls), [summary]);
```

On a run 400 tool calls deep, each 50ms flush clones a 400-entry Map and re-scans all 400 tool calls even when the event only nudged a partial-input snapshot — the memo never holds, so the "single-pass incremental" tracker still costs O(toolCalls) per frame in its consumer. Bumping the revision only on diff-relevant changes (or exposing it so the memo keys on a number rather than a cloned Map) would fix both halves.

packages/ui/src/features/sessions/components/GeneratingIndicator.tsx
```diff
@@ -152,7 +152,7 @@ export function GeneratingIndicator({
     const startTime = startedAt ?? Date.now();
     const interval = setInterval(() => {
       setElapsed(Math.max(0, Date.now() - startTime - pausedRef.current));
-    }, 100);
+    }, 1000);
```
k11kirky: **100ms → 1000ms without an immediate first tick means the indicator shows `0s` for a full second on every mount.**

`elapsed` is `useState(0)` and written *only* inside this callback, and the render consumes `formatDuration(elapsed, 0)` rather than recomputing from `startedAt`. Remount against a `startedAt` 30s in the past — reopening a session view mid-run, or the virtualized thread remounting the footer — and the first paint plus the following ~1000ms read `(Esc to stop • 0s)` before snapping to `30s`. At the old 100ms cadence the wrong value was invisible.

The two sibling timers this same PR touches (`RetryingStatusView`, `CompactingStatusView` in `StatusNotificationView.tsx`) both call `tick()` before `setInterval` for exactly this reason:

```ts
const tick = () => setElapsed(Math.max(0, Date.now() - startTime - pausedRef.current));
tick();
const interval = setInterval(tick, 1000);
```

packages/core/src/sessions/sessionService.ts
```diff
@@ -2553,9 +2563,36 @@ export class SessionService {
     const events = this.pendingSessionEvents.get(taskRunId);
     if (!events) return;
     this.pendingSessionEvents.delete(taskRunId);
-    for (const acpMsg of events) {
-      this.handleSessionEvent(taskRunId, acpMsg);
+    this.handleSessionEventBatch(taskRunId, events);
+  }
+
+  private handleSessionEventBatch(
```
k11kirky, line 2617: **Nit — this `else` branch is unreachable.**

The only writer of `pendingSessionEvents` content is `enqueueSessionEvent`, and its single call site (`onData`, below) gates on `isPassiveStreamEvent(event)`. Both callers of `handleSessionEventBatch` pass arrays drained from that map, so every element satisfies the predicate and `flushPassive()` / `handleSessionEvent(...)` in the `else` can never run.

That makes the mutable accumulator, the closure, the per-event re-classification and the trailing `flushPassive()` all dead. The method collapses to one `appendPassiveEvents(taskRunId, events)`: one session lookup, one `initialPrompt` clear, one `appendEvents`, one `updatePromptStateFromEvents`. If the invariant is worth guarding, assert it rather than branching on it.

(If you take the suggestion on the `onData` comment — batching all `session/update`s — then the branch becomes live and this comment is moot.)

packages/workspace-server/src/services/process-tracking/process-utils.ts
```diff
@@ -1,46 +1,171 @@
...
+  schedule: (callback: () => void, delayMs: number) => void;
+}
+
+export function killUnixProcessTrees(
+  rootPids: readonly number[],
+  initialProcesses: readonly ProcessEntry[],
+  ownPgid: number | undefined,
+  deps: UnixProcessKillerDeps,
+): void {
+  const trees = rootPids
```
k11kirky: **Perf nit, in a PR about resource use:** `findProcessTree` rebuilds the full `ppid → children` Map over the entire process table on *every* call, and this `.map` calls it once per root pid — O(rootPids × table) before the `trees.flat()` dedupe. Building the index once outside the map and passing it in makes it O(table + rootPids).

Separately, `snapshotUnixProcesses()` is a blocking `execFileSync("ps", ["-axo", ...])` full-table dump, and the per-pid path (`killProcessTree` → `ProcessTrackingService.kill`) pays one per call plus a second in the 5s SIGKILL callback. Closing 25 shells one at a time is 50 synchronous full-table dumps, and `@preDestroy killAll()` adds one to app shutdown.

Related: `ownPgid` comes from a single unguarded `processes.find(e => e.pid === process.pid)?.pgid`, and tracked agent children share the server's pgid (spawned without `detached` — `packages/agent/src/pi/rpc-client.ts`). I couldn't construct a realistic way to make that lookup miss while other rows parse, but if it ever did, `excludedPgid` is `undefined`, the filter `pgid !== excludedPgid` passes, and the killer SIGTERMs `-ownPgid` — the app itself. A fallback to the current process's group, or an assertion, is cheap insurance.

packages/ui/src/features/git-interaction/components/CloudGitInteractionHeader.tsx
```diff
@@ -35,7 +35,14 @@ export function CloudGitInteractionHeader({
   taskId,
   task,
 }: CloudGitInteractionHeaderProps) {
-  const session = useSessionForTask(taskId);
+  const inProgress = useSessionSelector(
+    taskId,
```
k11kirky: Nit: this duplicates the existing `useSessionHandoffInProgress` hook in `packages/ui/src/features/sessions/useSession.ts:191` — which this same PR correctly switches to in the sibling `TaskHeaderActions.tsx`. Worth using it here too so a future change to what "handoff in progress" means lands in one place.

packages/core/src/sidebar/buildSidebarData.ts
```diff
@@ -83,6 +83,7 @@ export function filterVisibleTasks(
 }
 
 export interface TaskSession {
+  status?: string;
```
k11kirky: Nit: the real type is a union — `SessionStatus = "connecting" | "connected" | "disconnected" | "error"` (`packages/shared/src/sessions.ts:43`). `status?: string` sits directly above a properly-typed `cloudStatus?: TaskRunStatus`, and this file already types the task-run field as `status?: TaskRunStatus | null`, so a typo'd status value here would never fail to compile.

(Also worth noting `useChannelTaskData.ts:27` builds a `TaskSession` literal without `status` — harmless today since `deriveTaskData` reads `cloudStatus`, not `status`, but it means the hand-maintained signature and the fields the sidebar actually renders can drift silently. Deriving the signature from the same projection the sidebar consumes would make this class of staleness impossible rather than fixing one instance.)

packages/workspace-server/src/services/process-tracking/process-utils.ts
richardsolomou: done, orphaned trees now fall back to an immediate tracked group/pid signal while delayed kills still require identity revalidation.

packages/workspace-server/src/services/process-tracking/process-utils.ts
richardsolomou: done, each Windows pid now has its own taskkill error boundary so one stale pid cant skip the rest.

packages/core/src/pi-runtime/piSessionController.ts
richardsolomou: done, initial session replacement now discards the overlapping pending buffer through the same helper as refresh. added the load-race regression too.

packages/ui/src/features/sessions/components/chat-thread/chatRowGrouping.ts
richardsolomou: done, turn completion now replaces every affected row (including child rows), so retained memoized turns see the flag change.

packages/core/src/sessions/sessionService.ts
richardsolomou: done, all session/update notifications are batched now. terminal, permission, response, and runtime events stay immediate, and the test covers interleaved text/tool traffic.

packages/ui/src/features/sessions/components/incrementalConversationItems.ts
richardsolomou: done, the builder tracks the lowest row touched by any completion in the batch. added the late cancelled-turn response case.

packages/core/src/task-detail/cloudToolChanges.ts
richardsolomou: done, changed-file derivation now has a separate revision keyed only to fields it consumes, and the tracker no longer clones the whole Map per tool update.

packages/ui/src/features/sessions/components/GeneratingIndicator.tsx
richardsolomou: done, the elapsed timer ticks immediately on mount. added a remount regression with an older startedAt.

packages/ui/src/features/git-interaction/components/CloudGitInteractionHeader.tsx
richardsolomou: done, this now reuses useSessionHandoffInProgress and keeps the separate cloudBranch selector.

packages/core/src/sidebar/buildSidebarData.ts
richardsolomou: done, TaskSession.status now uses the shared SessionStatus union.

packages/core/src/sessions/sessionService.ts
richardsolomou: this one is moot now: the buffer intentionally contains every session/update variant, so the mixed-event branch is live again for future callers and immediate event types never enter it.

packages/workspace-server/src/services/process-tracking/process-utils.ts
richardsolomou: done, the process table is indexed once for the whole root batch. a missing own pgid takes the conservative raw-root fallback instead of targeting discovered groups.

packages/workspace-server/src/services/process-tracking/process-utils.ts
```diff
@@ -1,44 +1,202 @@
...
+
+export function killUnixProcessTrees(
+  rootPids: readonly number[],
+  initialProcesses: readonly ProcessEntry[],
+  ownPgid: number | undefined,
+  deps: UnixProcessKillerDeps,
+): void {
+  const children = indexProcesses(initialProcesses);
+  const trees = rootPids
+    .map((pid) => findProcessTreeFromIndex(pid, initialProcesses, children))
```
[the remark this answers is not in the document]
richardsolomou: done, fallback is now tracked independently per requested root, so a missing root still gets its raw group/pid signal when another root is present. added the mixed-batch regression.

packages/core/src/sessions/sessionService.ts
```diff
@@ -1535,6 +1535,11 @@ function classifyTurnEventKind(
   return "other";
 }
 
+function isStreamUpdateEvent(acpMsg: AcpMessage): boolean {
+  const msg = acpMsg.message;
+  return "method" in msg && msg.method === "session/update";
```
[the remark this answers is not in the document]
richardsolomou: done, config_option_update now stays on the side-effectful path so the store and persisted permission mode update immediately. added a regression covering the persistence call.

charlesvien:
Hey @richardsolomou, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3808, opened 2026-07-25
feat(sidebar): show repository context on pinned tasks

haacked:
## Problem

Pinned tasks only show a title. They sit above the per-project groups, so unlike every other row they never get a repository from a group header, and once you have a dozen of them pinned there's no way to tell which checkout one belongs to. Supacode's pinned list shows the repo under each task and I wanted the same thing here.

## Changes

Before:

Pinned tasks showing only titles

After:

Pinned tasks with a repository and branch under each title

- Pinned rows get a second line: the repository, plus the branch when the task has one of its own
- The repo name comes from the registered folder, so it reads the same as the group header the task would sit under. A task with no repo stays on one line (last row above)
- The branch is the task's linked branch, falling back to a worktree's checked-out branch. Local tasks sitting on the default branch show no branch at all, so you don't get a `· main` repeated down the whole list
- Rows inside a project group are untouched, they already have the header above them
- Quill's `Button` pins a fixed height, so the two-line row opts out with `h-auto!` the same way `CommandMenu` and `MessageJumpPicker` already do

<details>
<summary>Light theme</summary>

Pinned tasks in the light theme

</details>

## How did you test this?

- `pnpm typecheck` and `biome check`, both clean
- Unit tests for the new `formatTaskContext`: worktree vs local vs cloud, linked branch vs checked-out branch, default-branch suppression, folder naming, and the repo-less fallback
- `vitest run src/features/sidebar` in packages/ui (111 pass) and `pnpm --filter @posthog/core test` (2507 pass)
- New `Sidebar/TaskItem` story, which is where the screenshots above come from. Also eyeballed the active-row and hover-toolbar states in both themes
- `node scripts/check-host-boundaries.mjs`, no new violations

## Automatic notifications

- [ ] Publish to changelog?

---


charlesvien:
Remade in the monorepo as PostHog/posthog#76471 (PostHog/code main is frozen after the desktop import).

PR #3838, opened 2026-07-27
fix(sessions): upload cloud attachments before sending

puemos:
## Problem

Cloud follow-up messages previously started uploading attachments only after the user pressed Enter. The composer cleared immediately, but the optimistic message could not appear until attachment processing and authentication lookups reached the send path. A second message could also be submitted while the first message’s files were still uploading, and the composer gave no attachment upload feedback.

## Changes

- Start the existing batched cloud attachment upload as soon as files are added to the composer
- Show a spinner or error indicator on every affected attachment
- Disable Enter and the send button until the attachment batch succeeds
- Keep failed uploads blocked with an error toast until files are removed or reattached
- Cache the prepared artifact IDs and reuse them during submission
- Insert the existing optimistic message before authentication lookup, rolling it back through existing queue and error handling
- Leave text-only and local-session submission unchanged

## How did you test this?

- `pnpm build`
- 192 focused session, composer, and attachment tests
- Biome checks on changed files

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/sessions/cloudTaskUpdateNotifications.test.ts
+++ b/packages/core/src/sessions/cloudTaskUpdateNotifications.test.ts
@@ -136,7 +136,6 @@ function createHarness() {
     },
     clearTailOptimisticItems: vi.fn(),
     appendOptimisticItem: vi.fn(),
-    updateOptimisticItem: vi.fn(),
     replaceOptimisticWithEvent: vi.fn(),
     clearMessageQueue: vi.fn(),
   };
--- a/packages/core/src/sessions/cloudTaskUpdateNotifications.test.ts
+++ b/packages/core/src/sessions/cloudTaskUpdateNotifications.test.ts
@@ -136,6 +136,7 @@ function createHarness() {
     },
     clearTailOptimisticItems: vi.fn(),
     appendOptimisticItem: vi.fn(),
+    updateOptimisticItem: vi.fn(),
     replaceOptimisticWithEvent: vi.fn(),
     clearMessageQueue: vi.fn(),
   };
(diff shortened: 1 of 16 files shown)


puemos (commented):
Two lifecycle issues remain:

- [P1] `packages/core/src/sessions/sessionService.ts:4103`: the optimistic item is appended only after `getAuthCredentialsStatus()` and `getCloudCommandAuth()` have both awaited `auth.getState` IPC calls. The composer is cleared before `sendPrompt`, so users still see the message disappear for the full authentication lookup—the exact delay this PR says it removes. Create the optimistic item before those awaits (and transition/remove it appropriately for queued/auth-failure outcomes).

- [P1] `packages/core/src/sessions/sessionStore.ts:426`: `replaceOptimisticWithEvent` still assigns `optimisticItems = []`. After one send fails and leaves a `deliveryStatus: "failed"` bubble, the next successful non-steer prompt echo clears that failed bubble along with its own optimistic item. This means failed messages are not actually preserved in the transcript once the user sends anything else. Reconcile only the optimistic item corresponding to the echoed prompt, while retaining failed/unrelated items.

puemos:
Addressed the still-relevant first review point in 06b4c7fb: the optimistic bubble is now inserted before both authentication awaits, with the existing queue/error paths clearing it when delivery cannot proceed. The attachment-first flow also finishes uploads before Enter is enabled.

The second point is obsolete after the lean rewrite: this PR no longer introduces `deliveryStatus` or preserves failed optimistic bubbles. Failures use the existing toast plus composer restoration, and `replaceOptimisticWithEvent` retains its prior behavior.

puemos:
Addressed the current React Doctor feedback in 9544ab43. Attachment changes are now reported from the editor’s existing add/remove/restore mutations instead of synchronizing parent state through a callback effect. This removes both `no-pass-data-to-parent` and `no-prop-callback-in-effect` warnings.

tatoalo (approved):

PR #3840, opened 2026-07-27
fix(git): record every PR created for a local task

frankh:
## Problem

On a **local (desktop worktree) task** that produced two PRs, the task header shows only **one** PR badge (the latest / current branch) with **no "Other PRs" submenu**. Multi‑PR‑per‑task support exists (#3227) via a `pr_urls` array, but for local tasks nothing recorded a PR *at creation time*.

`pr_urls` was only populated as a side effect of polling `getPrStatus` on the **currently checked‑out branch** during revalidation. So if the agent opens PR #1 on branch A, then continues and opens PR #2 on branch B before any revalidation fires while A is current, PR #1 is never recorded — only PR #2 survives, and there's nothing to populate the submenu.

## Fix

1. **Event‑driven capture (primary).** The Create‑PR saga now accumulates the created PR URL the moment the PR is made, independent of which branch is later checked out — mirroring how cloud already works. New `accumulatePrUrl` seam: `CreatePrHost` → `GitWorkspaceLookup` → `IGitPrStatus` → `TaskPrStatusService.accumulatePrUrl` (does `updatePrCache(..., accumulate: true)` + emits `taskPrInfoChanged`), wired in the composition root. No schema change — the `pr_urls` column already exists. Array order is preserved, so promote‑to‑primary and dedupe still hold.

2. **Local agent hardening.** `attachPrIfCreatedThisRun` enforced the author match unconditionally, so a genuinely agent‑created PR was silently dropped whenever `gh api user` returned no login. Now the author check applies only when a login actually resolved (matching `agent-server.ts`); the 5‑minute recency gate still scopes attribution.

## Tests
- New `TaskPrStatusService.accumulatePrUrl` suite: first PR sets primary + `pr_urls=[A]`; second distinct PR yields `[A,B]`; re‑accumulating A is idempotent; no‑op when the task has no workspace row.
- Saga test asserts `accumulatePrUrl` is called on success.
- Explicit `resolveTaskPrUrls` regression case for two accumulated local PRs surfacing an "Other PRs" entry.

## Verification
- Typecheck: `@posthog/core`, `@posthog/workspace-server`, `@posthog/host-router`, `apps/code` — clean.
- Tests: core (2503), ui (2164), `task-pr-status` (12) — pass.
- Lint clean, zero `noRestrictedImports` in `packages/core`; no new host‑boundary violations.

---

### DIFF
--- a/apps/code/src/main/di/container.ts
+++ b/apps/code/src/main/di/container.ts
@@ -584,6 +584,11 @@ container
       getWorkspace: (taskId) => workspace().getWorkspace(taskId),
       linkBranch: (taskId, branch, source) =>
         workspace().linkBranch(taskId, branch, source),
+      accumulatePrUrl: (taskId, prUrl) => {
+        void ctx
+          .get<IGitPrStatus>(GIT_PR_STATUS_PROVIDER)
+          .accumulatePrUrl(taskId, prUrl);
+      },
     };
   });
 container.load(gitHostModule);
--- a/packages/core/src/git-pr/git-pr.test.ts
+++ b/packages/core/src/git-pr/git-pr.test.ts
@@ -126,6 +126,7 @@ function makeHost(over: Partial<CreatePrHost> = {}): CreatePrHost {
       prUrl: "https://github.com/o/r/pull/1",
     }),
     linkBranch: vi.fn(),
+    accumulatePrUrl: vi.fn(),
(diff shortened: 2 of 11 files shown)


charlesvien:
Hey @frankh, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3848, opened 2026-07-27
feat(loops): edit loop details inline

MattPua:
## Problem

Editing loop metadata, schedules, or model configuration requires navigating to the full edit flow.

## Changes

- Edit titles, descriptions, and instructions inline
- Edit schedule and model configuration from loop details
- Preserve unsaved configuration across unrelated loop updates
- Save metadata on blur, with keyboard cancel support

## How did you test this?

- UI typecheck and loop tests covered by CI
- Focused QA across logic, contracts/security, and performance/reliability

## Automatic notifications

- [ ] Publish to changelog?

---


charlesvien:
Does this work with GitHub triggers and the other options? Maybe we should just prompt that same wizard view and make it a shared dialog and just have an edit button.

MattPua:
> Does this work with GitHub triggers and the other options? Maybe we should just prompt that same wizard view and make it a shared dialog and just have an edit button.

do you mean like show the wizard view as a modal vs the inline?

i havent checked the other triggers yet, selfishly only looked at the stuff i needed

PR #3858, opened 2026-07-28
fix(tasks): restore cloud artifact uploads

puemos:
## Problem

Cloud artifact uploads can report that storage is not configured before making any API or object-storage request. The sandbox credential refresh work in [PostHog/posthog#72540](https://github.com/PostHog/posthog/pull/72540) deliberately separated rotating credentials from general `/tmp/agent-env` metadata:

- GitHub credentials live in `/tmp/agent-github-env`
- the PostHog OAuth credential lives in `/tmp/agent-oauth-env`
- an existing but empty credential file is an explicit logout/revocation

The shared PostHog API client resolver still looked for `POSTHOG_PERSONAL_API_KEY` in `/tmp/agent-env` or the process-start environment. That misses refreshed credentials in Codex local-tool processes and could resurrect a stale process token after revocation.

Codex exposed the missing-key path after [#3844](https://github.com/PostHog/code/pull/3844) restored the task run ID passed to its local-tools process: calls then advanced to the storage-configuration guard because no PostHog API client could be constructed.

**Why:** Agent-created reports and other non-code deliverables need the current managed sandbox identity to prepare and finalize durable task artifact uploads. Credential refresh and revocation must both take effect without restarting the agent.

## Changes

- Read the rotating personal API key exclusively from `/tmp/agent-oauth-env`
- Resolve the key from one file snapshot for every request and authentication retry
- Require the managed OAuth credential file; missing, empty, or unreadable files all fail closed
- Keep non-secret API URL and project metadata resolution unchanged
- Share one tested NUL-delimited environment parser across metadata and credential files
- Add regression coverage for rotation between requests, authenticated retries, explicit revocation, and invalid managed files

## How did you test this?

- `pnpm --filter @posthog/agent exec vitest run src/signed-commit-artefacts.test.ts src/posthog-api.test.ts` (19 tests)
- `pnpm --filter @posthog/agent typecheck`
- `pnpm exec biome check packages/agent/src/signed-commit-artefacts.ts packages/agent/src/signed-commit-artefacts.test.ts`

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/agent/src/signed-commit-artefacts.ts
+++ b/packages/agent/src/signed-commit-artefacts.ts
@@ -42,22 +42,20 @@ function readSandboxEnvFile(envFilePath: string): Record<string, string> {
 }
 
 function readSandboxOauthToken(oauthEnvFilePath: string): string | undefined {
-  let raw: string;
   try {
-    raw = readFileSync(oauthEnvFilePath, "utf8");
+    const raw = readFileSync(oauthEnvFilePath, "utf8");
+    for (const entry of raw.split("\0")) {
+      const prefix = "POSTHOG_PERSONAL_API_KEY=";
+      if (entry.startsWith(prefix)) {
+        return entry.slice(prefix.length);
+      }
+    }
+    return "";
   } catch {
     // The dedicated credential channel is mandatory. Missing and unreadable
     // files both fail closed.
     return undefined;
   }
-
-  // The backend revokes OAuth access by truncating this managed file. Its
-  // presence is authoritative even when empty.
-  if (raw.trim() === "") {
-    return "";
-  }
-  const oauthEnv = readSandboxEnvFile(oauthEnvFilePath);
-  return oauthEnv.POSTHOG_PERSONAL_API_KEY ?? "";
 }
 
 export function resolveSandboxPosthogApi(
@@ -89,7 +87,10 @@ export function createSandboxPosthogClient(
   return new PostHogAPIClient({
     apiUrl: api.apiUrl,
     projectId: api.projectId,
-    getApiKey: () => api.apiKey,
+    getApiKey: () =>
+      readSandboxOauthToken(oauthEnvFilePath ?? SANDBOX_OAUTH_ENV_FILE) ?? "",
+    refreshApiKey: () =>
+      readSandboxOauthToken(oauthEnvFilePath ?? SANDBOX_OAUTH_ENV_FILE) ?? "",
   });
 }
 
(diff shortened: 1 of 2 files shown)


puemos:
@tatoalo i'd love your eyes on that since I'm not sure if that's the right approach here. my trust in the fix isn't 100% since the changes in the env file were merged last week so I wonder why we didn't have issues until now

packages/agent/src/signed-commit-artefacts.ts
```diff
@@ -41,28 +41,48 @@ function readSandboxEnvFile(envFilePath: string): Record<string, string> {
...
+    // files both fail closed.
+    return undefined;
+  }
+
+  // The backend revokes OAuth access by truncating this managed file. Its
+  // presence is authoritative even when empty.
+  if (raw.trim() === "") {
+    return "";
+  }
+  const oauthEnv = readSandboxEnvFile(oauthEnvFilePath);
```
tatoalo: I believe here we are reading thefile a second time after validating the first snapshot, a refresh between those reads could validate one generation and parse another. Could we parse `POSTHOG_PERSONAL_API_KEY` directly from raw so each resolution uses one consistent snapshot?

tatoalo (commented):
`getApiKey` always returns the token captured when the client is created, if the managed OAuth file is rotated or truncated between requests, subsequent requests and the built-in 401/403 retry will reuse the stale token, potentially leaving artifact reporting partially complete which is a bit of a mess for users to end-up in, could we reread the managed file in getApiKey/refreshApiKey and add a test that changes the file between requests?

packages/agent/src/signed-commit-artefacts.ts
puemos: Addressed in 93fc7d0: the resolver now parses the token directly from the single `readFileSync` snapshot, so validation and parsing cannot observe different file generations.

puemos:
@tatoalo Addressed both points in 93fc7d0: `getApiKey` and `refreshApiKey` now reread the managed OAuth file for every resolution, and focused tests rotate the real file between ordinary requests and a 401 retry. The parser uses one snapshot per resolution. Would appreciate a re-review when convenient.

---

Sorry my agent decided to post without asking me :/

puemos:
Follow-up in 77c4658: extracted a pure, directly tested NUL-delimited environment parser and replaced the key-specific loop with an object lookup. The OAuth file is still read exactly once per credential resolution.

tatoalo (approved):
thanks for addressing, looks great!

PR #3874, opened 2026-07-28
feat: add recent activity hover popover

puemos:
## Problem

Recent task activity takes a full-page navigation to review.

**Why:** People should be able to quickly scan recent updates from the activity bell without leaving their current view.

## Changes

- Add a delayed, right-aligned activity popover inspired by Slack’s compact activity panel.
- Show recent activity using the existing shared feed API and cache.
- Reuse existing row navigation and per-task mark-read behavior, including a batched “Mark all as read” action for visible unread rows.

## How did you test this?

- Focused activity and hover tests: 10 passed.
- UI, API-client, and web-host typechecks passed.
- Biome checks passed for changed files.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
+++ b/packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
@@ -14,7 +14,7 @@ import { useEffect, useMemo } from "react";
 export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
   const client = useOptionalAuthenticatedClient();
   const { data: currentUser } = useCurrentUser({ client });
-  const { items, isLoading } = useTaskActivity();
+  const { items, unreadCount, isLoading } = useTaskActivity();
   const unreadItems = items.filter((item) => item.isUnread);
   const { mutate: markTasksRead, isPending: isMarkingRead } =
     useMarkTaskActivityRead();
@@ -67,7 +67,9 @@ export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
             onClick={markAllRead}
           >
             <ChecksIcon size={14} />
-            Mark all as read
+            {unreadItems.length === unreadCount
+              ? "Mark all as read"
+              : "Mark visible as read"}
           </Button>
         )}
       </div>
--- a/packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
+++ b/packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
@@ -1,5 +1,14 @@
-import { ChecksIcon } from "@phosphor-icons/react";
-import { Button, PopoverContent, Spinner } from "@posthog/quill";
+import { BellIcon, ChecksIcon } from "@phosphor-icons/react";
+import {
+  Button,
+  Empty,
+  EmptyDescription,
+  EmptyHeader,
+  EmptyMedia,
+  EmptyTitle,
+  PopoverContent,
+  Spinner,
+} from "@posthog/quill";
 import { ANALYTICS_EVENTS } from "@posthog/shared/analytics-events";
 import { useOptionalAuthenticatedClient } from "@posthog/ui/features/auth/authClient";
 import { useCurrentUser } from "@posthog/ui/features/auth/useCurrentUser";
@@ -7,12 +16,26 @@ import { ActivityRow } from "@posthog/ui/features/canvas/components/ActivityView
 import { useChannels } from "@posthog/ui/features/canvas/hooks/useChannels";
 import { useMarkTaskActivityRead } from "@posthog/ui/features/canvas/hooks/useMarkTaskActivityRead";
 import { useTaskActivity } from "@posthog/ui/features/canvas/hooks/useTaskActivity";
-import { normalizeChannelName } from "@posthog/ui/features/canvas/hooks/useTaskChannels";
 import { useInView } from "@posthog/ui/primitives/hooks/useInView";
 import { track } from "@posthog/ui/shell/analytics";
-import { useEffect, useMemo } from "react";
+import { useEffect, useMemo, useState } from "react";
+import {
+  activityReadPayload,
+  channelIdForName,
+  createChannelIdByName,
+  getUnreadActivityItems,
+  markLoadedReadLabel,
+} from "./activityFeed";
 
-export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
+interface ActivityHoverCardProps {
+  onClose: () => void;
+  side?: "bottom" | "right";
+}
+
+export function ActivityHoverCard({
+  onClose,
+  side = "right",
+}: ActivityHoverCardProps) {
   const client = useOptionalAuthenticatedClient();
   const { data: currentUser } = useCurrentUser({ client });
   const {
@@ -23,21 +46,17 @@ export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
     isFetchingNextPage,
     fetchNextPage,
   } = useTaskActivity();
+  const [scrollRoot, setScrollRoot] = useState<HTMLDivElement | null>(null);
   const [loadMoreRef, loadMoreInView] = useInView<HTMLDivElement>({
+    root: scrollRoot,
     rootMargin: "100px 0px",
   });
-  const unreadItems = items.filter((item) => item.isUnread);
+  const unreadItems = getUnreadActivityItems(items);
   const { mutate: markTasksRead, isPending: isMarkingRead } =
     useMarkTaskActivityRead();
   const { channels } = useChannels();
   const folderIdByName = useMemo(
-    () =>
-      new Map(
-        channels.map((channel) => [
-          normalizeChannelName(channel.name),
-          channel.id,
-        ]),
-      ),
+    () => createChannelIdByName(channels),
     [channels],
   );
   useEffect(() => {
@@ -47,27 +66,22 @@ export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
     });
   }, []);
   useEffect(() => {
-    if (loadMoreInView && hasNextPage && !isFetchingNextPage) {
+    if (loadMoreInView && hasNextPage) {
       void fetchNextPage();
     }
-  }, [fetchNextPage, hasNextPage, isFetchingNextPage, loadMoreInView]);
+  }, [fetchNextPage, hasNextPage, loadMoreInView]);
 
   const markRead = (taskId: string, activityAt: string) => {
     markTasksRead([{ task_id: taskId, seen_before: activityAt }]);
   };
 
   const markAllRead = () => {
-    markTasksRead(
-      unreadItems.map((item) => ({
-        task_id: item.taskId,
-        seen_before: item.activityAt,
-      })),
-    );
+    markTasksRead(activityReadPayload(unreadItems));
   };
 
   return (
     <PopoverContent
-      side="right"
+      side={side}
       align="start"
       sideOffset={8}
       className="w-[380px] gap-0 overflow-hidden p-0"
@@ -83,34 +97,37 @@ export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
             onClick={markAllRead}
           >
             <ChecksIcon size={14} />
-            {unreadItems.length === unreadCount
-              ? "Mark all as read"
-              : "Mark visible as read"}
+            {markLoadedReadLabel(unreadItems.length, unreadCount)}
           </Button>
         )}
       </div>
-      <div className="max-h-[480px] overflow-y-auto p-1.5">
+      <div ref={setScrollRoot} className="max-h-[480px] overflow-y-auto p-1.5">
         {isLoading && items.length === 0 ? (
           <div className="flex justify-center py-10">
             <Spinner />
           </div>
         ) : items.length === 0 ? (
-          <div className="px-2 py-8 text-center text-muted-foreground text-sm">
-            No recent activity.
-          </div>
+          <Empty className="border-0 py-8">
+            <EmptyHeader>
+              <EmptyMedia variant="icon">
+                <BellIcon />
+              </EmptyMedia>
+              <EmptyTitle>No recent activity</EmptyTitle>
+              <EmptyDescription>
+                New task updates will appear here.
+              </EmptyDescription>
+            </EmptyHeader>
+          </Empty>
         ) : (
           <div className="flex flex-col gap-0.5">
             {items.map((item) => (
               <ActivityRow
                 key={item.taskId}
                 item={item}
-                folderChannelId={
-                  item.channelName
-                    ? (folderIdByName.get(
-                        normalizeChannelName(item.channelName),
-                      ) ?? null)
-                    : null
-                }
+                folderChannelId={channelIdForName(
+                  folderIdByName,
+                  item.channelName,
+                )}
                 onOpen={(activity) =>
                   markRead(activity.taskId, activity.activityAt)
                 }
@@ -123,13 +140,11 @@ export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
                 compact
               />
             ))}
-            {hasNextPage && (
-              <div ref={loadMoreRef} className="flex h-8 justify-center py-2">
-                {isFetchingNextPage && <Spinner />}
-              </div>
-            )}
           </div>
         )}
+        <div ref={loadMoreRef} className="flex h-8 justify-center py-2">
+          {hasNextPage && isFetchingNextPage && <Spinner />}
+        </div>
       </div>
     </PopoverContent>
   );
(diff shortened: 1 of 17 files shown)


packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { normalizeChannelName } from "@posthog/ui/features/canvas/hooks/useTaskChannels";
+import { track } from "@posthog/ui/shell/analytics";
+import { useEffect, useMemo } from "react";
+
+export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
+  const client = useOptionalAuthenticatedClient();
+  const { data: currentUser } = useCurrentUser({ client });
+  const { items, isLoading } = useTaskActivity();
+  const unreadItems = items.filter((item) => item.isUnread);
+  const { mutate: markTasksRead, isPending: isMarkingRead } =
```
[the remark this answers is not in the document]
puemos: Valid — fixed in 59b958f0. The action now compares loaded unread rows with the authoritative global unread count and says “Mark visible as read” whenever unloaded unread activity remains.

packages/ui/src/features/canvas/components/ChannelNav.tsx
```diff
@@ -94,15 +127,39 @@ export function ChannelNav() {
...
-      />
+      <Popover
+        open={!isActivity && activityOpen}
+        onOpenChange={(open) => setActivityOpen(!isActivity && open)}
+      >
+        <PopoverTrigger
+          openOnHover
+          delay={300}
+          closeDelay={30}
+          disabled={isActivity}
```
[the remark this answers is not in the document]
puemos: Valid — fixed in 59b958f0. Both triggers now use the repository’s 100 ms travel-popover precedent, preserving access to the offset card while still closing promptly.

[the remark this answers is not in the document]
puemos: done

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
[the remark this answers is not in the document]
puemos: done

packages/ui/src/features/canvas/components/ChannelNav.tsx
```diff
@@ -94,15 +127,39 @@ export function ChannelNav() {
...
-      />
+      <Popover
+        open={!isActivity && activityOpen}
+        onOpenChange={(open) => setActivityOpen(!isActivity && open)}
+      >
+        <PopoverTrigger
+          openOnHover
+          delay={300}
+          closeDelay={100}
+          disabled={isActivity}
```
k11kirky: **`disabled` reaches the DOM and genuinely disables the Activity bell.**

Base UI's `PopoverTrigger` defaults `nativeButton = true` and passes `disabled` through `useButton` → `useFocusableWhenDisabled`, which sets `additionalProps.disabled = disabled`; `useRenderElement` then merges it into the `render` element. `NavButton` destructures only `icon, label, isActive, onClick, badge, className, ref`, so `disabled` stays in `...buttonProps` and is spread onto `<button>`.

On the Activity view the bell drops out of the tab order and its `onClick` (track + `navigateToActivity`) can never fire — unlike Inbox and Command Center, which stay clickable while active.

It's also redundant: `open={!isActivity && activityOpen}` plus the `onOpenChange` guard already make the popover unopenable here. Deleting this line loses nothing and restores a clickable, tabbable bell.

k11kirky (commented):
Automated review pass (8 findings, verified against the real `@base-ui/react` 1.3.0 stack where behaviour was in question). Ranked most-severe first as inline comments.

The two highest-impact items are functional regressions in the trigger wiring: `disabled={isActivity}` genuinely disables the nav bell, and the sidebar trigger has no `isActive` guard so a plain click both navigates *and* pops the card open. Both reproduced empirically rather than inferred.

Also checked and **refuted** (no action needed): the `unreadItems.length === unreadCount` label logic (both are unread-*task* counts and the optimistic update decrements them in lockstep — `hasNextPage` would be strictly worse); the `pr-14` "dead click strip" (it's padding inside the button, and the overlay is `pointer-events-none`, so clicks land fine); and `import { ActivityHoverCard } from "./ActivityHoverCard"` (AGENTS.md prohibits *deep* relative imports — same-directory `./` is established practice in that folder).

One lower-severity item not worth its own thread: the `view_activity` mount effect fires on every 300ms hover-open (both call sites conditionally mount the card), merging cheap hovers into an existing metric that only `surface` distinguishes. I searched project 2's saved insights and found no currently-broken chart, so it's a trap for future queries rather than a live break — but worth a `surface` filter in any dashboard that uses it.

---

packages/ui/src/features/sidebar/components/items/ActivityItem.tsx
```diff
@@ -32,7 +36,23 @@ export function ActivityItem({
...
+      onClick={() => {
+        setOpen(false);
+        onClick();
+      }}
     />
   );
+  if (isActive) return item;
+
+  return (
+    <Popover open={open} onOpenChange={setOpen}>
```
k11kirky: **No `isActive` guard on `open`/`onOpenChange`, so a plain click both navigates and toggles the card open over the page it just navigated to.**

Reproduced against the real quill / `@base-ui/react` 1.3.0 stack: a single click with no hover yields `navCalls=1` **and** the hover card present, trigger at `aria-expanded="true"`.

`useClick` is enabled unconditionally on the trigger (`openOnHover` only adds a *separate* hover interaction), and nothing stops propagation between `SidebarItem`'s inner `<button>` and the trigger — so `setOpen(false); onClick()` runs, then the trigger toggles the closed popover open.

`isActive` derives from async router state (`navigateToActivity` is a fire-and-forget `router.navigate`), so React commits `open=true` before navigation lands and the 380px card flashes over the Activity page. `if (isActive) return item` only rescues this once the route commits.

ChannelNav guards both directions explicitly; this call site has neither.

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  markRead(activity.taskId, activity.activityAt)
+                }
+                currentUser={currentUser}
+                surface="activity_panel"
+                onNavigate={onClose}
+                compact
+              />
+            ))}
+            {hasNextPage && (
+              <div ref={loadMoreRef} className="flex h-8 justify-center py-2">
```
k11kirky: **Conditionally-mounted sentinel + mount-only observer = auto-pagination permanently dead for that popover instance.**

`useInView` does `const el = ref.current; if (!el) return; observer.observe(el)` with deps `[rootMargin, once]` — no callback ref, no re-run when the node appears later. This sentinel only renders under `items.length > 0 && hasNextPage`.

Open the card while the first `useTaskActivity` page is in flight (cold start, slow network), or before the auth client resolves (`enabled: !!client` → `isLoading` false with zero items, so the *empty* branch renders): `loadMoreRef.current` is null, the observer is never created, `loadMoreInView` stays false forever, and scrolling the 480px list never calls `fetchNextPage`.

The new test can't catch this — `ActivityHoverCard.test.tsx:46` mocks `useInView` to `[{ current: null }, true]`, stubbing out the exact wiring that's broken. The other `useInView` call sites (`ChannelFeedView`, `WebsiteDashboardsIndex`) attach the ref to an unconditionally rendered element; doing the same here (render the sentinel container always, gate only the spinner) fixes it.

packages/ui/src/features/sidebar/components/items/ActivityItem.tsx
```diff
@@ -32,7 +36,23 @@ export function ActivityItem({
...
+      <PopoverTrigger
+        openOnHover
+        delay={300}
+        closeDelay={100}
+        render={<div className="w-full">{item}</div>}
```
k11kirky: **`render={<div>}` without `nativeButton={false}` → invalid `type` on a div, popup ARIA on a roleless wrapper, a dead extra tab stop, and no keyboard path to the card.**

Verified by rendering this markup against `@posthog/quill@0.3.0-beta.24` / `@base-ui/react@1.3.0`. Emitted DOM:

```html
<div class="w-full" type="button" tabindex="0" aria-haspopup="dialog" aria-expanded="false">
  <button type="button" tabindex="0">…
```

Base UI also logs a dev-mode error: *"A component that acts as a button expected a native `<button>` because the `nativeButton` prop is true… Use a real `<button>` in the `render` prop, or set `nativeButton` to `false`."*

Focusing the div and dispatching Enter and Space gives `navCalls=0 hovercard=false`, because `getButtonProps`' key synthesis is gated on `!isNativeButton`. Net: a keyboard or screen-reader user gets two tab stops for one control, is told a popup exists on the one that can't open it, and can never reach the hover card.

ChannelNav's sibling trigger renders a real `<button>` and is unaffected. Fix is either `nativeButton={false}` or making the `SidebarItem` button itself the trigger.

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+      unreadItems.map((item) => ({
+        task_id: item.taskId,
+        seen_before: item.activityAt,
+      })),
+    );
+  };
+
+  return (
+    <PopoverContent
+      side="right"
```
k11kirky: **Hardcoded `side="right"` makes the 380px panel cover the Command Center icon on the horizontal nav rail.**

`ChannelNav` renders `<div className="flex shrink-0 gap-2 px-2 pt-2 pb-1">` with three `size-8` buttons in order Inbox, Activity, Command Center. `gap-2` is 8px and `sideOffset={8}`, so the panel's left edge lands *exactly* on the Lightning button's left edge and its 380px width fully covers the Command Center icon and its unread badge while open.

Base UI's positioner only flips on viewport collision, and there's ample room in the left sidebar, so nothing rescues it.

`ActivityHoverCard({ onClose })` exposes no `side`/`align` escape hatch, so the two call sites — which genuinely want different placements (`side="right"` *is* correct for the vertical sidebar row) — can't differ. Worth lifting placement to a prop, or to the caller's `PopoverContent`.

packages/ui/src/features/canvas/components/ChannelNav.tsx
```diff
@@ -94,15 +127,39 @@ export function ChannelNav() {
...
+          render={
```
k11kirky: **The Activity bell silently loses its tooltip.**

`NavIcon` wraps every icon in `<Tooltip content={label} shortcut={shortcut} side="bottom">`, and Inbox and Command Center still go through it. This branch calls `NavButton` directly from `render`, and `NavButton` only emits `<button aria-label={label}>` — `Tooltip` is now used solely by `NavIcon`.

Hovering the bell on the Activity view shows nothing at all (the popover is suppressed there); off it, the icon-only bell has no name until a 380px panel opens after 300ms. The `aria-label` survives so the a11y name is intact, but the visual affordance is gone and the rail is inconsistent.

(The old Activity `NavIcon` passed no `shortcut`, so no keyboard hint is lost — just the label.)

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+      ),
+    [channels],
+  );
+  useEffect(() => {
+    track(ANALYTICS_EVENTS.CHANNEL_ACTION, {
+      action_type: "view_activity",
+      surface: "activity_panel",
+    });
+  }, []);
+  useEffect(() => {
```
k11kirky, line 68: **`isFetchingNextPage` in the deps makes each settled page re-trigger the fetch, chaining unrequested pages on one 300ms hover.**

`useInView` defaults `once = false` and keeps `inView` true while intersecting, so `isFetchingNextPage` flipping true→false re-runs this effect and fires `fetchNextPage()` again — no scroll, no click. Compact rows are ~44-46px so ~11 fill the 480px container; if the backend page size is below that, the chain continues until `hasNextPage` is false and one hover pulls the entire feed sequentially. Bounded above that (overflow clipping empties the sentinel's intersection rect once rows exceed 480px), so a single spurious extra page per hover is the common case.

I couldn't confirm the server page size — `getTaskActivity` sends no `limit` and the default lives in the backend repo — so flagging this as plausible rather than certain. Worth checking against the real page size.

Separately: `rootMargin: "100px 0px"` is inert here. `useInView` sets no IntersectionObserver `root`, so the margin expands the *viewport* rect, not this popover's `overflow-y-auto` clip — there's no early prefetch, just a spinner stall at the bottom. The pre-existing `ActivityView` deliberately used an explicit "Load more" button instead.

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { useChannels } from "@posthog/ui/features/canvas/hooks/useChannels";
+import { useMarkTaskActivityRead } from "@posthog/ui/features/canvas/hooks/useMarkTaskActivityRead";
+import { useTaskActivity } from "@posthog/ui/features/canvas/hooks/useTaskActivity";
+import { normalizeChannelName } from "@posthog/ui/features/canvas/hooks/useTaskChannels";
+import { useInView } from "@posthog/ui/primitives/hooks/useInView";
+import { track } from "@posthog/ui/shell/analytics";
+import { useEffect, useMemo } from "react";
+
+export function ActivityHoverCard({ onClose }: { onClose: () => void }) {
+  const client = useOptionalAuthenticatedClient();
```
k11kirky, line 39: **The whole activity-feed shell is duplicated from `ActivityView`, and the two copies have already diverged.**

Lines 16-61 mirror `ActivityView.tsx:258-296`: the same `useTaskActivity` / `useMarkTaskActivityRead` / `useChannels` destructuring, an identical `new Map(channels.map(c => [normalizeChannelName(c.name), c.id]))`, hand-built `{ task_id, seen_before }` payloads, and the same mount-effect `track(CHANNEL_ACTION, { action_type: "view_activity" })`.

They already differ in a way that's a live bug: this popover relabels its button "Mark visible as read" when loaded unread rows don't cover `unreadCount`, while the page unconditionally says "Mark all as read" from the same partially-loaded `items` — so **the page's label is now wrong** and there's no single place to fix it.

The empty state here is also a hand-rolled div where `ActivityView` uses quill's `<Empty>`, which AGENTS.md:235 requires: *"Empty/placeholder/loading screens (canvas and elsewhere) are a `@posthog/quill` `<Empty>` … Don't hand-roll the centered Flex + dashed icon box."* Same PR, two different empty states for the same data.

Extracting a shared `useActivityFeed(surface)` + `ActivityList` would remove ~35 duplicated lines and leave each surface only its chrome.

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
puemos: Fixed: pagination observes the popover scroll container and no longer retriggers when a request settles.

packages/ui/src/features/canvas/components/ChannelNav.tsx
puemos: Fixed: removed the disabled prop; the Activity bell stays clickable and tabbable while the open-state guard suppresses the popover.

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
puemos: Fixed: the top rail opens below the bell, while the vertical sidebar opens to the right.

packages/ui/src/features/sidebar/components/items/ActivityItem.tsx
puemos: Fixed: click-driven opening is suppressed, so clicking navigates without flashing the hover card.

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
puemos: Fixed: the sentinel is always mounted, and the observer reconnects when its element appears.

packages/ui/src/features/canvas/components/ChannelNav.tsx
puemos: Fixed: restored the Activity tooltip, including on the Activity page.

packages/ui/src/features/canvas/components/ActivityHoverCard.tsx
puemos: Fixed: both surfaces now share unread payload, label, and channel-lookup helpers; the popover also uses Quill Empty components.

packages/ui/src/features/sidebar/components/items/ActivityItem.tsx
puemos: Fixed: SidebarItem is now the native popover trigger, removing the invalid wrapper and duplicate tab stop.

k11kirky (approved):
lets get this in!

PR #3878, opened 2026-07-28
fix(spaces): quill up the spaces nav, list highlight and search field

adamleithp:
Design nits on the spaces layout, following #3864. All behind `code-spaces-layout`.

## What changes

- **The search field moves over the content pane** — out of the `< >` group in the title bar, into a Slack-style field spanning the content column. It's a button wearing a field's clothes: magnifier, placeholder, `⌘K` hint. Opens the same command menu.
- **The nav icons are quill Buttons**, with the `data-selected` treatment the channel rows already use, so the nav and the list read as one control set in either theme.
- **The list's highlight is quill's list convention, not its popup one.** quill renders every `AutocompleteItem` with `data-highlighted:ring-2 ring-offset-1` — right for a floating listbox, a stray outline on a sidebar row, and loud enough at dark-theme contrast to outshout the selected row beside it. Overridden to the hover fill, matching `ProjectSwitcher`.
- **One tooltip provider across the three nav icons.** `primitives/Tooltip` mounts a provider inside *every* tooltip, and the skip-delay window is provider state — three isolated providers never share it, so every hop between neighbours paid the full warm-up again. They move to quill's `Tooltip` under a single provider.
- **The last two sidebar tooltips move to quill too** — the create FAB and the "Back to spaces" row still rendered the old dark card. The FAB carried both implementations (quill off the layout, legacy on it); its branches collapse into one quill tooltip that only chooses the text.
- **The Inbox envelope fills when active**, like the bell and the bolt; it was the only icon left at a fixed weight.

## Notes

`--fill-selected` and `--fill-hover` are per-theme tokens (6%/10% and 4%/7% of `foreground`), so highlight sits below selected in weight in both light and dark. The light-mode gap is the narrow one — worth an eyeball.

The other call sites of `primitives/Tooltip` have the same isolated-provider limitation; only these three changed here.

## Testing

Typecheck clean; 306 tests pass across canvas + command.

1. Hover Inbox, then slide across Activity → Command Center — each tooltip appears instantly. Step away and back — full delay again.
2. Arrow through the spaces list — highlight is a fill, no outline ring; the space you're in stays distinguishable. Check both themes.
3. Search field stays centred over the content column as the window resizes and the sidebar is dragged wider; click and ⌘K both open the menu; the bar beside it still drags the window.
4. Flag off — tab strip back, no search field, nav untouched.

### DIFF
--- a/packages/ui/src/router/routes/__root.tsx
+++ b/packages/ui/src/router/routes/__root.tsx
@@ -2,7 +2,6 @@ import {
   ArrowSquareOut,
   CaretLeftIcon,
   CaretRightIcon,
-  MagnifyingGlass,
 } from "@phosphor-icons/react";
 import { useHostTRPC, useHostTRPCClient } from "@posthog/host-router/react";
 import { Button, ButtonGroup } from "@posthog/quill";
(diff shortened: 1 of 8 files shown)


packages/ui/src/router/routes/__root.tsx
```diff
@@ -399,58 +394,38 @@ function RootLayout() {
...
-              strip is also the only global owner of Cmd+W, so something has to
-              keep holding that key when it isn't mounted. */}
+              task view); inbox/activity live in the sidebar nav. The strip is
+              also the only global owner of Cmd+W, so something has to keep
+              holding that key when it isn't mounted. */}
           {channelsLayout ? (
-            <TabShortcutFallback enabled />
+            <>
+              <TabShortcutFallback enabled />
+              <CommandSearchBar onClick={toggleCommandMenu} />
```
puemos, line 427: Not sure why but this is not in the middle

Image

Image

i think we can make it shorter and also position in the absolute center

puemos (approved):
Ok to merge but i do think the search can be even better

PR #3890, opened 2026-07-28
feat(terminal): add a default directory setting for new terminals

lukebaber-posthog:
## Problem

There's no way to tell PostHog Desktop which directory a new terminal should open in. Today the working directory is inferred: one registered folder gets used automatically, several means a folder prompt every time, and nothing registered falls back to the home directory. On top of that, the directory has to be a registered *project* folder — so a plain notes/scripts directory can't be the default at all.

Raised by a teammate who wanted new terminals to consistently open in one personal working directory.

## Changes

- **Settings → Terminal → Default directory** — a native folder picker that accepts any path, not just a registered project folder. Once set, new terminals open there with no folder prompt.
- A **"Terminal in…"** entry stays in the new-terminal footer as the one-off escape hatch for opening somewhere else. When the setting is empty, behaviour is byte-for-byte what it is today.
- `ShellService.resolveWorkingDir` now expands a leading `~` before its `existsSync` check. Previously a tilde path failed that check and was silently discarded in favour of the home directory — which also quietly affected the `cwd="~"` used by the CLI onboarding step.

The preference lives in the existing persisted `settingsStore` alongside the other terminal settings; no new tRPC procedure or host-side store. The row is hidden on hosts without local workspaces, since there's no local pty there.

## How did you test this?

Ran locally:

- `pnpm --filter @posthog/workspace-server exec vitest run src/services/shell/shell.test.ts` — 14 passed, including new cases covering tilde expansion and the fallbacks for empty/missing/undefined cwd.
- `pnpm --filter @posthog/ui exec vitest run src/features/settings/settingsStore.test.ts` — 37 passed, including new persistence/rehydration cases for the setting.
- `pnpm --filter @posthog/ui exec vitest run src/features/command-center` — 33 passed.
- `biome check` clean on all touched files.
- `pnpm --filter @posthog/ui typecheck` reports 69 errors, all pre-existing — identical count on a stashed tree, caused by unbuilt `@posthog/agent`/`@posthog/enricher` dists in this sandbox. None are in the touched files.

From my local inspection, behavior is as expected and new terminal sessions open in the selected directory.  Here is a screenshot of the change to the terminal settings section: 
image

## Automatic notifications

- [ ] Publish to changelog?

---


andrewm4894 (approved):

charlesvien:
Hey @lukebaber-posthog, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3906, opened 2026-07-29
Clarify diff view controls

dmarticus:
## Problem

Task diff controls use “review panel” terminology, making the diff surface harder to discover. Cloud runs also expose panel splitting despite not supporting the local workspace content that makes split panes useful, leaving behind an empty pane.

**Why:** Users should be able to identify where code changes can be reviewed without accidentally creating a cloud pane they cannot use or dismiss.

## Changes

Rename the relevant controls to use “diff view,” hide panel splitting for cloud runs while preserving it for local runs, and add a top-right close action for previously-created empty cloud panes.

## How did you test this?

Tested locally in the Electron dev app:

- Verified a local run retains the panel splitter with a single Chat tab.

Screenshot 2026-07-28 at 9 44 39 PM

- Verified a cloud run does not offer the panel splitter.

Screenshot 2026-07-28 at 9 44 47 PM

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/command/CommandMenu.tsx
+++ b/packages/ui/src/features/command/CommandMenu.tsx
@@ -340,7 +340,7 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
         ? [
             {
               id: "open-review-panel",
-              label: "Open review panel",
+              label: "Open diff view",
               icon: (
                 <ViewVerticalIcon className="h-3 w-3 rotate-180 text-gray-11" />
               ),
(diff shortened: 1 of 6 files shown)


packages/ui/src/features/command/CommandMenu.tsx
```diff
@@ -340,7 +340,7 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
         ? [
             {
               id: "open-review-panel",
-              label: "Open review panel",
+              label: "Open diff view",
               icon: (
                 <ViewVerticalIcon className="h-3 w-3 rotate-180 text-gray-11" />
```
puemos, line 345: 🟡 Non-blocking: I wonder if we should change the icon to reflect that it's just a diff wheel view and not a split

puemos (approved):
Looks good, just one unblocking commit and also the greptile finding

PR #3909, opened 2026-07-29
feat(ui): unify the activity pane's rows and make them navigable

puemos:
## Problem

The activity pane's rows didn't line up with each other:

- Two avatar sizes (small on "created this task", large on everything below), and
  quill's `ThreadItemAuthor`/`ThreadItemBody` default to 14px while the rest of
  the pane is 13px — so a person's name read as a heading over their own message.
- Several typographic treatments competing: bold-name-plus-muted-verb on events,
  greyed-out lifecycle rows, timestamps at a third size.
- Messages showed four wrapped lines, burying the entries around them.
- PR cards were inconsistent about where they take you: the timeline card always
  left the app for GitHub, the Artifacts row always stayed in it, and the card sat
  jammed against its title.
- Nothing in the timeline was clickable, so it couldn't be used to navigate.

## Changes

| Before | After |
|--------|--------|
| image | image |

Two row species, and no more:

- **Lifecycle events** (task created, run finished) — icon bubble plus a single
  line, in the same size and colour as every other row. "Task created" gets its
  own icon instead of the creator's avatar, so it matches "Task completed" rather
  than impersonating an authored row.
- **Authored rows** (messages, PR/canvas artifacts) — avatar, author line,
  content below with room to breathe.

Plus:

- One avatar size (`sm`) and one text size (13px), with timestamps a step down at
  11px, sitting beside the actor. The timeline's vertical line moves to 1.75rem, which is just the gutter
  centre, so the old magic number derived from the large avatar's width is gone.
- Message rows clamp to the first written line. The Comments tab still renders
  messages in full — that's where you read them.
- **Clicking a message row scrolls the transcript to that message.** A small
  `threadNavigationStore` brokers the request, since the pane is a sibling of the
  transcript — the same shape `reviewNavigationStore` already uses for
  scroll-to-file. Both transcripts are wired, as which one renders depends on the
  `useNewChatThread` setting.
- PR cards pick a target: the card opens the in-app review split, a dedicated
  trailing button opens GitHub. The timeline card and the Artifacts row share it.
- Where the task's own view *isn't* mounted alongside — the channel-home sidebar —
  a `canOpenInPlace` flag keeps rows inert and sends PRs straight to GitHub, so
  neither interaction becomes a dead click. This also fixes a pre-existing dead
  click on the Artifacts tab's PR row in that surface.

Worth flagging for review: this deletes a panel-level
`[data-slot=thread-item-timestamp]` override that never matched anything, because
quill's `TooltipTrigger` overwrites the wrapped element's `data-slot` with its own.
Timestamp styling now lives in `ThreadTimestamp` itself.

Not included: automation-authored messages (e.g. the CI re-entry prompt) still
render under your name. That text originates in the Tasks backend and arrives as
an ordinary user message with no marker, so it needs a change there first.

## How did you test this?

- `pnpm --filter @posthog/ui test` — 271 files / 2316 tests pass, including new
  coverage for the row's accessible name (distinct per row, mirroring what's
  shown), click-to-jump filing a scroll request, rows staying inert with no
  transcript alongside, both PR open targets *and* the external fallback, and the
  scroll-request store. Presentation (clamp, sizes, alignment) is deliberately not
  unit-tested — jsdom computes no layout, so such assertions only restate the
  source; the visual-regression suite covers it.
- `pnpm typecheck` across the workspace; `biome check packages/ui/src` clean.
- Design reviewed as a rendered mockup built from quill's real tokens at the
  pane's real width (attached to the task), not from the running app — this
  sandbox has no display, so the layout and the scroll-to-message jump have not
  been exercised in Electron.

## Automatic notifications

- [ ] Publish to changelog?

---


adamleithp (approved):
nice

adamleithp:
I actually don't like the timestamp on the right, this makes it a one-off design

puemos:
> I actually don't like the timestamp on the right, this makes it a one-off design

@adamleithp I hear you. Do you prefer close to the actor? since it's a timeline i went with the timestamp on the right, but i don't have strong opinion about it, can still change before the merge

PR #3937, opened 2026-07-29
Extract thread grouping logic and add stable wrapper identity

adamleithp:
## Problem

The chat thread's tool-run and agent-turn grouping logic was embedded in `ChatThread.tsx`, making it hard to test and reuse. Additionally, during streaming, the grouping wrappers (`ToolGroupItem`, `AgentTurn`) were recreated on every chunk even when their member items hadn't changed, causing unnecessary re-renders of memoized rows and defeating the performance benefit of memoization.

## Changes

**Extract grouping logic to new module:**
- Created `threadGrouping.ts` with `groupToolRuns()`, `groupIntoTurns()`, and the new `createStableTurnGrouper()` function
- Moved helper functions (`isToolCallItem`, `isInvisibleItem`, `INVISIBLE_UPDATES`) to the new module
- Added comprehensive unit tests in `threadGrouping.test.ts` covering all grouping scenarios

**Implement stable wrapper identity:**
- `createStableTurnGrouper()` preserves wrapper object identity across calls when member items are reference-equal
- Reuses previous `ToolGroupItem` and `AgentTurn` wrappers if their member arrays haven't changed
- Enables memoized rows to skip re-renders during streaming (only active-turn wrappers change per chunk)

**Update ChatThread to use stable grouper:**
- Instantiate `createStableTurnGrouper()` once per mounted thread via `useRef`
- Call `turnGrouper.update(items)` in the rows memo instead of `groupIntoTurns(groupToolRuns(items))`
- Added effect to clear group overrides when collapse mode changes (matching legacy thread behavior)
- Increased diff worker pool size from default 8 to 2 (conversation diffs don't need parallelism)

**Update ToolGroup component:**
- Accept `groupId` prop (the run's first tool id) to key manual expand/collapse overrides
- Implement collapse-mode logic: "all" keeps groups collapsed, "partial" streams live turn expanded then collapses on completion, "none" keeps everything expanded
- Manual toggle overrides the mode until mode changes (thread wipes overrides then)
- Use controlled `open` state instead of `defaultOpen` so groups actually collapse on turn completion

**Update VirtualThreadScrollBody:**
- Add `measureElementImmediately` callback to resize rows synchronously on mount
- Fixes paint-then-jump issue where rows scrolled into view would render at 80px estimate then resize

**Update sessionViewStore:**
- Add `useGroupOverride(id)` hook for fine-grained subscription to a single group's override
- Prevents re-rendering all mounted groups when any group's override changes

## How did you test this?

- Added 12 unit tests in `threadGrouping.test.ts` covering:
  - Tool run grouping with invisible items and visible breaks
  - Turn grouping with user-initiated boundaries
  - Stable wrapper identity preservation across streaming chunks
  - Wrapper rebuilding when members change or runs grow
- Updated existing `ToolGroup.test.tsx` to pass new `groupId` prop
- Verified existing tests pass with the refactored logic

## Automatic notifications

- [ ] Publish to changelog?

https://claude.ai/code/session_01EAy2CqYiHQeDTgCEAvAZtG


charlesvien:
Hey @adamleithp, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3944, opened 2026-07-29
feat(agent): tell cloud agents to upload deliverables as artifacts

puemos:
## Problem

Cloud agents can create downloadable files but leave them only in the workspace, making those deliverables inaccessible to users. Even after uploading a file, an agent may reference its local path instead of the working download URL.

## Changes

- Tell cloud agents to upload non-code file deliverables before their final response.
- Require the final response to link the download URL returned by `upload_artifact`, never the local workspace path.
- Keep source code and repository changes in commits and pull requests.

**Why:** Users should receive a working download link whenever an agent produces a file, without needing to remind the agent how to deliver it.

## How did you test this?

- `git diff --check`
- Biome and the agent typecheck could not run in this fresh clone because dependencies are not installed.

## Automatic notifications

- [ ] Publish to changelog?

---


packages/agent/src/adapters/local-tools/tools/upload-artifact.ts
```diff
@@ -2,12 +2,23 @@ import { readFile, realpath, stat } from "node:fs/promises";
...
-import { defineLocalTool, type LocalToolResult } from "../registry";
+import {
+  defineLocalTool,
+  type LocalToolResult,
+  qualifiedLocalToolName,
+} from "../registry";
 
 const MAX_ARTIFACT_UPLOAD_BYTES = 30 * 1024 * 1024;
 
+export const UPLOAD_ARTIFACT_TOOL_NAME = "upload_artifact";
```
k11kirky: this all feels a bit overkill

packages/agent/src/server/agent-server.ts
```diff
@@ -3559,6 +3560,21 @@ Optimize for the fewest shell round trips.
 - Read multiple files at once.
 - Never rerun a command solely to reproduce output you already have.`;
 
+    // Slack sessions cannot deliver artifacts (the Slack host injects a runtime
+    // constraint saying so), so only surface this guidance elsewhere. The
```
k11kirky: very verbose robot, i think this whole comment can go

packages/agent/src/server/agent-server.ts
```diff
@@ -3559,6 +3559,12 @@ Optimize for the fewest shell round trips.
 - Read multiple files at once.
 - Never rerun a command solely to reproduce output you already have.`;
 
+    const artifactInstructions = isSlack
```
puemos: Why only on Slack? I think it should always be there.

packages/agent/src/server/agent-server.ts
```diff
@@ -3625,6 +3625,10 @@ Optimize for the fewest shell round trips.
 - Read multiple files at once.
 - Never rerun a command solely to reproduce output you already have.`;
 
+    const artifactInstructions = `
+## Delivering non-code files (artifacts)
+When you create a non-code deliverable the user should be able to download (a report, chart, image, data file, rendered canvas), call the \`upload_artifact\` tool with its path before your final reply, then mention it. Files left in the workspace don't reach the user; uploaded ones show up in their artifact panel. Don't upload source code or repository changes — those belong in a commit or PR.`;
```
tatoalo: I would point to the presigned URL, would like to avoid agent linking to an hyperlink going nowhere

puemos: can you elaborate? i think now when we upload using this tool we general a presigned URL and that's what's shared. I might missing some context

tatoalo: we do generate it already (pre-this-PR), but I have some instances of agent output not adding the URL correctly. So the problem is not anymore on the artifact being uploaded or not to the S3 bucket (it is), but rather being _referenced_ in the agent output message

puemos: ohh, so you suggest adding this to the prompt. right?

tatoalo: yep, my 2 cents, ofc free to still merge as is. Sorry if I wasn't super clear :sweat_smile:

tatoalo (approved):

packages/agent/src/server/agent-server.ts
puemos: make total sense, i also encounter that a few times. thanks!

puemos: Updated the prompt to require linking the download URL returned by `upload_artifact` in the final response and to explicitly forbid linking the local workspace path. I also removed “rendered canvas” because canvases are separate first-class artifacts, not files delivered through this tool.

PR #3957, opened 2026-07-29
fix(loops): group loops by visibility

benben:
## Summary

- group personal and team loops by their visibility instead of creator ownership
- move team-visible loops created by the current user into Team loops
- remove the current-user request that was only needed for ownership grouping
- add regression coverage for team loops created by the signed-in user

## Testing

- `pnpm exec vitest run src/features/loops/components/LoopsListView.test.tsx` (4 passed)
- `pnpm exec biome check src/features/loops/components/LoopsListView.tsx src/features/loops/components/LoopsListView.test.tsx`
- full typecheck/pre-commit hook is currently blocked by the unrelated missing `typebox` dependency in `@posthog/harness`

### DIFF
--- a/packages/ui/src/features/loops/components/LoopsListView.test.tsx
+++ b/packages/ui/src/features/loops/components/LoopsListView.test.tsx
@@ -48,12 +48,11 @@ function controlledPanel(tab: HTMLElement): HTMLElement {
 }
 
 describe("LoopsListViewPresentation", () => {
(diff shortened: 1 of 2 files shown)


charlesvien:
Hey @benben, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3958, opened 2026-07-29
Open the diff view responsively

dmarticus:
## Problem

The “Open review” footer action and keyboard shortcut should behave consistently, while still choosing a layout that fits the available window space.

**Why:** A side column preserves conversation context on wide windows, but leaves too little room on smaller screens. I also didn't like how the keyboard shortcut had different behavior than the click.

## Changes

Both the footer action and keyboard/command entry points now use one responsive rule: split view at 1280px and wider, expanded view below 1280px. Also use the existing Git diff icon for the “Open diff view” command.

## How did you test this?

Visual regression test babyyy

Demo: https://screen.studio/share/S7kfY2sL

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/code-review/getDefaultReviewMode.test.ts
+++ b/packages/ui/src/features/code-review/getDefaultReviewMode.test.ts
@@ -1,0 +1,15 @@
+import { describe, expect, it } from "vitest";
+import {
+  getDefaultReviewMode,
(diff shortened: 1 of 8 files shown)


puemos (approved):
Love it

PR #3971, opened 2026-07-29
Remove skill buttons from the task header

cvolzer3:
## Summary

- remove the skill-action split button and dropdown from the task header
- delete skill-button prompt metadata, analytics, optimistic state, transcript presentation, stories, and visual baselines
- preserve the experiment-discovery prompt while rendering historical tagged prompts as ordinary user messages

## Why

This removes the top-right shortcuts that automatically sent PostHog instrumentation skill prompts to the agent.

## Impact

Users no longer see or invoke Track events, Track errors, Trace LLM calls, Capture logs, Add feature flag, or Run experiment from the task header. Existing local transcripts keep their underlying prompt text and render it as a standard user message.

## Validation

- `pnpm typecheck` (commit hook)
- `pnpm --filter @posthog/core test` (2,808 tests)
- `pnpm --filter @posthog/ui test` (2,375 tests)
- targeted legacy/group tests (37 tests)
- `pnpm --filter code build-storybook`
- Biome check
- host boundary check


charlesvien:
Hey @cvolzer3, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #3411, opened 2026-07-14
feat(desktop): Implement loops ➿

charlesvien:
## Problem

Loops put recurring PostHog work on autopilot. Instead of kicking off the same agent task by hand every time, you define a loop once and it runs itself on PostHog's managed cloud infrastructure: on a schedule, when a GitHub event fires, or when your own code calls its API. This PR adds the desktop UI to create and manage them. The backend API lives in the posthog repo [[companion PR]](https://github.com/PostHog/posthog/pull/70604).

## Changes

- New top-level **Loops** section (Alpha-badged in the sidebar): a landing page with a prompt composer, your loops and starter templates, plus a loop detail view with run history, enable/pause and "Run now".
- The composer launches an **interactive cloud loop-builder**: a sandbox agent drafts the loop with you and presents it through the `loops-review` card, whose Create button is the only path that creates the loop. Templates and "Create manually" open the guided wizard (Prompt, When, Options, Review) with a clickable stepper.
- Triggers: schedule via a plain frequency picker (Hourly / Daily / Weekdays / Weekly / Once) that shows cron it can't represent as read-only "Custom" instead of rewriting it, GitHub events per repository, or an authenticated API endpoint.
- **Attach a loop to a context**: post runs to the feed, keep `context.md` updated and/or maintain a canvas (forces team visibility). Each context gets its own **Loops** tab with a scoped composer.
- Per-channel notifications (push, email, Slack), an **Auto-fix pull requests** toggle and model/adapter/reasoning effort behind **Advanced**; leaving the model unset lets the backend pick per run.
- Safety-limit rejections from the backend surface as a distinct "Safety limit reached" message with the reason instead of a generic failure.
- Hand-written `@posthog/api-client` loops module (CRUD, run/trigger with idempotency keys, run history, preview) until the OpenAPI client regenerates.
- New agent `finish` tool, gated to background cloud runs: the agent marks its task run completed or failed so the sandbox tears down instead of idling to a timeout.
- MCP apps get a lazy config resolver so cloud runs (which never open a local session) can render the `loops-review` card on desktop, and OAuth now requests the explicit scope list (including `loop:read`/`loop:write` and `llm_gateway:read`) instead of `*`, with a scope version bump forcing re-auth.

## How did you test this?

- CI is green on the head commit: typecheck, unit-test, e2e, integration-test, visual-regression, build and quality all pass.
[...]

PR #3974, opened 2026-07-29
feat(oauth): re-land explicit scopes now that Code ceilings are seeded

MattBro:
## Problem

PostHog Code still requests `scope=*` at sign-in. That blocks retiring the wildcard server-side and keeps the desktop token full-privilege.

#3411 shipped an explicit scope list (plus `llm_gateway:read`) and bumped `OAUTH_SCOPE_VERSION`, but the US/EU Code OAuth app ceilings were not seeded yet. Fresh sign-ins hit `invalid_scope`, so #3668 reverted to `*`. Seeding + `*`→ceiling narrowing are now live in both regions, so the client change is unblocked.

## Changes

- `OAUTH_SCOPES`: regenerate from live `scopes_supported` on `us.posthog.com` (203 scopes, includes `loop:*` and scopes added after #3411) **plus** privileged `llm_gateway:read` last. Drops `agents:*` / `agent_approvals:*`, which left the advertised set.
- `OAUTH_SCOPE_VERSION` 5 → 6 so existing installs re-authorize onto the narrower set.
- Comment on `OAUTH_SCOPES` documents the deploy-order guardrail (seed ceiling in US+EU **before** shipping a non-`*` client / version bump) so this doesn't get bundled again without the seed.
- Restore structural tests (no dups, `llm_gateway:read` last, well-formed strings) + fingerprint guard.

## Preconditions (already done)

1. `#70538` `*`→ceiling narrowing deployed US+EU
2. Code app ceilings seeded to `["@default", "llm_gateway:read"]` in US+EU
3. Verified: real `*` sign-in narrowed correctly in US; zero `invalid_scope` from this path

## How did you test this?

- `pnpm --filter @posthog/shared test` — 767 passed, including `oauth.test.ts` (fingerprint + structural guards)
- Pre-commit `pnpm typecheck` clean

## Related

- Original attempt: #2647 (closed), shipped inside #3411, reverted by #3668
- Server narrowing: PostHog/posthog#70538
- Wildcard retirement: PostHog/posthog#60342

## Automatic notifications

- [ ] Publish to changelog?


charlesvien (approved):

charlesvien:
Moved to the monorepo as PostHog/posthog#76812

PR #3976, opened 2026-07-29
perf(sessions): persist conversation build caches across task re-opens

arnohillen:
## Problem

Clicking a task in the sidebar takes seconds before the transcript renders, even when the session data is already warm in the store. The task detail route fully unmounts on navigation, and the incremental conversation builder (`useConversationItems`) and thread grouper both live in component-scoped refs, so every re-open falls back to a full O(n) re-parse of the transcript on the main thread. `yieldToPaint` already documents the mount blocking "for hundreds of ms"; with multi-MB transcripts it is seconds.

## Changes

- New `conversationDerivedCache` module: a small LRU cache (8 tasks per scope, backed by `lru-cache`, already a workspace dependency via harness) keyed per call site + task that keeps the incremental builder state and thread grouper alive across mounts.
- `useConversationItems` accepts an optional persist key; without one, behavior is unchanged (per-component ref).
- `ConversationView`, `AcpChatThread`, and `ChatThreadFooter` opt in with distinct scopes (distinct scopes matter: the thread and its footer feed different inputs for the same task, and sharing one builder would make them invalidate each other every render).
- Entries are swept when the session residency system evicts a session's events (or the session is torn down), so the memory reclaim that eviction exists for still works. Entries for tasks that never had a session in the store (e.g. archive surfaces) are exempt from sweeping, so unrelated store commits don't cause rebuild churn.

Warm re-opens (session still in the store) now reuse already-built items: idle sessions return the memoized result identity, and streaming sessions take the append-only fast path because the store keeps appending while the view is unmounted and immer preserves element identity.

## How did you test this?

- New unit tests for the cache module: entry persistence, scope isolation, LRU bound, eviction sweep on `evictEvents`/`removeSession`, and the no-session exemption (guards rebuild churn on archive surfaces).
- New hook tests: identical result identity across remount with a persist key (idle), completed-turn item reuse across remount while streaming, and unchanged rebuild behavior without a persist key.
- Full `@posthog/ui` vitest suite: 281 files, 2386 tests passing.
- `pnpm typecheck` and biome via lint-staged.

## Automatic notifications

- [ ] Publish to changelog?

---


arnohillen:
Good catch on the LRU counting scope+task pairs. Fixed in 35a5be0: entries are now grouped per scope with the cap applied inside each scope, so the three conversation scopes no longer shrink the per-task capacity. Added a test asserting a task in one scope survives filling another scope to the cap.

charlesvien:

Shout if you get stuck.

arnohillen:
@charlesvien ported to the monorepo as PostHog/posthog#76788, via the /porting-code-prs skill (original commits and authorship preserved with git am, lockfile change re-derived against the monorepo lockfile). Fine to close this one once the port merges.

PR #3977, opened 2026-07-29
perf(agent): skip session JSONL sanitize when file is unchanged

arnohillen:
## Problem

Part 2 of the task re-open latency work (see #3976). Every agent reconnect runs `sanitizeSessionJsonl` over the native Claude transcript (`~/.claude/projects/.../<sessionId>.jsonl`): a full read plus a `JSON.parse` per line. These files grow to tens of MB on long sessions, so every re-open of a cold task pays seconds of parse work before the session can resume, even when the file was already verified clean moments earlier.

## Changes

- Remember the file stat (mtime + size) of the last clean or healed pass in a small bounded module cache (`lru-cache`, already a workspace dependency), and return early when the current stat matches.
- The Claude SDK only appends to the JSONL, which changes both size and mtime, so any real change forces a full re-parse. A stale memo can only cause extra work, never a skipped heal. The stat-equality assumption is the same one the existing concurrent-writer guard already relies on.

## How did you test this?

- New test proving the skip through the public interface: a dirty file whose stat is restored to the memoized clean-pass stat must not be healed (would fail if the memo check were removed), and a file that grows past a clean pass must be re-sanitized.
- Full `jsonl-hydration` suite: 66 tests passing.

## Automatic notifications

- [ ] Publish to changelog?

---


charlesvien:

Shout if you get stuck.

arnohillen:
@charlesvien ported to the monorepo as PostHog/posthog#76791, via the /porting-code-prs skill (original commits and authorship preserved with git am, lockfile change re-derived against the monorepo lockfile). Fine to close this one once the port merges.

PR #3978, opened 2026-07-29
perf(sessions): extend transcript eviction grace from 20s to 3min

arnohillen:
## Problem

Part 3 of the task re-open latency work (see #3976, #3977). A settled task's parsed transcript is evicted 20 seconds after its view unmounts. Anyone flipping between a few tasks slower than that pays a full disk read plus a three-pass re-parse of the log on every return, which is seconds of main-thread work for large transcripts.

## Changes

- `SESSION_EVENT_EVICT_GRACE_MS` goes from 20s to 3 minutes. Eligibility is unchanged: only disconnected, settled sessions with no pending prompt are evicted, and the eviction/rehydration machinery is untouched.
- Exported the constant and the residency test now imports it instead of duplicating the value.

The memory tradeoff is bounded: only transcripts viewed in the last 3 minutes stay warm, and everything else about the residency system (rehydrate on return, retry after failed reads) behaves exactly as before.

## How did you test this?

- Existing residency suite covers the behavior around the constant (evict after grace, cancel on return, rehydrate from disk, retry on failed read); it now imports the constant so it can't drift. All 6 tests pass.
- No new tests: the change is a tuning constant; asserting its literal value would be a change-detector.

## Automatic notifications

- [ ] Publish to changelog?

---


charlesvien:

Shout if you get stuck.

arnohillen:
@charlesvien ported to the monorepo as PostHog/posthog#76792, via the /porting-code-prs skill (original commits and authorship preserved with git am). Fine to close this one once the port merges.

PR #3983, opened 2026-07-30
feat(sessions): default messaging mode to steer

tatoalo:
## Problem

New sessions currently start with messages in **queue** mode, steer mode in cloud is very stable. Let's use that

### DIFF
--- a/packages/ui/src/features/sessions/hooks/useMessagingMode.ts
+++ b/packages/ui/src/features/sessions/hooks/useMessagingMode.ts
@@ -4,15 +4,21 @@ import {
   useMessagingModeStore,
 } from "@posthog/ui/features/sessions/messagingModeStore";
 import { useSessionStore } from "@posthog/ui/features/sessions/sessionStore";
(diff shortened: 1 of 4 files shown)


packages/ui/src/features/sessions/hooks/useMessagingMode.ts
```diff
@@ -4,15 +4,21 @@ import {
...
+ * for the turn to end), else the global default from settings.
+ */
 export function useMessagingMode(taskId: string | undefined): MessagingMode {
   const override = useMessagingModeStore((s) =>
     taskId ? s.modesByTaskId[taskId] : undefined,
   );
+  const isCloud = useSessionIsCloud(taskId);
   const globalDefault = useSettingsStore((s) => s.defaultMessagingMode);
-  return override ?? globalDefault;
+  return override ?? (isCloud ? "steer" : globalDefault);
```
puemos: should we have globalDefault for cloud and globalDefault for local instead of hard code it here?

puemos (approved):

PR #4008, opened 2026-07-30
refactor(spaces): rework sidebar row status, pins and row menus

adamleithp:
Reworks the spaces sidebar: what a row says about itself, where its actions live, and what colour "unread" is.

## Row status vocabulary

- The quiet state reads "All caught up".
- A task's state is a single dot — blue wants a decision, amber is working or unread, grey is quiet — drawn from one table (`taskStatusVocabulary.ts`) instead of a per-call-site glyph. A working row gets a `DotRingSpinner` in the same 8px box, so labels start in one place down the list.
- Identity moves to a stacked avatar group at the row's right edge: origin, cloud run, PR/branch state. Origins deliberately share one glyph; the PR badge is the one that earns colour.
- Tooltips share a per-row provider, so moving from the dot to a badge doesn't re-wait the open delay.
- A canvas keeps its template glyph, now in canvas violet, in that same stack.
- New `/design-system` route renders the whole vocabulary as a table (`taskIconSpecs.ts`), so a state that has no name is visible.

## Pins

- The `PINNED` section is gone. Pins sit in the one sessions list, marked by an amber pin badge that joins the row's stack — first, which in a reversed stack puts it leftmost and underneath.
- They sort ahead of newer items: a pin is a request not to lose something, and below recency order it would fall off the 30-row cap.

image

## Row actions

- The `…` button is gone from rows — at rest a row shows status, not controls.
- Actions live in the hover card and on right-click, both built from one item definition so they can't drift.
- Canvases get that menu: pin, and delete via the existing `deleteCanvasWithUndo` window (8s toast, nothing sent to the host until it expires). Filing and command-centre cells are task-shaped, so a canvas's menu drops them rather than showing them dead.
- The hover card is quill's `Card`/`Item` parts throughout — no hand-rolled shadow or type scale, and the card's own padding is off so its rules run edge to edge and the action rows highlight full width. It's controlled, so it survives its own portaled "File to…" flyout.
- "File to…" is a searchable flyout (`SearchableMenuFlyout`), pinned to its trigger rather than re-aligning as the list filters, and it opens on hover in the card the way a submenu does on right-click.
- Deleting a canvas confirms first, with the copy the artifacts grid and canvas header already use — it goes for everyone in the space — and the undo window follows the confirm. `Delete…` is quill's destructive button.
- A canvas waiting out its undo window flashes a red dot labelled "Deleting…": it stays in the list rather than vanishing, so the row has to say what's happening to it.
- No separators in either surface.

## Sidebar chrome

- Project switcher moves to the top, above the nav row (channels layout only; the code layout keeps it in the footer).
- "New session" leads the space pane's list of pages, as a sidebar row.
- Space rows drop the cube glyph — it said nothing the name didn't. `#me` keeps its lock; the SPACES heading keeps its icon.
- Feed / Context / Loops / Artifacts drop their icons for the same reason.
- Back row, sessions-header buttons and the menu trigger are quill buttons now, at quill's sizes and radii; rows use quill's focus ring instead of a custom amber outline.
- The "Created by" filter is hidden in `#me`, where it can only answer "you" — and its value is neutralised there, so a filter set in a shared space can't silently empty the list with no control left to undo it.

image

image

## Unread colour

- Count badges, unread dots, the row's status dot and the working spinner all read `--primary` — yellow in dark, orange in light — so "there is something here for you" is one colour across the window. The numeral takes `--primary-foreground`, so contrast follows the theme rather than a token picked per scale. Red is left for failure.
- The pin badge keeps its own amber: primary means a claim on your attention, and a pin is a shelf.

image

## Fixes

- ⌘-slot hints were clipped 4px: quill's `.quill-button kbd` hangs a shortcut into its button's padding, and the row's `truncate` wrapper cut it off. Diagnosed against the running app over CDP — extra row padding doesn't help (the `ml-auto` re-absorbs it), zeroing the margin does.
- A task with an open PR no longer reads as "working": the cloud workflow holds a run `in_progress` while it babysits CI, which under a merge queue can outlast the agent by hours.
- Base UI no longer warns on every task-row render (`nativeButton={false}` on the nested menu trigger).

## Testing

`pnpm typecheck` (24 packages) and `pnpm --filter @posthog/ui test` (2367 tests) pass. Row/menu/glyph behaviour is covered by unit tests; the badge and hover-card changes were checked against the running app.

https://claude.ai/code/session_01JrvaQrZNBNSexddG78u4Nt


packages/ui/src/features/canvas/components/ChannelItemRow.tsx
```diff
@@ -167,52 +312,116 @@ export function ChannelItemRow({
...
+                </ItemContent>
+              </Item>
+              {statusLabel && (
+                <div className="px-2 pb-2">
+                  <Badge variant={runStatusVariant(item.rawStatus)}>
+                    {statusLabel}
+                  </Badge>
                 </div>
+              )}
+              {author && (
```
puemos, line 349: 🟣 Nice-to-have:

I think this should be much smaller, it's taking so much space. In a team , people should know a person just by the avatar? not sure but I'm sure this can have smaller footprint

Image

packages/ui/src/features/canvas/components/ChannelSidebar.tsx
```diff
@@ -80,35 +82,39 @@ function RecentSectionHeader({
...
           aria-label="Search"
           aria-pressed={searchOpen}
           onClick={onToggleSearch}
           className={cnHeaderButton(searchOpen)}
         >
           <MagnifyingGlass size={12} />
-        </button>
+        </Button>
         <DropdownMenu>
           <DropdownMenuTrigger
```
puemos, line 108: 🟣 Nice-to-have:

I wonder if we should align these statuses with the new statuses we have (the groups, not each one, like, waiting, working, done). 

Image

packages/ui/src/features/sidebar/components/items/TaskStatusDot.tsx
```diff
@@ -0,0 +1,186 @@
...
+    </Tooltip>
+  );
+}
+
+/**
+ * A task's state as a single dot: blue wants a decision, the brand yellow is
+ * working or unread, grey is quiet. The trigger renders as a span because rows are
+ * `<button>`s — a nested button would be invalid HTML.
+ */
+export function TaskStatusDot({ dot }: { dot: TaskDot }) {
```
puemos, line 65: 🟣 Nice-to-have:

I really like the stack and the way you did it, but I do think that it's hard to see when something is pinned at a glance. I'm not sure if it needs to have maybe a color? 

Image

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
```diff
@@ -0,0 +1,263 @@
...
+      style: "solid",
+      pulse: true,
+      label: "Needs permission — blocked on you",
+    };
+  }
+  // A live local session streaming output still counts as working — that's the
+  // agent typing right now, not a run status we inferred.
+  const runClaimsWork =
+    props.taskRunStatus === "in_progress" || props.taskRunStatus === "queued";
+  if (props.isGenerating || (runClaimsWork && !hasPullRequest(props))) {
```
puemos: 🟡 Non-blocking warning

When I run cloud or a local task, the progress icon appears, nice, but after the agent finishes a round, the icon stays in progress.

I’d prefer the yellow dot to indicate pending, since no work is being done.

Image

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
```diff
@@ -0,0 +1,263 @@
...
+  } else if (props.hasDiff) {
+    badges.push({
+      key: "branch",
+      Icon: GitBranch,
+      label: "Has changes",
+      tone: "yellow",
+    });
+  }
+  if (badges.length === 0) {
+    badges.push({ key: "local", Icon: ChatCircle, label: "Local task" });
```
puemos: 🟡 Non-blocking warning

From far, it's a bit difficult to distinguish between the icons for local and cloud. 

Image

packages/ui/src/features/canvas/components/ChannelsSidebar.tsx
```diff
@@ -234,6 +234,13 @@ export function ChannelsSidebar() {
 
         {channelsLayout ? (
           <>
+            {/* Which project you're in is the outermost thing about this window,
+                so under the layout it sits above the nav row rather than in the
+                footer. Its menu opens downward, which is the right direction
+                from the top of a sidebar. */}
+            <Box className="shrink-0 px-2 pb-1">
```
puemos, line 241: 🟡 Non-blocking warning

I feel like this area became a bit loaded with different features that it's a bit hard. 

My recommendation is to move the project selector to the top-right

Image

packages/ui/src/features/canvas/components/ChannelItemRow.tsx
```diff
@@ -108,8 +252,13 @@ export function ChannelItemRow({
     );
   }
 
-  return (
-    <PreviewCard.Root>
+  // One tooltip provider per task row, shared by its dot and badges so moving
+  // between them doesn't re-wait the open delay. Canvas rows have neither.
+  const row = (
```
puemos, line 257: 🟣 Nice-to-have

I don't like using vars for components because it's harder to reason about later (but I'm only human...). Maybe this belongs in a component? I could be wrong

puemos (approved):
Looks 🔥 (in a good way)

I did add some feedback, but none of it is blocking

tatoalo (approved):

packages/ui/src/features/canvas/components/ChannelItemRow.tsx
adamleithp: ill think about this

packages/ui/src/features/canvas/components/ChannelsSidebar.tsx
adamleithp: addressed, will push my changes through and we'll tweak it later (didn't move the project switcher, just made it more or less noticeable)

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
adamleithp: fixed

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
adamleithp: fixed

packages/ui/src/features/sidebar/components/items/TaskStatusDot.tsx
adamleithp: fixed

packages/ui/src/features/canvas/components/ChannelSidebar.tsx
adamleithp: will address later, good idea though

packages/ui/src/features/canvas/components/ChannelItemRow.tsx
adamleithp: tightened it up and aligned it with top icon

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
puemos: @adamleithp this works way better now, but there's another minor issue: when the task is starting, it shows it as a yellow dot, which is okay but not the best 

image

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
puemos: lovely

packages/ui/src/features/canvas/components/ChannelsSidebar.tsx
puemos: it's looks more tidy now already with the change, thanks!

packages/ui/src/features/canvas/components/ChannelItemRow.tsx
adamleithp: but ya more can be done here. I want to spend some time redoing this preview card alone.

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
adamleithp: ooo nice catch, ill do a follow up

adamleithp: also the running state in light mode is not noticeable enough, will address in follow up

packages/ui/src/features/canvas/components/ChannelItemRow.tsx
puemos: make sense

packages/ui/src/features/sidebar/components/items/taskStatusVocabulary.ts
puemos: super

PR #4014, opened 2026-07-30
test: verify PR output remains immutable

k11kirky:
## Problem

The PR output utility should remain a pure transformation so callers can safely reuse their existing output objects.

## Changes

Add a regression test confirming `buildPrOutput` does not mutate its input.

## How did you test this?

- `git diff --check`
- Automated test not run: repository dependencies are not installed, so `biome` and `vitest` are unavailable.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/shared/src/pr-urls.test.ts
+++ b/packages/shared/src/pr-urls.test.ts
@@ -137,4 +137,12 @@ describe("buildPrOutput", () => {
       pr_urls: [A],
     });
   });
+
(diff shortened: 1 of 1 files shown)


charlesvien:
Hey @k11kirky, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

PR #4019, opened 2026-07-30
feat(notifications): native OS notifications for @-mentions

adboio:
## Problem

When a teammate @-mentions you in a task thread, nothing surfaces outside the Activity feed — and there, one-row-per-task projection means the mention row is often overwritten by the next agent event before it's seen. Agent events (completed, needs input) already fire native desktop notifications; human mentions, the highest-signal event in a thread, do not.

## Why

Raised while dogfooding channels: mentions from teammates were effectively invisible unless you happened to be watching the Activity page at the right moment.

## Changes

- New `MentionNotificationsContribution` watches the task-mentions query the channels UI already polls (`useMentionActivity`) and dispatches new mentions through the existing `NotificationBus`, so mentions get the same tiering (suppressed while viewing the task / toast while focused / native + sound otherwise), notification settings gating, and click-through-to-task as agent activity. No new fetch path — one poll, one source of truth; notifications live where the spaces layout lives.
- Pure watch-state logic (`baselineMentionWatch` / `advanceMentionWatch`) lives in `@posthog/core/canvas/mentionNotifications`: the first page after boot or login baselines silently (the backlog isn't news), updates dedupe by message id, and a batch of several new mentions collapses into a single notification — one banner, one sound (same-task batches keep click-through to the task; mixed-task batches fall back to a count), and the auth-scoped query's removal on logout resets the baseline so another account's backlog stays silent.

## How did you test this?

- Unit tests for the core watch-state (baseline, ordering, dedupe, watermark, cap) and the contribution (silent first page, per-mention dispatch shape, no re-notify on refresh, burst summary, logout re-baseline, auth-scoped meta gating) — 14 tests, all passing.
- Ran the existing notifications and canvas suites — passing.
- Biome clean; `biome lint packages/core` shows zero `noRestrictedImports`.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/canvas/mentionNotifications.test.ts
+++ b/packages/core/src/canvas/mentionNotifications.test.ts
@@ -1,0 +1,103 @@
+import type { TaskMention } from "@posthog/shared/domain-types";
+import { describe, expect, it } from "vitest";
+import {
+  advanceMentionWatch,
+  baselineMentionWatch,
+  INITIAL_MENTION_WATCH_STATE,
+} from "./mentionNotifications";
+
+function mention(overrides: Partial<TaskMention>): TaskMention {
+  return {
+    id: "mention-1",
+    message_id: "message-1",
+    task_id: "task-1",
+    task_title: "Fix flaky tests",
+    content: "@[Adam](adam@posthog.com) can you look?",
+    created_at: "2026-07-30T10:00:00Z",
+    ...overrides,
+  };
+}
+
+describe("baselineMentionWatch", () => {
+  it("absorbs the backlog without notifying and remembers the newest timestamp", () => {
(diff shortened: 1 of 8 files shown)


adboio:
CI note: the `visual-regression` failure here is not this PR — all 187 story screenshots pass. The job dies in "Install Visual Review CLI" because `products/visual_review/cli`'s lockfile in PostHog/posthog went out of sync with the sharp bump (PostHog/posthog#74334, ~21:46 UTC); every PostHog/code branch run since ~21:48 fails the same way. Fix: PostHog/posthog#75785. Re-run this job once that merges.

charlesvien:

Shout if you get stuck.

PR #4026, opened 2026-07-30
fix(chat-thread): stop the thread yanking back to an earlier prompt

adboio:
## Problem

In a task's chat thread the view very often scrolls itself back to the top, and the scroll-to-bottom arrow only moves you one prompt further down instead of to the bottom.

Both come from marking user rows as `scrollAnchor`. The quill/shadcn scroller engine treats an anchored row as one it owns the scroll position for: on each content change it pins the first anchor it has **not** pinned before to the top of the viewport. Anchors that exist at mount are never recorded as pinned, so the first row swap after opening a task jumps to the very first prompt, the next swap to the second, and so on. That also eats the scroll-to-bottom: you land at the bottom, then the next swap moves you to the next un-pinned prompt.

Separately, the arrow never re-armed auto-follow (the code comment claimed it did), so hitting bottom mid-reply lasted exactly until the next streamed row.

**Why:** reported as extremely disruptive while reading a running task — you lose your place constantly and can't get back to the live tail.

## Changes

- No thread row is a `scrollAnchor` any more, which removes the whole class of engine-initiated scroll jumps.
- The current turn (minimap highlight, and the anchor handed to the windowed body when a thread crosses the virtualization threshold) is measured from the viewport instead, reusing the `computeStickyAnchor` helper the virtualized body already uses.
- The scroll-to-bottom button re-arms auto-follow, so it keeps the bottom while a reply is still streaming.
- `MessageMinimap` no longer subscribes to the engine's per-scroll visibility state — both bodies now pass the anchor in.

## How did you test this?

- `pnpm typecheck` (all 24 packages), `pnpm lint`, `pnpm --filter @posthog/ui test` (2411 tests) — all pass.
- Not manually verified in the running app; the fix is derived from reading the scroller engine's content-change handling, so a dogfood pass on a long thread is worth doing before undrafting.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/ui/src/features/sessions/components/chat-thread/ChatThread.tsx
+++ b/packages/ui/src/features/sessions/components/chat-thread/ChatThread.tsx
@@ -931,15 +931,13 @@ function ThreadTurnAnchor({
   const [anchorId, setAnchorId] = useState<string | null>(null);
   const probeRef = useRef<HTMLSpanElement>(null);
   const measureRef = useRef<(() => void) | null>(null);
-  const promptIds = useMemo(
-    () =>
-      new Set(
-        items
-          .filter((item) => item.type === "user_message")
-          .map((item) => item.id),
-      ),
-    [items],
-  );
+  const promptIds = useMemo(() => {
+    const ids = new Set<string>();
+    for (const item of items) {
+      if (item.type === "user_message") ids.add(item.id);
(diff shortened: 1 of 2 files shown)


charlesvien:

Shout if you get stuck.

PR #3970, opened 2026-07-29
feat: add comments to task artifacts

puemos:
> **Moved to #4067** — recreated on a `posthog-code/` branch (rebased onto current `main`, squashed for the signed-commit workflow). Continuing there; closing this PR.

## Summary

Adds Google Docs/Figma-style commenting to task artifact tabs:

- inline text comments on Markdown and HTML
- yellow circular comment pins on images
- document-level comments for every artifact
- sidebar threads with replies, mentions, Open/Resolved filtering, resolve/reopen, and anchor navigation
- comment counts in artifact rows

Resolved threads remain available under the Resolved filter but are not rendered on the artifact: no text highlight, click target, or image pin.

## Architecture

- Reuses PostHog's generic Django `Comment` API with `scope=task_artifact`; no new persistence model.
- Reuses the generated OpenAPI client for paginated reads and writes.
- Reuses Thread's `MentionComposer`/`MentionText` and the API's existing `mentions` field.
- Reuses Quill for cards, controls, filters, badges, empty states, avatars, and comment inputs.
- Uses one small Zod-validated anchor union: text quote+position, normalized image region, or document.
- Stores only `{ anchor, threadState? }` in `item_context`; artifact identity remains the existing `item_id`.
- Groups roots/replies and calculates resolved state once, shared by the sidebar and renderers.
- Uses PAT-compatible thread-state replies for resolve/reopen because the generic `/complete` action rejects PAT access.

### HTML artifacts

Inspired by [Peek's open-source annotation bridge](https://github.com/puemos/peek/blob/main/internal/web/assets/bridge.js):

- render original authored HTML in an opaque-origin iframe
- inject a small selection/highlight bridge
- exchange validated anchors over `postMessage`
- preserve authored CSS/resources while keeping `allow-same-origin` disabled

### Images

The existing `ZoomableImage` primitive now accepts an optional overlay. Pins and placement transform with the image automatically, without geometry polling or format-specific zoom logic.

## Validation

- API, core, and UI typechecks
- 36 focused tests
- Biome
- host-boundary check
- React Doctor: zero errors

PR #4036, opened 2026-07-31
feat: name the task a comment belongs to so mentions reach its feed

puemos:
## Why

A comment on an artifact or canvas points at an id that lives in a run's JSON, not in a table the server can join against, so the backend had no way to work out which task the comment belonged to. Mentions on those comments reached people by email and the web inbox but never by this app's Activity page — the one surface where the comment is actually readable.

## Changes

The task id now rides along in `item_context` (declared on `commentContextSchema`, which strips unknown keys, so it has to be part of the schema to survive a round-trip). `useCreateComment` takes the task id and merges it in at the single choke point all three call sites go through, so the optimistic row and the request agree.

It is client-supplied, so the backend checks it against the team before acting on it and re-checks visibility when the feed is read.

Stacked on `feat/artifact-comments` (#3970). Pairs with PostHog/posthog#75910, which consumes it — this is inert until that lands, since `item_context` is free-form JSON.

## Testing

`TaskCommentsList` and `ArtifactPreview` suites pass (39 tests), typecheck and lint clean on `@posthog/ui` and `@posthog/core`. No manual testing — no running app in this environment. The behaviour this enables is only observable once the backend side is deployed.

---

### DIFF
--- a/packages/core/src/comments/anchors.ts
+++ b/packages/core/src/comments/anchors.ts
@@ -59,6 +59,9 @@ export type CommentAnchor = z.infer<typeof commentAnchorSchema>;
 export const commentContextSchema = z.object({
   anchor: commentAnchorSchema,
   threadState: z.enum(["resolved", "open"]).optional(),
+  // The task the commented resource belongs to. Artifact and canvas ids live in a run's
+  // JSON rather than a table, so the server can't get back to the task without being told.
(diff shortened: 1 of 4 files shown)


puemos:
Folding this into #3970 — it belongs with the rest of the artifact-comments work, no reason for a separate PR.

PR #4040, opened 2026-07-31
Keep the main page stable when selecting a space

k11kirky:
## Problem

Selecting a space immediately replaced the main window with that space’s feed.

**Why:** Space selection should scope the sidebar first, while navigation should happen only after the user chooses content within that space.

## Changes

- Keep the current main-window route when a space is selected in the Spaces layout.
- Preserve the existing sidebar scoping and keyboard-selection behavior.

## How did you test this?

- `pnpm exec vitest run src/features/canvas/components/ChannelsList.test.tsx` (20 tests)
- `pnpm exec biome check src/features/canvas/components/ChannelsList.tsx src/features/canvas/components/ChannelsList.test.tsx`

## Automatic notifications

- [ ] Publish to changelog?

---


tatoalo (approved):
noice

PR #4047, opened 2026-07-31
fix(ui): distinguish local Pi initialization

jonathanlab:
## Problem

Local Pi sessions reuse cloud initialization copy, so the UI can claim it is connecting to a cloud runner.

## Changes

- Rename the initializer to a session-wide component with an explicit execution target.
- Show local Pi copy for local sessions and retain cloud provisioning copy for cloud sessions.
- Add coverage for both initialization targets.

## How did you test this?

- `pnpm --filter @posthog/ui exec vitest run src/features/sessions/components/SessionInitializingView.test.tsx`
- `pnpm exec biome check --write packages/ui/src/features/sessions/components/SessionInitializingView.tsx packages/ui/src/features/sessions/components/SessionInitializingView.test.tsx packages/ui/src/features/sessions/components/SessionView.tsx packages/ui/src/features/pi-sessions/PiSessionView.tsx`

## Automatic notifications

- [ ] Publish to changelog?


jonathanlab:
Fixed in 395c24302. `PiSessionView` now receives the cloud/local target from `TaskDetail`, where task and workspace metadata are already available, instead of inferring it from the asynchronously populated `cloudStatus`. Added coverage for a cloud initializer before its status arrives.

PR #4059, opened 2026-07-31
feat(skills): always-on skills injected into every new task

adboio:
## Problem

Skills only activate today when a task's prompt explicitly references them (`/name` or a composer skill chip). There's no way to say "apply this skill to everything I do" — the way a channel's CONTEXT.md is injected automatically, or the way a SessionStart-hook plugin (e.g. i-have-adhd) puts its instructions in context from turn one.

Refs #3735 (earlier attempt, auto-closed as stale — this supersedes it).

## Changes

- **Always-on toggle** on installed skills (Skills settings: switch on each card + detail panel) for `user`/`marketplace`/`codex`/`bundled` sources. Repo-source skills are excluded so behavior is identical for repo and repo-less tasks. Toggles persist in the settings store as `{name, source}` refs and re-resolve against the live skill list at task creation, so moved/deleted skills self-heal.
- **Cloud tasks** reuse the existing skill-bundle pipeline: resolved refs are appended to the transport's `skillBundles` (uploaded with the first message, warm and cold paths) and a `<always_on_skills>` manifest of `/name` mentions is folded into the first message, which the sandbox's existing inliner expands into each bundle's full SKILL.md. Zero sandbox/server changes.
- **Local tasks** inline each SKILL.md body into the initial prompt (read via the existing `skills.readFile` tRPC), with per-skill/total size caps degrading to path references.
- The injected block is stripped from the transcript and shown as an "Always-on skills (n)" chip; prompt recall and echo dedupe strip it too.

## How did you test this?

- New unit tests: ref toggle/resolution/sanitization, both prompt builders (incl. sandbox mention-regex compatibility and size-cap degrade), saga injection for cloud (bundle upload + first-message manifest) and local (initialPrompt block), settings persistence + rehydrate sanitizer, transcript extractor.
- Ran full suites on touched packages: `@posthog/core` (2863), `@posthog/ui` (2452), `@posthog/shared` (777) — all green; `pnpm typecheck` green across all 24 packages; `biome lint packages/core` clean.

## Automatic notifications

- [ ] Publish to changelog?

---

### DIFF
--- a/packages/core/src/editor/prompt-builder.test.ts
+++ b/packages/core/src/editor/prompt-builder.test.ts
@@ -1,5 +1,9 @@
+import type { ResolvedAlwaysOnSkill } from "@posthog/core/skills/alwaysOnSkills";
 import { describe, expect, it } from "vitest";
 import {
+  ALWAYS_ON_SKILL_MD_MAX_BYTES,
+  buildAlwaysOnSkillsBlock,
+  buildAlwaysOnSkillsCloudText,
   buildChannelContextBlock,
   buildChannelContextText,
   buildCustomInstructionsText,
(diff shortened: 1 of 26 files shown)


adboio:
Ported to the monorepo as PostHog/posthog#76742 (the desktop tree now lives at `products/desktop/` and this repo's main is frozen). The port carries both commits with original authorship, plus a third commit with the composer per-task-exclusion follow-up that never got pushed here. Keeping this PR open until the port merges.

PR #4060, opened 2026-07-31
feat(ui): prototype static spaces sidebar

puemos:
## Problem

Prototype for a new sidebar design: get rid of the moving (sliding) sidenav and replace it with a single static nav where every space shows its tasks beneath it, and where spaces only appear once a user adds them to their sidebar (by adding an existing space or creating a new one) instead of listing all spaces by default.

## Changes

- Under the existing `code-spaces-layout` flag, `ChannelsSidebar` now renders a static `SpacesSidebarNav` instead of the 200%-width panes slider (the `ChannelPanes` track, swipe gesture, and route-scoping side effects are removed from the shell).
- New `SpacesSidebarNav`: Home / Tasks / Inbox nav rows (Inbox has a "+" affordance), then every pinned (starred) space with its tasks expanded inline beneath it, an "Add space" menu (finding an existing space stars it; "New space" opens the create modal), and a pinned-agents placeholder section.
- Pinning = starring: reuses the existing starred-channel shortcuts as the sidebar membership model — no backend changes. Expanded state per space is a persisted local store (`spacesSidebarStore`).
- Space task rows reuse the existing `useChannelItems` + `ChannelItemRow`, so a space's tasks render identically to the old channel pane.

Screenshots to follow once run against a live backend.

## How did you test this?

- `pnpm --filter @posthog/ui test` — all 289 files / 2443 tests pass (rewrote `ChannelsSidebar.test.tsx` for the static nav; kept Archived-row and tracking coverage).
- `pnpm --filter @posthog/ui typecheck` — clean.
- `pnpm biome check packages/ui/src` — clean.

## Automatic notifications

- [ ] Publish to changelog?

---


puemos:
Pushed a design pass over the prototype (e9de196): removed the Home/Tasks/Inbox rows (the icon nav already covers them), dropped the per-space tab strip in favour of the channel header's existing tabs, made space rows open-on-click with a proper disclosure caret and hover-revealed pin star, turned All spaces into a section-labelled directory whose rows open the space (clicking used to silently pin), and restored a create path in this layout via the FAB plus a New space row. Details in the commit message; `AllSpacesSection` now has unit coverage.

---

PR #4063, opened 2026-08-01
fix(toast): auto-dismiss toasts when the app window is unfocused

adamleithp:
## Problem

Toasts in the desktop app don't reliably auto-dismiss — the "Task archived" toast in particular hangs on screen until you close it by hand.

Root cause: the toast primitive wraps quill, which wraps base-ui's Toast. base-ui pauses a toast's auto-dismiss timer whenever the window isn't OS-focused (`hovering || focused || !isWindowFocused`), not only while it's hovered. On a desktop app the window is frequently not the frontmost OS window, so the timer stays paused and the toast never expires. This affects every toast; archive is just where it's most visible, since you often move to another app right after archiving.

## Changes

Added a blur-only fallback in the shared toast wrapper (`packages/ui/src/primitives/toast.tsx`): once a toast's time is up, if the window is unfocused, dismiss it. When the window is focused we do nothing and leave base-ui's own timer — and its hover-to-pause — in charge. Loading toasts and never-expiring (`Infinity`) toasts like the offline banner are left untouched.

## How did you test this?

`vitest run src/primitives/toast.test.ts` in `packages/ui` — 17 pass, including 4 new cases covering: dismissal when unfocused (explicit duration and provider default), no dismissal when focused, and no dismissal for loading/Infinity toasts. Also ran biome check and tsc on the changed files (clean).

## Automatic notifications

- [ ] Publish to changelog?

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09G8Q32R6F/p1785532925449359?thread_ts=1785532925.449359&cid=C09G8Q32R6F)*

### DIFF
--- a/packages/ui/src/primitives/toast.test.ts
+++ b/packages/ui/src/primitives/toast.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const quill = vi.hoisted(() => {
   let n = 0;
@@ -32,6 +32,7 @@ import { toast } from "./toast";
 
 beforeEach(() => {
   vi.clearAllMocks();
+  vi.useFakeTimers();
   quill._reset();
   settings.toastNotifications = true;
   // clearAllMocks resets the level fns to undefined returns; restore ids.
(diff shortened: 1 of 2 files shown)


charlesvien:
Hey @adamleithp, we've migrated this repo into the [PostHog/posthog](https://github.com/PostHog/posthog) monorepo, so this PR needs to be recreated there.

Shout if you get stuck.

